PR #2, opened 2020-11-13
Initial prototype for slurm stats

jovial:


main.tf
```diff
@@ -25,6 +25,29 @@ variable "node_image" {
...
+resource "openstack_networking_secgroup_rule_v2" "secgroup_slurm_login_rule_egress_v4" {
+  direction         = "egress"
+  ethertype         = "IPv4"
+  security_group_id = openstack_networking_secgroup_v2.secgroup_slurm_login.id
+}
+
+resource "openstack_networking_secgroup_rule_v2" "secgroup_slurm_login_rule_ingress_tcp_v4" {
+  direction         = "ingress"
+  ethertype         = "IPv4"
+  # FIXME: only allow ports needed
```
jovial, line 44: Just to clarify, we need to open up the port for slurmdbd.

requirements.yml
```diff
@@ -1,9 +1,15 @@
 ---
+collections:
+- name: containers.podman
+roles:
 - src: stackhpc.nfs
-# - src: stackhpc.openhpchttps://github.com/stackhpc/ansible-role-openhpc
+- src: https://github.com/stackhpc/ansible-role-openhpc
+  name: stackhpc.openhpc
+  version: feature/accounting
```
jovial, line 8:
```suggestion
  version: master
```
Need to merge https://github.com/stackhpc/ansible-role-openhpc/pull/46 before we can revert this

main.tf
sjpb: Looks like 3306 from [docs](https://slurm.schedmd.com/slurmdbd.conf.html)

roles/slurm-stats/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
+---
```
sjpb, line 1: How come just this bit of slurm-stats.yml is broken out as a role? Should this be in the new slurm-tools repo? Or just inlined back into slurm-stats.yml given this is "just" a demo?

sjpb (changes requested):
Could do with some docs in the readme too - i.e. what slurmdb.yml sets up and what slurm-stats.yml adds.

sjpb:
@jovial , do you think we should "namespace" this repo somehow? Either:
- Have subdirectories for incompatible sets of tf/ansible, e.g. `simple/`, `stats/`, or
- Just make the inventory .tpl files match the main ansible playbook, so `simple.tf` and `simple.yml` vs `stats.*` - with any dependent playbooks having that as a prefix? (can still all produce just `inventory`).

jovial:
Yeah, agree that having loads of files in one directory is a good way to
get lost. Would be good to share as much as possible so that we don't have
to maintain two almost identical examples. I half wonder if we should just
have the batteries included demo that does everything (possibly with flags
to disable features).

roles/slurm-stats/tasks/main.yml
jovial: Moved into collection.

slurm-stats.yml
```diff
@@ -0,0 +1,118 @@
...
+        env:
+          # FIXME: assumes same host as elastic
+          ELASTICSEARCH_URL: https://localhost:9200
+          ELASTICSEARCH_HOSTS: https://localhost:9200
+
+- name: Setup slurm stats
+  hosts: slurm_stats
+  tasks:
+    - include_role:
+        name: template
```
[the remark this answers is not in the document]
jovial: Got rid of the .meta madness.

jovial:
Superseded by: https://github.com/stackhpc/openhpc-demo/tree/feature/multi-environments

PR #18, opened 2021-02-17
Fix nfs when control===login

jovial:
The role was trying to mount the share from the same server. Systemd detected
this loop and refused to start the nfs server.

This makes it easier to construct an inventory with a shared control/login node.

### DIFF
--- a/environments/common/inventory/group_vars/all/nfs.yml
+++ b/environments/common/inventory/group_vars/all/nfs.yml
@@ -5,7 +5,10 @@ nfs_server_default: "{{ hostvars[groups['control'] | first ].internal_address }}
 nfs_configuration:
   nfs_enable:
       server:  "{{ inventory_hostname in groups['control'] }}"
-      clients: "{{ inventory_hostname in groups['compute'] or inventory_hostname in groups['login']  }}"
+      # Don't mount share on same server as client...
+      # Could do something like nfs_clients: '"nfs_servers" not in group_names' instead.
+      # See also constructed inventory: https://docs.ansible.com/ansible/devel/collections/ansible/builtin/constructed_inventory.html
+      clients: "{{ inventory_hostname in groups['compute'] or (inventory_hostname in groups['login'] and not inventory_hostname in groups['control']) }}"
   nfs_server: "{{ nfs_server_default }}"
   nfs_export: "/mnt/nfs/"
   nfs_client_mnt_point: "/mnt/nfs/"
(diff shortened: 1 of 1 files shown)


sjpb:
Hmm, I'm not sure this smells right to me TBH, seems to leak info about slurm groups into the nfs setup. Following what we've done elsewhere, shouldn't we just have an nfs_client group and an nfs_server group and set them appropriately? Then actually the hosts for this role could just be both those groups.

Might have to think how would this work for multiple NFS configs though, which might not all have the same client/server setups I guess? I.e. a host can be both client and server at the same time, just not for the same share. (most likely case: entire cluster is client for e.g. "external" Isilon , plus login exports some directory to computes)

jovial:
> Hmm, I'm not sure this smells right to me TBH, seems to leak info about slurm groups into the nfs setup. Following what we've done elsewhere, shouldn't we just have an nfs_client group and an nfs_server group and set them appropriately? Then actually the hosts for this role could just be both those groups.
> 
> Might have to think how would this work for multiple NFS configs though, which might not all have the same client/server setups I guess? I.e. a host can be both client and server at the same time, just not for the same share. (most likely case: entire cluster is client for e.g. "external" Isilon , plus login exports some directory to computes)

Yeah, like you say, not sure how groups would work with multiple shares. This change concerns the `/mnt/nfs` share only. Maybe the logic could be simplified to: `inventory_hostname in groups['cluster'] and inventory_hostname not in groups['control']`.

jovial:
Could also change the role to not mount the share on the server from where it is exported.

sjpb:
> Could also change the role to not mount the share on the server from where it is exported.

I think there are actually some use-cases for that, although whether it's the best idea/something this role needs to support is a fair point.

sjpb:
> The other issue with using groups is that you most likely need to use a constructed inventory to exclude the server, as, for example:
> 
> ```
> [nfs_server:children]
> control
> [nfs_clients:children]
> cluster
> ```
> 
> also includes the server as the client.

With this server/clients groups approach, I was thinking the default would just be:
```
[nfs_server:children]
control
[nfs_clients:children]
compute
login
```

but you're right that you'd have to delete login if that's combined with control

sjpb:
Current fix looks sensible for default case with current groups given that needs to handle both separate and combined login/control groups

sjpb (approved):
Current fix looks sensible for default case with current groups given that needs to handle both separate and combined login/control groups

PR #21, opened 2021-02-17
Remove unnecessary exporters group from minimal layout

### LINKED ISSUE #16: Remove exporters and mysql from minimal layout?
Like the layouts addition. But does `openhpc-demo/environments/common/layouts/minimal` actually need exporters or mysql groups? I'd have just expected nfs and openhpc.

jovial:
Fixes #16


sjpb:
LGTM

PR #25, opened 2021-02-18
remove setuptools from README

sjpb:
Isn't needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ a monitoring stack.
     python3 -m venv venv
     . venv/bin/activate
     pip install -U pip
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -16,7 +16,6 @@ a monitoring stack.
     . venv/bin/activate
     pip install -U pip
     pip install -U setuptools
-    pip install -r requirements.txt
```
jovial, line 19: wrong line?

jovial:
Could we also delete it from: `dev/setup-env.sh`, which is what the CI runs, to verify this in CI?

Issue #13, opened 2021-02-12
Need to document groups


sjpb:
Mostly clear to me but probably only cause of deploying monitoring, would be worth going thro the list (both master and both private versions)

sjpb:
Also wonder if "implementation" groups like e.g. `podman` (are there any others) should be dynamically constructed. I.e. if a tool needs podman to run then it should add it's hosts to that group.

jovial:
> Also wonder if "implementation" groups like e.g. podman (are there any others) should be dynamically constructed. I.e. if a tool needs podman to run then it should add it's hosts to that group.

Seems like a nice idea. Do you see any issues with doing something like
```
[podman:children]
opendistro
kibana
filebeat
```
?

> Mostly clear to me but probably only cause of deploying monitoring, would be worth going thro the list (both master and both private versions)

Are you thinking a table with all the groups that are available and a short summary of what they do?

sjpb:
Ah, that's nice, LGTM.

Or maybe, we just add comments into `common/inventory/groups`? And point to it from the README(s)? Harder to get out of sync at least.

sjpb:
Document that this ONLY supports:

1. Separate control and login nodes. The `login` group is passed as the `openhpc_login_only_nodes` to the openhpc role.
2. Configless mode.

This is intentional to cut down logic paths and make reasoning about groups etc simpler

sjpb:
Should probably check that `cluster` and `openhpc` groups are used appropriately.

sjpb:
Document something like:

- The hosts should be listed in a file `environments/<env>/inventory/hosts` which must define the following groups:
  - `control`: Containing a single host to make the Slurm control node. Multiple (high availability) control nodes are not supported.
  - `login`:  Containing one or more hosts to make Slurm login nodes. Note combined control/login nodes are not supported.
  - `compute`: Containing all hosts to become Slurm compute nodes (in all partitions).
  
  This file is usually created by the deployment automation.

- The cluster name must be defined for all hosts using `openhpc_cluster_name` - setting this in the hosts file using `[all:vars]` is usually convenient.
- Each Slurm partition must contain a homogeneous set of hosts. For each partition:
  - Define an ansible inventory group for each partition as `<cluster_name>_<partition_name>`.
  - Define the partition by setting `openhpc_slurm_partitions` in `environments/<env>/inventory/group_vars/openhpc/overrides.yml`.

[...]

PR #29, opened 2021-02-19
Remove locked memory limits on compute nodes

sjpb:
Fixes [slurm role issue #86](https://github.com/stackhpc/ansible-role-openhpc/issues/86)

### DIFF
--- a/ansible/slurm.yml
+++ b/ansible/slurm.yml
@@ -15,5 +15,11 @@
   tags:
     - openhpc
   tasks:
-   - import_role:
+    - import_role:
         name: stackhpc.openhpc
+    - name: set memory limits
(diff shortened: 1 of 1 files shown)


sjpb:
Actually this isn't quite right: it should change compute and job submission nodes. The current `"'control' not in group_names"` logic works for separate control and login nodes but not for the combined case.

sjpb:
Document that we only support separate login nodes, not combined login/control.

sjpb:
Added comment to #13 for this.

sjpb:
Note that we could just use the compute,login group then for this action and remove logic.

jovial (approved):

PR #32, opened 2021-02-24
Move centos /home directory

sjpb:
When mounting /home via e.g. NFS we really want `centos` to remain a local user.

I've moved it by default to `/var/lib/centos`, see `ansible_user_home`.

Comments show some non-obvious issues.

### DIFF
--- a/ansible/bootstrap.yml
+++ b/ansible/bootstrap.yml
@@ -23,10 +23,6 @@
             regexp: "{{ connection_user.home }}"
             replace: "{{ ansible_user_home }}"
           register: etc_passwd
-         # NB: reboot is not always needed - not clear yet which service is causing problems
-        - name: reboot
-          reboot:
-          when: etc_passwd.changed
         # TODO: delete old homedir
       when:
         - ansible_user_home is defined # TODO: and defined etc etc
--- a/ansible/bootstrap.yml
+++ b/ansible/bootstrap.yml
@@ -1,9 +1,38 @@
 ---
 
 - hosts: cluster
-  gather_facts: false
+  gather_facts: true
+  become: yes
   tasks:
+    - name: Get homedir for ansible_user # note ansible_env.HOME is /root when using become!
+      user:
+        name: "{{ ansible_user }}"
+        state: present
+      register: connection_user
+    - block:
+        - name: Copy ansible_user homedir
+          ansible.posix.synchronize:
+            src: "{{ connection_user.home }}/" # trailing-/ needed so we don't copy the directory itself, just contents
+            dest: "{{ ansible_user_home }}"
+            delete: yes
+            recursive: yes
+          delegate_to: "{{ inventory_hostname }}"
+        - name: modify /etc/passwd
+          replace:
+            path: /etc/passwd
+            regexp: "{{ connection_user.home }}"
+            replace: "{{ ansible_user_home }}"
+          register: etc_passwd
+         # NB: reboot is not always needed - not clear yet which service is causing problems
+        - name: reboot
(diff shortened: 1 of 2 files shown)


jovial:
Just wondered if you considered adding the centos user to `appliances_local_users:`. Then that could function as defacto reference for the local users on the system. Thinking something like:

```
appliances_local_users_default: "{{ [appliances_local_users_podman] if appliances_local_users_podman_enable else [] + [appliances_local_users_centos] + appliances_local_users_extra }}"

appliances_local_users_centos_home: /var/lib/centos
appliances_local_users_centos:
    name: centos
    home: "{{ appliances_local_users_centos }}"
    move_home: true
```
Of course, you might have considered this already and know a reason why that is a crazy idea!

Think that will also move the files if new home doesn't exist.

sjpb:
The problem is that its different because its the user we connect as - so if we try to move it using `move_home: true` you get an error. It should probably have a comment to that effect in the code. The other option would be to connect as root for that change, but then we'd have to add ssh keys to root etc - at least temporarily - and that felt like a worse idea.

jovial:
This seems to work:
```
diff --git a/ansible/bootstrap.yml b/ansible/bootstrap.yml
index ddd03a5..e36a276 100644
--- a/ansible/bootstrap.yml
+++ b/ansible/bootstrap.yml
@@ -1,9 +1,12 @@
 ---
 
 - hosts: cluster
-  gather_facts: false
+  gather_facts: true
   tasks:
     - name: Add users
       ansible.builtin.user: "{{ item }}"
       with_items: "{{ appliances_local_users }}"
+      become_method: "sudo"
+      # Need to change working directory otherwise we try to switch back to non-existent directory.
+      become_flags: '-i'
       become: true
diff --git a/environments/common/inventory/group_vars/all/defaults.yml b/environments/common/inventory/group_vars/all/defaults.yml
index 23474e1..60f7ce5 100644
--- a/environments/common/inventory/group_vars/all/defaults.yml
+++ b/environments/common/inventory/group_vars/all/defaults.yml
@@ -18,11 +18,19 @@ prometheus_address: "{{ hostvars[groups['prometheus'].0].api_address }}"
 
 ############################# bootstrap: local user configuration #########################
 
-appliances_local_users_default: "{{ [appliances_local_users_podman] if appliances_local_users_podman_enable else [] + appliances_local_users_extra }}"
+appliances_local_users_default: "{{ ([appliances_local_users_podman] if appliances_local_users_podman_enable else []) + [appliances_local_users_ansible_user] + appliances_local_users_extra }}"
 # Overide this to add extra users whilst keeping the defaults.
 appliances_local_users_extra: []
 appliances_local_users: "{{ appliances_local_users_default + appliances_local_users_extra }}"
 
+appliances_local_users_ansible_user_name: "{{ ansible_user_id }}"
+appliances_local_users_ansible_user_home: /var/lib/{{ appliances_local_users_ansible_user_name }}
+appliances_local_users_ansible_user:
+    name: "{{ appliances_local_users_ansible_user_name }}"
+    home: "{{ appliances_local_users_ansible_user_home }}"
+    move_home: true
+    local: true
+
 appliances_local_users_podman_enable: "{{ groups['podman'] | length > 0 }}"
 appliances_local_users_podman_home: /var/lib/podman
 appliances_local_users_podman:
```

sjpb:
Somehow this seems to be in `main` already??

jovial:
> Somehow this seems to be in `main` already??

Don't think it is - possibly checking the wrong repo?

sjpb:
Used #34 instead.

PR #34, opened 2021-02-25
Move ansible user home to local storage

jovial:
Alternative to #32


environments/common/inventory/group_vars/all/defaults.yml
```diff
@@ -18,11 +18,19 @@ prometheus_address: "{{ hostvars[groups['prometheus'].0].api_address }}"
...
 
-appliances_local_users_default: "{{ [appliances_local_users_podman] if appliances_local_users_podman_enable else [] + appliances_local_users_extra }}"
+appliances_local_users_default: "{{ (appliances_local_users_podman if appliances_local_users_podman_enable else []) + [appliances_local_users_ansible_user] + appliances_local_users_extra }}"
 # Overide this to add extra users whilst keeping the defaults.
 appliances_local_users_extra: []
 appliances_local_users: "{{ appliances_local_users_default + appliances_local_users_extra }}"
 
+appliances_local_users_ansible_user_name: "{{ ansible_user_id }}"
+appliances_local_users_ansible_user_home: /var/lib/{{ appliances_local_users_ansible_user_name }}
+appliances_local_users_ansible_user:
```
jovial, line 28: Noticed that this is slightly inconsistent with `appliances_local_users_podman` which is a list. The reason why `appliances_local_users_podman` is a list is because I imagined one user per podman service i.e kibana, opendistro, filebeat and I was thinking they'd all be added to that list.

sjpb:
The merge was a bit messy but I think I've got it right but this doesn't look correct:

```
TASK [Add users] *************************************************************************************************************************************
[WARNING]: Using a variable for a task's 'args' is unsafe in some situations (see https://docs.ansible.com/ansible/devel/reference_appendices/faq.html#argsplat-unsafe)
<snip>
ok: [hpc-2] => (item={'name': 'podman', 'comment': 'Used for running all containers', 'home': '/var/lib/podman'})
<snip>
ok: [hpc-2] => (item={'name': 'root', 'home': '/var/lib/root', 'move_home': True, 'local': True})
<snip>
```

jovial:
Hmm, we get the correct result in CI:

https://github.com/stackhpc/openhpc-demo/runs/1988987637#step:7:129

You haven't ended up with the become on the whole play have you?

sjpb:
Hmm maybe. I'll have to create a new cluster and retry.

sjpb:
I just tried this on a new cluster, copying the `ansible/bootstrap.yml` and `environments/common/inventory/group_vars/all/defaults.yml` directly out of this branch and it tries to move `root`: 
```
"item": {
        "home": "/var/lib/root",
...
```
which unsurprisingly fails. Will see if I can fix.

jovial:
> I just tried this on a new cluster, copying the `ansible/bootstrap.yml` and `environments/common/inventory/group_vars/all/defaults.yml` directly out of this branch and it tries to move `root`:
> 
> ```
> "item": {
>         "home": "/var/lib/root",
> ...
> ```
> 
> which unsurprisingly fails. Will see if I can fix.

Very strange - you aren't using fact caching are you? We are using the `ansible_user_id` fact. Maybe we should switch to something like this in common/environments/defaults.yml:

```
ansible_user: "{{ ansible_ssh_user | default('centos') }}"
```

and then:

```
appliances_local_users_ansible_user_name: "{{ ansible_user }}"
```

We can then set `gather_facts` to false.

environments/common/inventory/group_vars/all/defaults.yml
```diff
@@ -18,11 +18,19 @@ prometheus_address: "{{ hostvars[groups['prometheus'].0].api_address }}"
 
 ############################# bootstrap: local user configuration #########################
 
-appliances_local_users_default: "{{ [appliances_local_users_podman] if appliances_local_users_podman_enable else [] + appliances_local_users_extra }}"
+appliances_local_users_default: "{{ (appliances_local_users_podman if appliances_local_users_podman_enable else []) + [appliances_local_users_ansible_user] + appliances_local_users_extra }}"
 # Overide this to add extra users whilst keeping the defaults.
 appliances_local_users_extra: []
 appliances_local_users: "{{ appliances_local_users_default + appliances_local_users_extra }}"
 
+appliances_local_users_ansible_user_name: "{{ ansible_user_id }}"
```
jovial, line 26:
```suggestion
appliances_local_users_ansible_user_name: "{{ ansible_ssh_user | default(ansible_user) }}"
```

ansible/bootstrap.yml
```diff
@@ -1,11 +1,14 @@
 ---
 
 - hosts: cluster
-  gather_facts: false
+  gather_facts: true
```
jovial, line 4:
```suggestion
  gather_facts: false
```

sjpb:
I think this is slightly broken - I've done  a modified version of it in the nrel repo `[nrel-pr34 2d64a8c] add modified version of PR34 to move centos home` which works - @jovial need to discuss how best to fix this PR

sjpb (approved):

PR #35, opened 2021-02-25
Fix issues with /opt mount during tests

sjpb:
- Currently the control node gets /opt mounted, which is unnecessary. Should be fixed by 5e674c8 but needs testing
- The /opt mount doesn't get undone - see [collection issue 21](https://github.com/stackhpc/ansible_collection_slurm_openstack_tools/issues/21)
- The /opt mount overrides the whole of /opt rather than just /opt/intel and /opt/ohpc which means it breaks e.g. the slurm-stats tools.


jovial:
Re-running CI - failure looks unrelated.

jovial:
Now has a merge conflict, but looks good on principle.

sjpb:
> Now has a merge conflict, but looks good on principle.

Fixed now.

jovial (approved):

PR #37, opened 2021-03-02
Docs

### LINKED ISSUE #11: Need to document environment inventory
We need to document what you have to/should define in `<env>/inventory`, otherwise you have to read all the role docs and figure out what common provides. I think it's something like:

1. A variable `openhpc_cluster_name=<cluster_name>` must be defined, e.g. in `<env>/inventory/group_vars/openhpc.yml`.
1. The `openhpc` role uses ansible groups of the form `<cluster_name>_<partition_name>`  to define Slurm partitions. The default partition name of "compute" is defined in `environments/common/inventory/group_vars/openhpc/defaults.yml`.
1. Hosts and groups can be defined in various ways but it is suggested that:
   - `environments/<env>/inventory/hosts` is produced by the deployment templating (e.g. from Terraform or Heat) and defines hosts directly in groups `<cluster_name>_compute` (if using default partition name) as well as `...login` and `...control`.
   - `environments/<env>/inventory/groups` is used to map these into the groups `{baremetal,vm}_{login,control,compute}`. These latter groups are what the playbooks use to set up hosts appropriately.
1. For more advanced options for Slurm partition definition see `https://github.com/stackhpc/ansible-role-openhpc#slurmconf`.

sjpb:
Add documentation to close #11 and #13.

As per discussion with @jovial  will aim for READMEs in environments to *only* cover aspects specific to that environment - everything else should be documented in the root README, or signposted there and covered in a directory-specific README (e.g. `packer/README.md`.

### DIFF
--- a/environments/common/inventory/groups
+++ b/environments/common/inventory/groups
@@ -20,7 +20,7 @@ control
 compute
 
 [builder]
-# Host for (optionally) building compute images - see separate documentation.
+# VM used to (optionally) build compute images - do not add hosts. See packer/README.md.
 
 [podman:children]
 # Hosts running containers for below services:
--- a/environments/common/inventory/groups
+++ b/environments/common/inventory/groups
@@ -1,49 +1,68 @@
-[cluster]
-
 [login]
+# Slurm login nodes. Combined control/login nodes are not supported.
 
 [control]
+# A single Slurm control node. Multiple (high availability) control nodes are not supported.
 
 [compute]
+# All Slurm compute nodes (in all partitions).
 
-[builder]
+[openhpc]
+# All Slurm nodes
+
+[cluster]
+# All nodes in the appliance - this allows for e.g. specific service nodes not running Slurm.
+# See default below.
 
 [cluster:children]
 login
 control
 compute
 
+[builder]
+# Host for (optionally) building compute images - see separate documentation.
+
 [podman:children]
+# Hosts running containers for below services:
 opendistro
 kibana
 filebeat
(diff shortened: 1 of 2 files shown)


environments/common/inventory/groups
```diff
@@ -1,49 +1,68 @@
...
 [mysql]
+# Runs database used for Slurm accounting.
 
 [node_exporter]
+# Monitor these hosts for hardware and OS metrics
 
 [openhpc_tests:children]
-cluster
+# Run post-deploy MPI-based tests using ansible/adhoc/test.yml
+cluster # TODO: FIXME in branch
```
jovial, line 64: Not clear what the fixme means. Can this one be removed now?
```suggestion
cluster
```

jovial (approved):

environments/common/inventory/groups
sjpb: Ah no it needs to be:
```ini
[openhpc_tests:children]
login
compute
```

cause we don't want the control node getting it's `opt/` overmounted (well, just don't want it involved at all really). @jovial you happy for me to make that change on this PR (not quite the right place but...)

jovial: Ahh, I see what you mean now. Did you intend to squash the changes? If so, I think should fix this in another PR so there isn't an unrelated change.

sjpb: Actually that fix is already done in #35 so I'm just going to leave the comment in here and merge.

PR #38, opened 2021-03-02
Add pam_slurm access control

sjpb:
Prevents non-privileged users logging in to compute nodes if they don't have a job running.

Note `centos` is not in `wheel` group so we allow both of those to log in without running jobs.


jovial:
This seems to be preventing the ansible_user from logging in to perform additional configuration:
```
user vagrant (uid=1000) has no active jobs on this node
```
see: https://github.com/stackhpc/openhpc-demo/pull/38/checks?check_run_id=2014487137#step:7:75

ansible/slurm.yml
```diff
@@ -31,3 +31,24 @@
...
+        insertafter: 'account\s+required\s+pam_nologin.so'
+        block: |
+          account    sufficient   pam_access.so
+          account    required     pam_slurm.so
+    - name: Configure login access control
+      blockinfile:
+        path: /etc/security/access.conf
+        block: |
+          +:wheel:ALL
+          +:centos:ALL
```
jovial, line 53:
```suggestion
          +:{{ ansible_user }}:ALL
```

sjpb:
> This seems to be preventing the ansible_user from logging in to perform additional configuration:
> 
> ```
> TASK [cloudalchemy.node-exporter : Get systemd version] ************************
> fatal: [testohpc-compute-1]: UNREACHABLE! => {
>     "changed": false,
>     "unreachable": true
> }
> 
> MSG:
> 
> Data could not be sent to remote host "127.0.0.1". Make sure this host can be reached over ssh: Warning: Permanently added '[127.0.0.1]:2202' (ECDSA) to the list of known hosts.
> 
> Access denied: user vagrant (uid=1000) has no active jobs on this node.
> Connection closed by 127.0.0.1 port 2202
> ```
> 
> see: https://github.com/stackhpc/openhpc-demo/pull/38/checks?check_run_id=2014487137#step:7:75

I'm not very clear how the ansible_user ends up being `vagrant` - looks like the README for this environment is outdated as it shows `centos`. Guess I'm a bit worried that other things also expect `centos` - e.g. the `sudoers` config does.

jovial:
> > This seems to be preventing the ansible_user from logging in to perform additional configuration:
> > ```
> > TASK [cloudalchemy.node-exporter : Get systemd version] ************************
> > fatal: [testohpc-compute-1]: UNREACHABLE! => {
> >     "changed": false,
> >     "unreachable": true
> > }
> > 
> > MSG:
> > 
> > Data could not be sent to remote host "127.0.0.1". Make sure this host can be reached over ssh: Warning: Permanently added '[127.0.0.1]:2202' (ECDSA) to the list of known hosts.
> > 
> > Access denied: user vagrant (uid=1000) has no active jobs on this node.
> > Connection closed by 127.0.0.1 port 2202
> > ```
> > 
> > 
> > see: https://github.com/stackhpc/openhpc-demo/pull/38/checks?check_run_id=2014487137#step:7:75
> 
> I'm not very clear how the ansible_user ends up being `vagrant` - looks like the README for this environment is outdated as it shows `centos`. Guess I'm a bit worried that other things also expect `centos` - e.g. the `sudoers` config does.

vagrant is the default login user for the vagrant images. Normally comes in via the inventory:
```
testohpc-control-0 ansible_ssh_host=127.0.0.1 ansible_ssh_user=vagrant ansible_ssh_common_args='-o StrictHostKeyChecking=no' ansible_ssh_private_key_file=/home/will/.vagrant.d/insecure_private_key ansible_ssh_port=2222
```
Although that reminds me that we have to handle `ansible_ssh_user` or `ansible_user`. Maybe we need something like:

```
ansible_ssh_user: "{{ ansible_ssh_user | default('centos') }}"
```

I think that this is a use-case we should support as it can't be guaranteed that centos is the login user for every image

sjpb:
Having looked at the vagrant docs, looks like it's just that vagrant uses the (depreciated) ansible_ssh_user? So could we just do:

```yaml
    - name: Configure login access control
      blockinfile:
        path: /etc/security/access.conf
        block: |
          +:wheel:ALL
          +:{{ ansible_ssh_user | ansible_user }}:ALL
          -:ALL:ALL
```

jovial:
> Having looked at the vagrant docs, looks like it's just that vagrant uses the (depreciated) ansible_ssh_user? So could we just do:
> 
> ```yaml
>     - name: Configure login access control
>       blockinfile:
>         path: /etc/security/access.conf
>         block: |
>           +:wheel:ALL
>           +:{{ ansible_ssh_user | ansible_user }}:ALL
>           -:ALL:ALL
> ```

Does that work in `ansible_ssh_user` is not defined?

sjpb:
> Does that work in `ansible_ssh_user` is not defined?

Fair point, I guess I meant `+:{{ ansible_ssh_user | defined(ansible_user) }}:ALL` but what I was trying to say was do we intend to fall back to `{{ ansible_user }}` not specifically `centos` if `ansible_ssh_user` isn't defined?

ansible/slurm.yml
```diff
@@ -31,3 +31,25 @@
...
+          account    sufficient   pam_access.so
+          account    required     pam_slurm.so
+    - name: Configure login access control
+      blockinfile:
+        path: /etc/security/access.conf
+        block: |
+          +:wheel:ALL
+          +:{{ ansible_ssh_user | default(ansible_user) }}:ALL
+          -:ALL:ALL
+      # vagrant uses (depreciated) ansible_ssh_user
```
jovial, line 55:
```suggestion
      # vagrant uses (deprecated) ansible_ssh_user
```

jovial (approved):

PR #40, opened 2021-03-04
Add adhoc play to restart slurm daemons

sjpb:
See title. Is helpful to include as is quite a common task.

### DIFF
--- a/ansible/adhoc/restart-slurm.yml
+++ b/ansible/adhoc/restart-slurm.yml
@@ -1,3 +1,9 @@
+# Restart all slurm daemons e.g. after changing configuration. Note that:
+# - `scontrol reconfigure` will handle most reconfiguration - see https://slurm.schedmd.com/scontrol.html#OPT_reconfigure
+#   for which options need a restart
+# - Adding or removing nodes by changing the `openhpc_` configuration and rerunning ansible/site.yml will automatically
+#   restart daemons as required.
+
 - hosts: compute,login
   become: yes
   gather_facts: no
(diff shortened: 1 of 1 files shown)


jovial:
CI failing as node-exporter role appears to have been removed from galaxy:

https://github.com/stackhpc/openhpc-demo/pull/40/checks?check_run_id=2029702226#step:5:204

Should be here: https://galaxy.ansible.com/cloudalchemy/node-exporter

Looks like we need to update to use an underscore.

ansible/adhoc/restart-slurm.yml
```diff
@@ -0,0 +1,28 @@
...
+# - Adding or removing nodes by changing the `openhpc_` configuration and rerunning ansible/site.yml will automatically
+#   restart daemons as required.
+
+- hosts: compute,login
+  become: yes
+  gather_facts: no
+  tasks:
+    - service:
+        name: slurmd
+        state: stopped
```
jovial, line 13:
```suggestion
        state: stopped

```

ansible/adhoc/restart-slurm.yml
```diff
@@ -0,0 +1,28 @@
...
+- hosts: control
+  become: yes
+  gather_facts: no
+  tasks:
+    - service:
+        name: slurmctld
+        state: restarted
```
jovial, line 20:
```suggestion
        state: restarted

```

jovial (commented):
Looks good. Would prefer a space between the plays in the playbook so it is easier to eyeball where they start and finish.

jovial (approved):

PR #42, opened 2021-03-05
[CI] Run vagrant workflow on PR

jovial:
This will check the merged result of the PR. This is useful when main is broken at the point of the branch. We can re-run the check without manually merging the target branch back into the source branch.

### DIFF
--- a/.github/workflows/vagrant.yml
+++ b/.github/workflows/vagrant.yml
@@ -1,6 +1,8 @@
 
 name: Test vagrant with ansible
-on: [push]
+on:
+  push:
+  pull_request:
 jobs:
   vagrant-example:
     runs-on: macos-latest


jovial:
I was hoping this meant that the pull_request workflow would be re-run if master was updated, but running "re-run all jobs" doesn't seem to be picking up the changes for master.

jovial:
Irritatingly the GITHUB_SHA on the pull request action (https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request) seems to be cached between re-runs. This means if you update master in the mean time, a re-run won't pull in those more recent commits. A workaround is close and re-open the pull request to update this.

jovial:
Merging as PR now green.

PR #44, opened 2021-03-10
Deploy rebuild machinery

sjpb:
Seems to have got missed in move to multi-environments.


sjpb:
Need to:
- fix validation so that if the role default for openhpc_rebuild_clouds is ok, that works
- fix CI so that `openstack token issue` "works" as in returns `rc=0`.

ansible/validate.yml
```diff
@@ -27,4 +27,26 @@
     - import_role:
         name: opendistro
         tasks_from: validate.yml
-      tags: validate
\ No newline at end of file
+      tags: validate
+
+- name: Validate rebuild configuration
+  hosts: openhpc
```
jovial, line 33: This looks a bit openstack specific. Is there a way we could only run this only if we are using openstack?

jovial: Yeah or possibly just a variable that sets the cloud type? E.g
````
appliance_cloud_type: openstack
```

ansible/validate.yml
```diff
@@ -27,4 +27,26 @@
...
+  tasks:
+    - block:
+      - stat:
+          path: "{{ openhpc_rebuild_clouds }}"
+        register: openhpc_clouds_path
+      - name: Check OpenRC file (openhpc_rebuild_clouds) exists
+        assert:
+          that: openhpc_clouds_path.stat.exists
+          fail_msg: "openrc file {{ openhpc_rebuild_clouds }} on localhost (specified by `openhpc_rebuild_clouds`) does not exist"
+      - name: Check OpenRC file (openhpc_rebuild_clouds) is usable
```
jovial, line 45: Add no_log here so that we don't leak a token in stdout.

sjpb:
Needs https://github.com/stackhpc/ansible_collection_slurm_openstack_tools/pull/26 and https://github.com/stackhpc/ansible_collection_slurm_openstack_tools/pull/25 merging before fixing this.

ansible/validate.yml
sjpb: All I can find on the [manpage](https://docs.openstack.org/python-openstackclient/latest/cli/man/openstack.html#manpage) is:
> --log-file <LOGFILE>¶
>  Specify a file to log output. Disabled by default.

ansible/validate.yml
sjpb: Have gone with a group as its consistent with all other features and there could be reasons why you only want to enable rebuild on a subset of nodes I guess.

ansible/validate.yml
sjpb: Validate tasks now defined in collection's role which does set this.

environments/common/layouts/everything
```diff
@@ -35,4 +35,7 @@ control
 control
 
 [filebeat:children]
-slurm_stats
\ No newline at end of file
+slurm_stats
+
+[rebuild:children]
+compute
```
jovial, line 41: Only issue here is that the vagrant example is using this file and it will fall the validation as it is not an openstack cloud, see: https://github.com/stackhpc/openhpc-demo/blob/c9fc16510513f1a655049c7576f9ce672a850f18/environments/vagrant-example/ansible.cfg#L8.

jovial:
Unsure why we hit: https://github.com/stackhpc/openhpc-demo/runs/2085005006#step:7:80 - ansible version related?

environments/common/layouts/everything
jovial: Now failing with this error: https://github.com/stackhpc/openhpc-demo/pull/44/checks?check_run_id=2085683879#step:7:106

sjpb: We've got some incompatible pieces here:
- Defining a group to turn on rebuild seems "right" as it's consistent with the way we do everything else
- Testing the `everything` template using vagrant seems right

Seems like there's a few approaches which would keep both of the above:
- Modify the inventory groups when running vagrant. Not easy to do directly
- Special-case the rebuild validate & deploy when running vagrant. Messy, and probably the problems above are really a specific case of a more general "we can't do this in vagrant" problem.
- Make vagrant pass the validation task - the "main" looks like it should be ok - might need `openhpc_rebuild_reconfigure: false`

jovial: > We've got some incompatible pieces here:
> 
> * Defining a group to turn on rebuild seems "right" as it's consistent with the way we do everything else
> * Testing the `everything` template using vagrant seems right
> 
> Seems like there's a few approaches which would keep both of the above:
> 
> * Modify the inventory groups when running vagrant. Not easy to do directly
> * Special-case the rebuild validate & deploy when running vagrant. Messy, and probably the problems above are really a specific case of a more general "we can't do this in vagrant" problem.
> * Make vagrant pass the validation task - the "main" looks like it should be ok - might need `openhpc_rebuild_reconfigure: false`

To me it seems like the right place to declare an openstack cloud is in the inventory generated by the provisioning script as it has all the info about how the machines were provisioned. We should only install the openstack rebuild script on an openstack cloud. We could do this with a feature flag or via computed inventory (put the hosts in openstack_rebuild if the cloud is openstack). What do you think?

dev/vagrant-example-configure.sh
```diff
@@ -18,6 +18,16 @@ echo "Running generate-passwords.yml"
...
 
+echo "Mocking openstack command for rebuild functionality"
+# we don't need to mock the cloud file as it only needs to exist
+mkdir /tmp/vagrant-example
+cat > /tmp/vagrant-example/openstack <<EOL
+#!/usr/bin/bash
+echo "fake openstack command"
+EOL
+chmod u+x /tmp/vagrant-example/openstack
+PATH=/tmp/vagrant-example:$PATH
```
jovial, line 29: I think you need to export this for it to be visible in the ansible process.

environments/common/layouts/everything
sjpb: Have implemented @jovial's proposed approach in c20faac.

jovial (approved):

PR #48, opened 2021-03-17
Reboot to change selinux state

sjpb:
Handle SELinux state changes which require reboots to take effect (e.g. enabled to disabled)

### DIFF
--- a/ansible/bootstrap.yml
+++ b/ansible/bootstrap.yml
@@ -11,7 +11,19 @@
 - hosts: selinux
   gather_facts: false
   become: yes
+  tags:
+    - selinux
   tasks:
-    - ansible.posix.selinux:
+    - name: Set SELinux state and policy
+      ansible.posix.selinux:
         state: "{{ selinux_state }}"
         policy: "{{ selinux_policy }}"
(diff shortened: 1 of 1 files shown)


jovial (commented):
Could we use `setenforce 0` to avoid the reboot?

jovial (approved):

sjpb:
> Could we use `setenforce 0` to avoid the reboot?

Hmm it'd neat - I guess only if reboot is required (then don't actually reboot). However on one environment it turned out setting permissive != disabled so this might give us a state which doesn't match the rebooted state. Maybe I'd rather take the hit on rebooting if its necessary for predictable behaviour.

jovial:
> > Could we use `setenforce 0` to avoid the reboot?
> 
> Hmm it'd be neat - I guess only if reboot is required (then don't actually reboot). However on one environment it turned out setting permissive != disabled so this might give us a state which doesn't match the rebooted state. Maybe I'd rather take the hit on rebooting if its necessary for predictable behaviour.

Good point, agreed that rebooting is the cleanest solution here.

PR #62, opened 2021-04-07
Use hostname instead of IP for default NFS server address

### LINKED ISSUE #61: NFS server specified by IP
Currently `environments/common/inventory/group_vars/all/nfs.yml` has:

```
nfs_server_default: "{{ hostvars[groups['control'] | first ].internal_address }}"
```

which means NFS breaks if the server IP moves (e.g. using a packer-built image but then rebuilding the cluster with TF).

NFS [allows](https://linux.die.net/man/5/nfs) an unqualified hostname so I wonder if we should switch to that? Presumably even in the case where internal DNS doesn't work we'll have an /etc/hosts so that should be ok?

sjpb:
Provide more flexibility by defining the default NFS mounts from the control host using its hostname not IP. This is in line with the approach used for specifying the slurm control host.

Fixes #61

### DIFF
--- a/environments/common/inventory/group_vars/all/nfs.yml
+++ b/environments/common/inventory/group_vars/all/nfs.yml
@@ -3,7 +3,7 @@
 # See: https://github.com/stackhpc/ansible-role-cluster-nfs
 # for variable definitions
 
-nfs_server_default: "{{ groups['control'] | first }}"
+nfs_server_default: "{{ hostvars[groups['control'] | first ].internal_address }}"
 
 nfs_configurations:
   - nfs_enable:
--- a/environments/common/inventory/group_vars/all/nfs.yml
+++ b/environments/common/inventory/group_vars/all/nfs.yml
@@ -3,7 +3,7 @@
 # See: https://github.com/stackhpc/ansible-role-cluster-nfs
 # for variable definitions
 
-nfs_server_default: "{{ hostvars[groups['control'] | first ].internal_address }}"
+nfs_server_default: "{{ groups['control'] | first }}"
 
 nfs_configurations:
   - nfs_enable:
--- a/environments/common/inventory/group_vars/all/defaults.yml
+++ b/environments/common/inventory/group_vars/all/defaults.yml
@@ -6,10 +6,10 @@ appliances_environment_root: "{{ lookup('env', 'APPLIANCES_ENVIRONMENT_ROOT') }}
 
 # Address(ip/dns) for internal communication between services. This is
 # normally traffic you do no want to expose to users.
-internal_address: "{{ ansible_host }}"
+internal_address: "{{ inventory_hostname }}"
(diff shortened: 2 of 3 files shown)


environments/common/inventory/group_vars/all/nfs.yml
```diff
@@ -3,7 +3,7 @@
 # See: https://github.com/stackhpc/ansible-role-cluster-nfs
 # for variable definitions
 
-nfs_server_default: "{{ hostvars[groups['control'] | first ].internal_address }}"
```
jovial, line 6: Should we default internal address to ansible_hostname? I had intended that to be set to either an ip address or hostname to control this kind of behaviour i.e it is the address used for internal communication beween servers e.g serverb connects to servera.internal_address, so by setting internal address we'd always use the hostname by default, but some could customise it use ansible_host if they didn't have working DNS.

sjpb: Might need an actual discussion to make sure I get the implications - in TF we also set some network vars, need to make sure we've got a consistent/docu'd approach here.

sjpb: So:
- Don't want to use `inventory_hostname` as this might not be routable w/o DNS
- Don't want to use `ansible_host` as this might be e.g. on management network not suitable for storage etc.
- TODO: check if `ansible_host` defaults to `inventory_hostname` if not defined
- Currently, TF tends to define:
 
      server_networks='${jsonencode({for net in control.network: net.name => [ net.fixed_ip_v4 ] })}'

   (based on p3-appliances). This is nice and general-purpose, e.g. for defining multihomed Lustre servers etc, but seems a bit complex.
- Considering defining `internal_address` defaulted to `inventory_hostname` in inventory. This could be overriden in environment's `hosts` file (e.g. with `ansible_host` or another network) if necessary.
- Needs documentation!

@jovial that summary ok? I will push an `internal_address` approach when I can.

jovial: Looks good to me. Agree that setting internal_address seems like the cleanest solution for now.

sjpb: > check if ansible_host defaults to inventory_hostname if not defined

Yes, it does.

sjpb: @jovial we also have `api_address` in `environments/common/inventory/group_vars/all/defaults.yml` which is currently defaulted to ansible_host - should we change this too?

sjpb:
Can't use `internal_address` for `openhpc_slurm_control_host` as there is a check that the latter is in `ansible_play_hosts` which is based on inventory_hostname. So, currently the slurm control host will be specified by inventory_hostname, i.e. we can't control (except by /etc/hosts) what interface this is contacted over.

sjpb:
@jovial ready for review

jovial (approved):

PR #63, opened 2021-04-07
Add packer VM groups variable

sjpb:
Currently packer can only build images for hosts in the `compute` group.

We need to be able to build images for different compute flavours (e.g. GPU)

We want to be able to build images for login and control hosts.

Currently has been tested for login hosts only.

### DIFF
--- a/packer/README.md
+++ b/packer/README.md
@@ -37,17 +37,22 @@ Steps:
         # or during development:
         PACKER_LOG=1 packer build --on-error=ask main.pkr.hcl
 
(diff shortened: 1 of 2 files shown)


jovial (approved):
Nice change :)

sjpb:
Not merging yet till I've had a chance to test building compute nodes.

sjpb:
Builds compute and login - can't build control nodes due to limits in the underlying openhpc role:
a) it expects to find the compute nodes to template out the slurm.conf, which fails in packer
b) if you remove the partition definitions to fix the above then it complains that all partitions are empty.

@jovial can you review now please?

PR #65, opened 2021-04-23
Add optional yum update step

sjpb:
Adds an optional rebuild step, which when enabled defaults to upgrading everything to latest - **including** the kernel.


sjpb:
Wonder if we should log what's been updated, using something like https://nbailey.ca/post/update-all-the-things/

jovial:
> Wonder if we should log what's been updated, using something like https://nbailey.ca/post/update-all-the-things/

Could login to the nodes and use `dnf history` as an alternative.

We do have the adhoc playbook: https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/adhoc/update-packages.yml, which has different behaviour i.e it updates the kernel. I wonder if they should be consistent. The other option is that we always use the adhoc playbook and do that before the packer build.

sjpb:
Will probably need a reboot too if changing the kernel - ideally only once.

sjpb:
> We do have the adhoc playbook: https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/adhoc/update-packages.yml, which has different behaviour i.e it updates the kernel. I wonder if they should be consistent. The other option is that we always use the adhoc playbook and run that in the packer build as well.

Somehow I had missed the adhoc..

1. However I think using it in ad-hoc mode is not really ideal for image build: I think that by default image build should update (all, or all except kernel*) update packages which come in the base image as a 1st step - everything installed subsequently by ansible will be `latest` anyway (unless specifically set not, for a reason). That seems better than building an entire environment, then updating part of it.

2. I do think we need an adhoc play too though, with some vars. Maybe we can use common code and call it from boootstrap.

@jovial thoughts on either?

jovial:
> > We do have the adhoc playbook: https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/adhoc/update-packages.yml, which has different behaviour i.e it updates the kernel. I wonder if they should be consistent. The other option is that we always use the adhoc playbook and run that in the packer build as well.
> 
> Somehow I had missed the adhoc..
> 
> 1. However I think using it in ad-hoc mode is not really ideal for image build: I think that by default image build should update (all, or all except kernel*) update packages which come in the base image as a 1st step - everything installed subsequently by ansible will be `latest` anyway (unless specifically set not, for a reason). That seems better than building an entire environment, then updating part of it.
> 2. I do think we need an adhoc play too though, with some vars. Maybe we can use common code and call it from boootstrap.
> 
> @jovial thoughts on either?

Yeah, I was thinking that the adhoc playbook could use the same variables as this update task albeit with `update_enable` always set to true. It is probably something you will want to run fairly regularly (if not using image based deploys) without having to run through the whole of site.yml. Or we could get rid of the adhoc playbook and just use tags? Just `-t update -e update_enable=true` is a little cumbersome.

### Alternative solution:

If `update-packages.yml` was updated to use the variables you added, could we create a wrapper playbook that did something like:

```
import_playbook: adhoc/update-packages.yml
import_playbook: site.yml
```

and then in the packer configuration:

```
...
playbook_file: "ansible/image-build.yml"
...
```
Where image-build.yml is the name of the wrapper.

sjpb:
Thanks.

- I agree an ad-hoc is better than tags
- I feel the wrapper solution is a bit messier - despite the fact that having a separate ad-hoc will duplicate stuff in boostrap too - that's only one task.

sjpb:
TODO: needs docs!

sjpb:
@jovial ready for review again I think please.

Added option to disable repos so e.g. could prevent openhpc updates entirely.

ansible/bootstrap.yml
```diff
@@ -22,14 +22,52 @@
...
+  tags:
+    - reboot
+  tasks:
+    - name: Check for pending reboot from package updates
+      stat: 
+        path: /var/run/reboot-required
+      register: update_reboot_required
+    - debug:
+        msg: "setstatus:{{ (sestatus.reboot_required | default(false)) }} packages: {{ (update_reboot_required.stat.exists | bool) }}"
+    - name: Reboot if required from SELinux state change or package upgrades
```
jovial, line 65: Nice work, mostly looks good to me. This is more of a thought than a suggestion, so feel free to shoot it down. Do you reckon this could be simplified if you use a handler? Could potentially notify on  `/var/run/reboot-required` existing or sestatus.reboot_required? Might need some cajoling on changed_when or a separate task.

jovial (approved):

sjpb:
Suggest:
1. `update facts` step should check for `sestatus.reboot_required` OR current check (experience from NREL)
2. Add a `post_reboot_delay` to reboot command too.

ansible/bootstrap.yml
sjpb: No as not inside a role.

PR #68, opened 2021-05-06
Add adhoc rebuild

sjpb:
Really useful for debugging/development, and also for situations where the slurm-controlled rebuild can't be used.


jovial (approved):
Seems useful - nice addition :)

PR #69, opened 2021-05-07
Use host networking for opendistro & kibana containers

sjpb:
Trying for fix for client multi-homed hosts


sjpb:
Not sure this is actually required, seemed to be an error with something else.

sjpb:
Closed as not required

PR #83, opened 2021-06-23
Update groups for common/everything/cookiecutter

### LINKED ISSUE #80: Revise cookiecutter groups
[These](https://github.com/stackhpc/ansible-slurm-appliance/blob/main/environments/skeleton/%7B%7Bcookiecutter.environment%7D%7D/inventory/groups) haven't been changed since the initial commit and e.g. specify the control node as being on the login node, which isn't recommended.

sjpb:
- Cookiecutter groups were very outdated - now linked to common/everything (fixes #80)
- Updated common/groups comments to e.g. clarify where single node required (based on monitoring docs)
- Updated common layout so that `cluster` group is equivalent to the `openhpc` group, or a superset of it if additional hosts added to group in a concrete environment. This is the design intent, so we might as well guarantee it.

### DIFF
--- a/environments/common/inventory/groups
+++ b/environments/common/inventory/groups
@@ -1,5 +1,5 @@
 [login]
-# Slurm login nodes. Combined control/login nodes are not supported.
+# All Slurm login nodes. Combined control/login nodes are not supported.
 
 [control]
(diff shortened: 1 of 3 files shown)


jovial (approved):
Looks good :+1:

PR #85, opened 2021-07-16
Remove duplicate content from skeleton README

sjpb:
This was based on the main README but was:
- outdated
- had errors in the install instructions
- contained terraform details which aren't necessarily applicable

The environment README should just contain any environment-specific instructions/documents, not duplicate the main README.


jovial (approved):
Seems good. Got a conflict now I merged the terraform example :(

brtknr:
The conflict undid what this commit tried to do so an easy one.

PR #89, opened 2021-07-23
Fix configdrive build without sudo rights

### LINKED ISSUE #88: Can't build configdrive if temporary file directory created by another user
Path here needs sudo to create:

https://github.com/stackhpc/ansible-slurm-appliance/blob/b11b2b2335b1590512b665b57f15e0d181a8c64b/packer/config-drive.yml#L16

sjpb:
Fixes #88

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,4 @@ terraform.tfstate*
 config-drive.iso
 venv
 *.pyc
-.terraform.lock.hcl
 packer/openhpc2
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ terraform.tfstate*
 config-drive.iso
 venv
 *.pyc
+.terraform.lock.hcl
+packer/openhpc2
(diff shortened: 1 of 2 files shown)


packer/config-drive.yml
```diff
@@ -13,7 +13,7 @@
       configdrive_fqdn: "test.example.com"
       configdrive_name: "openhpc2"
       configdrive_ssh_public_key: "{{ lookup('file', ssh_public_key_path) }}"
-      configdrive_config_dir: "/tmp/configdrive/"
+      configdrive_config_dir: "{{ playbook_dir }}"
```
jovial, line 16: Worth adding the generated files to gitignore?

sjpb: They get deleted after a successful run anyway

.gitignore
```diff
@@ -3,3 +3,5 @@ terraform.tfstate*
 config-drive.iso
 venv
 *.pyc
+.terraform.lock.hcl
```
jovial, line 6:
```suggestion
```
You are supposed to commit lock files since they contain the exact version of any dependencies (modules, providers) that you used.

jovial (approved):

PR #91, opened 2021-07-23
Fix tests mountpoint now /home is default NFS share

sjpb:
This got missed in #81 which changed the default NFS share to /home.

Not we can't use /home/centos/whatever as centos's home is local on /var/lib/centos.


sjpb:
Tagging Mat for review as he's testing this!

mkjpryor (approved):
All tests pass on CaaS cluster with equivalent change.

PR #92, opened 2021-07-23
Create production-safe tests

sjpb:
Currently the tests ad-hoc uses /opt NFS exports to avoid putting packages into the compute images. However this means you can't use it (e.g. for debugging) on an in-use cluster.

I propose that adding openhpc packages is OK, and we _default_ to installing the gcc+openmpi (w/ UCX) stack that provides.


sjpb:
@mkjpryor this is very much draft (& I need to rebase!) but might be useful ... only has pingmatrix at the moment.

sjpb:
This is really just a port of pingpong/pingmatrix from https://github.com/stackhpc/ansible_collection_slurm_openstack_tools/tree/main/roles/test, except that the appliance installs the required FOSS toolchain into the nodes, rather than the horrible nfs export of /opt.

There is no HPL yet, that'll be hard.

I think this is worth getting into the appliance ASAP personally, although whether we should remove the equivalent tests from the previous collection is up for discussion.

@jovial please can you check whether the `hpctests` groups (in common & everything) and related group vars are done appropriately?

sjpb:
@jovial probably also worth noting there's now `openhpc_default_packages` and `openhpc_extra_packages`. This actually means that current appliance configs which merge this (using only openhpc_packages) will have their CURRENT behaviour, i.e. not the new default packages, which is kind-of-nice, but more importantly means environments can add packages without having to remember to include the defaults too. Same approach you'd used elsewhere.

ansible/roles/hpctests/tasks/pingpong.yml
```diff
@@ -0,0 +1,41 @@
...
+    path: "{{ hpctests_rootdir }}/pingpong/{{ item }}"
+    state: touch
+  loop:
+    - pingpong.sh
+
+- name: Create sbatch script
+  template:
+    src: pingpong.sh.j2
+    dest: "{{ hpctests_rootdir }}/pingpong/pingpong.sh"
+
```
jovial, line 19: nit: Too many newlines
```suggestion
```

ansible/roles/hpctests/tasks/pingmatrix.yml
```diff
@@ -0,0 +1,56 @@
...
+- name: Expand node list
+  shell: "scontrol show hostnames {{ openhpc_tests_nodes if openhpc_tests_nodes is defined else (hpctests_computes.stdout_lines | join(',')) }}"
+  register: scontrol_hostnames
+- name: Create sorted node expression
+  shell: "scontrol show hostlistsorted {{ scontrol_hostnames.stdout_lines | join(',') }}"
+  register: scontrol_hostlistsorted
+- name: Expand node list again
+  shell: "scontrol show hostnames {{ scontrol_hostlistsorted.stdout_lines | join(',') }}"
+  register: slurm_names
+- name: Process output
```
jovial, line 40: nit: Not consistent with style of other files which use a newline between tasks

ansible/validate.yml
```diff
@@ -2,6 +2,14 @@
 
 # Fail early if configuration is invalid
 
+- name: Validate mysql configuration
```
jovial, line 5: Doesn't seem related to this change, but maybe is on top of another PR?

environments/common/inventory/group_vars/all/openhpc.yml
```diff
@@ -15,8 +15,11 @@ openhpc_slurmdbd_mysql_username: slurm
 openhpc_slurm_control_host: "{{ groups['control'] | first }}"
 openhpc_slurm_partitions:
   - name: "compute"
-openhpc_packages:
-  - slurm-libpmi-ohpc
+openhpc_default_packages:
+  - slurm-libpmi-ohpc # to allow intel mpi to work properly
+  - ohpc-gnu9-openmpi4-perf-tools # for hpctests
+openhpc_extra_packages: []
+openhpc_packages: "{{ openhpc_default_packages + openhpc_extra_packages}}"
```
jovial, line 22:
```suggestion
openhpc_packages: "{{ openhpc_default_packages + openhpc_extra_packages }}"
```

jovial (commented):
Mostly looks good to me. Can we add this to the end of the current CI job?

jovial:
> This is really just a port of pingpong/pingmatrix from https://github.com/stackhpc/ansible_collection_slurm_openstack_tools/tree/main/roles/test, except that the appliance installs the required FOSS toolchain into the nodes, rather than the horrible nfs export of /opt.
> 
> There is no HPL yet, that'll be hard.
> 
> I think this is worth getting into the appliance ASAP personally, although whether we should remove the equivalent tests from the previous collection is up for discussion.
> 
> @jovial please can you check whether the `hpctests` groups (in common & everything) and related group vars are done appropriately?

Looks sane to me.

sjpb:
> Mostly looks good to me. Can we add this to the end of the current CI job?

Done.

PR #97, opened 2021-08-18
Rename secrets

sjpb:
Changes names in the secrets generation code to reduce technical debt:

- The use of `openhpc` in the secrets generation code does not comply with our conventions as this is not related to the openhpc role.
- [Current ansible guidance](https://docs.ansible.com/ansible/latest/user_guide/playbooks_best_practices.html#keep-vaulted-variables-safely-visible) suggests using `vault_` as a prefix for vaulted vars.

Backwards compatibility is maintained so that a user's current secrets file will work with an appliance containing this PR.

Note that the actual var name of the `openhpc_passwords` -> `slurm_appliance_secrets` dict is almost irrelevant - this is not intended to be overriden at all, and is just used for templating. However the templating means the keys in that dict become ("top level") vars, so their name is important.


ansible/roles/passwords/defaults/main.yml
```diff
@@ -1,16 +1,14 @@
...
-openhpc_passwords:
-  secrets_openhpc_grafana_admin_password: "{{ secrets_openhpc_grafana_admin_password | default(lookup('password', '/dev/null')) }}"
-  secrets_openhpc_elasticsearch_admin_password: "{{ secrets_openhpc_elasticsearch_admin_password | default(lookup('password', '/dev/null')) }}"
-  secrets_openhpc_elasticsearch_kibana_password: "{{ secrets_openhpc_elasticsearch_kibana_password | default(lookup('password', '/dev/null')) }}"
-  secrets_openhpc_mysql_root_password: "{{ secrets_openhpc_mysql_root_password | default(lookup('password', '/dev/null')) }}"
-  secrets_openhpc_mysql_slurm_password: "{{ secrets_openhpc_mysql_slurm_password | default(lookup('password', '/dev/null')) }}"
-  # TODO: Might be cleaner to write this to a file, but needs modification of the role
-  secrets_openhpc_mungekey: "{{ secrets_openhpc_mungekey | default(secrets_openhpc_mungekey_default) }}"
+slurm_appliance_secrets:
+  vault_grafana_admin_password: "{{ secrets_openhpc_grafana_admin_password | default(vault_grafana_admin_password | default(lookup('password', '/dev/null'))) }}"
```
jovial, line 4: Just wondering if we need to set boolean=true[1] in second call to the default filter as the value coming out of the first will be the empty string i.e default(value, true). Although if you've tested this and it works ignore this comment.

[1] https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.default

ansible/bootstrap.yml
```diff
@@ -1,5 +1,15 @@
 ---
 
+- hosts: localhost
+  gather_facts: false
+  become: false
+  tags:
+    - depreciated
```
jovial, line 7:
```suggestion
    - deprecated
```

ansible/bootstrap.yml
```diff
@@ -1,5 +1,15 @@
...
+  tasks:
+    - debug:
+        msg: "Variables prefixed secrets_openhpc_* are depreciated and should be replaced with vault_*, please see ansible/roles/passwords/defaults/main.yml"
```
jovial, line 10:
```suggestion
        msg: "Variables prefixed secrets_openhpc_* are deprecated and should be replaced with vault_*, please see ansible/roles/passwords/defaults/main.yml"
```
Is it worth failing and advising to re-run the password generation script?

sjpb: If having to rerun that is OK, why do we need to support secrets_openhpc_* at all? I thought the idea was to support updating the appliance while using the existing environment's secrets file.

ansible/roles/passwords/defaults/main.yml
sjpb: Interesting. It does work as-is.

ansible/bootstrap.yml
jovial: Running the password generation script should rename the variables for you. If you don't run it, then you will still have, for example, ``secrets_openhpc_grafana_admin_password`` and not ``vault_grafana_admin_password`` defined (I think, unless I missed something).

sjpb: Ah, I'd missed the fact that wouldn't regenerate the actual secrets, that's neat.

jovial:
Need to update: `"{{ secrets_openhpc_mungekey | b64decode }}"`  in environments/common/inventory/group_vars/all/openhpc.yml

jovial:
```diff
diff --git a/environments/common/inventory/group_vars/all/openhpc.yml b/environments/common/inventory/group_vars/all/openhpc.yml
--- environments/common/inventory/group_vars/all/openhpc.yml
+++ environments/common/inventory/group_vars/all/openhpc.yml
@@ -19,7 +19,7 @@
   - slurm-libpmi-ohpc # to allow intel mpi to work properly
   - ohpc-gnu9-openmpi4-perf-tools # for hpctests
 openhpc_extra_packages: []
 openhpc_packages: "{{ openhpc_default_packages + openhpc_extra_packages }}"
-openhpc_munge_key: "{{ secrets_openhpc_mungekey | b64decode }}"
+openhpc_munge_key: "{{ vault_openhpc_mungekey | b64decode }}"
 openhpc_slurm_configless: true
 openhpc_login_only_nodes: login
\ No newline at end of file

```

jovial:
```diff
diff --git a/environments/common/inventory/group_vars/all/openhpc.yml b/environments/common/inventory/group_vars/all/openhpc.yml
--- environments/common/inventory/group_vars/all/openhpc.yml
+++ environments/common/inventory/group_vars/all/openhpc.yml
@@ -19,7 +19,7 @@
   - slurm-libpmi-ohpc # to allow intel mpi to work properly
   - ohpc-gnu9-openmpi4-perf-tools # for hpctests
 openhpc_extra_packages: []
 openhpc_packages: "{{ openhpc_default_packages + openhpc_extra_packages }}"
-openhpc_munge_key: "{{ secrets_openhpc_mungekey | b64decode }}"
+openhpc_munge_key: "{{ vault_openhpc_mungekey | b64decode }}"
 openhpc_slurm_configless: true
 openhpc_login_only_nodes: login
\ No newline at end of file

```

jovial:
```diff
diff --git a/environments/common/inventory/group_vars/all/openhpc.yml b/environments/common/inventory/group_vars/all/openhpc.yml
--- environments/common/inventory/group_vars/all/openhpc.yml
+++ environments/common/inventory/group_vars/all/openhpc.yml
@@ -19,7 +19,7 @@
   - slurm-libpmi-ohpc # to allow intel mpi to work properly
   - ohpc-gnu9-openmpi4-perf-tools # for hpctests
 openhpc_extra_packages: []
 openhpc_packages: "{{ openhpc_default_packages + openhpc_extra_packages }}"
-openhpc_munge_key: "{{ secrets_openhpc_mungekey | b64decode }}"
+openhpc_munge_key: "{{ vault_openhpc_mungekey | b64decode }}"
 openhpc_slurm_configless: true
 openhpc_login_only_nodes: login
\ No newline at end of file

```

jovial:
```diff
diff --git a/environments/common/inventory/group_vars/all/openhpc.yml b/environments/common/inventory/group_vars/all/openhpc.yml
--- environments/common/inventory/group_vars/all/openhpc.yml
+++ environments/common/inventory/group_vars/all/openhpc.yml
@@ -19,7 +19,7 @@
   - slurm-libpmi-ohpc # to allow intel mpi to work properly
   - ohpc-gnu9-openmpi4-perf-tools # for hpctests
 openhpc_extra_packages: []
 openhpc_packages: "{{ openhpc_default_packages + openhpc_extra_packages }}"
-openhpc_munge_key: "{{ secrets_openhpc_mungekey | b64decode }}"
+openhpc_munge_key: "{{ vault_openhpc_mungekey | b64decode }}"
 openhpc_slurm_configless: true
 openhpc_login_only_nodes: login
\ No newline at end of file

```

jovial (approved):

PR #98, opened 2021-08-18
Add fail2ban functionality

sjpb:
This adds a fail2ban role, configured to protect ssh. No email alerts are set up so logs will have to be manually reviewed.

In the `everything` layout the `fail2ban` group is `login`.


ansible/roles/fail2ban/meta/main.yml
```diff
@@ -0,0 +1,45 @@
+galaxy_info:
+  author: your name
```
jovial, line 2: Fill this in.

ansible/roles/fail2ban/README.md
```diff
@@ -0,0 +1,44 @@
+Role Name
```
jovial, line 1:
```suggestion
fail2ban
```

ansible/roles/fail2ban/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+
+- name: Install fail2ban packages
+  package:
+    name:
+      - fail2ban-server
+      - fail2ban-firewalld
+    state: present
+
+- name: Allow cluster subnet through filewalld
+  ansible.posix.firewalld:
```
jovial, line 14: Personally, I think it would be cleaner if we had a separate role to configure the firewall, otherwise it can be tough to determine where all these rules came from. Could make the firewalld role run on hosts where fail2ban is enabled.

sjpb: I need to investigate this on a clean host. ISTR there was a problem that fail2ban configured firewalld too restrictively, which messed things up as soon as you started fail2ban, hence this approach. Putting to draft till resolved.

ansible/roles/fail2ban/tasks/main.yml
```diff
@@ -10,12 +10,8 @@
...
-- name: Allow cluster subnet through filewalld
-  ansible.posix.firewalld:
-    permanent: yes
-    source: "{{ fail2ban_cluster_subnet }}"
-    zone: "{{ fail2ban_cluster_subnet_zone }}"
-    state: enabled
+- name: Apply filewalld configs
+  ansible.posix.firewalld: "{{ item }}"
   notify: Restart filewalld
   loop: "{{ fail2ban_firewalld_configs }}"
```
sjpb, line 16: @jovial note the existing look here is borked, it just applies the same config multiple times!

ansible/roles/fail2ban/tasks/main.yml
sjpb: @jovial as fail2ban installs firewalld it's hard to add a separate role for firewalld, as it would need to do:
- install and configure fail2ban (fail2ban role)
- configure and start firewalld (firewalld role)
- start fail2ban (either fail2ban role, or externally but that seems a bit odd)

However I do agree firewalld config needs to live in one place and be flexible, just hit the fact that OOD needs this configuring as well if you're running fail2ban and OOD on the login node.

What about this new approach in 08589c6?

sjpb:
Base branches changed to refresh github diff. @jovial I'd appreciate re-review of this.

ansible/roles/fail2ban/tasks/main.yml
jovial: Couldn't you just make the firewalld role run first? Just seems like the role is doing two things.

sjpb:
Ok @jovial I've added a `firewalld` role, and set the default group for it to the `fail2ban` group.

I'm not totally convinced by the name `cluster_subnet_cidr` (defines range to allow all ports through firewalld) but it does at least match the `cluster` group name. What do you think?

sjpb:
Think CI just ran out of hosts on smslabs.

sjpb:
Almost certainly need to add https service to enable OOD

sjpb:
@jovial ready for review again. After discussion with Stig changed firewalld to default to be permissive, as relying on the openstack security groups is considered better and it means the cluster's network range doesn't need to be specified.

jovial (approved):
I like how this looks now, thanks :)

PR #99, opened 2021-08-18
Add hardening functionality

sjpb:
This uses the openstack [ansible-hardening](https://github.com/openstack/ansible-hardening) role to apply STIG security configurations to hosts in the `hardening` group. The `environments/common/layouts/everything` template used by cookiecutter puts the `login` group into the `hardening` group.

**NB:** This depends on PR #97.

The `ansible-hardening` role is run with default settings and the following additional changes are made to hosts in the `hardening` group:
- "FIPS mode" for cryptography standards is enabled (ideally this should actually be enabled during system installation but this is clearly not possible using ansible).
- A password is autogenerated for the `centos` user by the `ansible/adhoc/generate-passwords.yml ` playbook and sudo rules are set for for `centos` and `wheel` groups to require a password. Note if other users have been added with sudo rights (e.g. via a `pre-` hook), they will not be affected and a warning will be generated by the `ansible-hardning` role.

Note that hosts in `hardening` will also have their `sestatus` set to enforcing, rather than the appliance default of `permissive`.

On a CentOS 8.4 system the `ansible-hardening` role will still generate warnings about:
- Users without a home directory (`cockpit-ws`, `cockpit-wsinstance`, `rngd` are missing theirs - note the cockpit-* ones are not removed by removing the `cockpit-ws` package).
- Various directories which should be on their own filesystem: `/var`, `/var/log/audit`, `/tmp`.
- "Output from syslog must be sent to another server." (STIG V-72209).

### DIFF
--- a/ansible/bootstrap.yml
+++ b/ansible/bootstrap.yml
@@ -11,6 +11,15 @@
       become_flags: '-i'
       become: true
 
+- hosts: fail2ban
+  gather_facts: false
+  become: yes
+  tags:
+    - fail2ban
+  tasks:
+    - import_role:
+        name: fail2ban
+
 - hosts: selinux
   gather_facts: false
   become: yes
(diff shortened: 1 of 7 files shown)


sjpb:
- TODO: Need to check this can run MPI jobs OK with sestatus = enforcing
- TODO: Add README

sjpb:
Interestingly `ansible/adhoc/tests.yml` fails when login is in `hardening` node with:

```
package intel-... does not verify: no digest
```

The workaround is to set `openhpc_slurm_login: "{{ groups['compute'][0] }}` in the environment's `openhpc_tests` overrides.

PR #103, opened 2021-08-27
Add HPL to hpctests role

sjpb:
This provides HPL by building it using an openhpc-provided openblas (which is added to the appliance default packages).

Unlike https://github.com/stackhpc/ansible_collection_slurm_openstack_tools/tree/main/roles/test this HPL won't require Intel processors. However that means we can't use TBB threads either so have to be cleverer about paralleisation and targeting N.

We'll use 1x process per physical core. and for hpl-all I think it should target either a near-square PQ or P=nodes Q=cores-per-node (or vice versa as needed to get P < Q).


sjpb:
Back-to-back testing on alaska showed ~4% less performance than the [intel implementation](https://github.com/stackhpc/ansible_collection_slurm_openstack_tools/tree/main/roles/test). That's better than I was hoping for and shows we're getting sufficient performance out of the parallelization strategy.

environments/vagrant-example/activate
```diff
@@ -1,6 +1,9 @@
 export APPLIANCES_ENVIRONMENT_ROOT=$(dirname $(realpath ${BASH_SOURCE[0]:-${(%):-%x}}))
 echo "Setting APPLIANCES_ENVIRONMENT_ROOT to $APPLIANCES_ENVIRONMENT_ROOT"
 
+APPLIANCES_ENVIRONMENT_NAME=$(basename $APPLIANCES_ENVIRONMENT_ROOT)
```
jovial, line 4: nit: unrelated to change

jovial (approved):

environments/vagrant-example/activate
sjpb: True, brings vagrant into line with cookiecutter though.

PR #104, opened 2021-09-01
Provide control over packer base image

sjpb:
Currently this is hardcoded to centos 8.2. This PR adds packer variables to allow this to be altered, and changes default to 8.4.

Although I think we should continue to push for appliance users to start from a GenericCloud image so that everything is defined in the code, this does allow for cases where client wants/is required to to use their own base image too.


jovial (approved):
Looks good.

PR #106, opened 2021-09-01
Add basic user handling

sjpb:
Provides a basic IAM capability via ansible. See [ansible/roles/basic_users/README.md](https://github.com/stackhpc/ansible-slurm-appliance/blob/feature/basic_users/ansible/roles/basic_users/README.md) for features/why this isn't totally straightforward.

Unlike the p3-appliances implementation, the datastructure used to define users etc is passed through to ansible's user module, so supports anything that module does.

Name chosen as `local_users` is already used for `podman` etc system users.


ansible/roles/basic_users/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+    label: "{{ item.name }}"
+  when:
+    - "item.state | default('present') == 'absent'"
+  
+- name: Create users and generate public keys
+  user: "{{ basic_users_userdefaults | combine(item) | remove_keys(['public_key']) }}"
+  loop: "{{ basic_users_users }}"
+  loop_control:
+    label: "{{ item.name }} [{{ item.state | default('present') }}]"
+  register: basic_users_info
```
jovial, line 16: Could you compute this once on localhost?

ansible/roles/basic_users/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+    - basic_users_manage_homedir
+  run_once: true
+
+- name: Write generated public key as authorized for SSH access
+  authorized_key:
+    user: "{{ item.name }}"
+    state: present
+    manage_dir: no
+    key: "{{ item.ssh_public_key }}"
+  loop: "{{ hostvars[ansible_play_hosts | first].basic_users_info.results }}"
```
jovial, line 38: I think `basic_users_info` is set as a fact on all hosts, right?

ansible/roles/basic_users/library/terminate_user_sessions.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/python
+
+# Copyright: (c) 2021, Steve Brasier <steveb@stackhpc.com>
+# Apache V2 licence
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
```
jovial, line 8: Update or remove the docs.

ansible/roles/basic_users/filter_plugins/remove.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class FilterModule(object):
+
+    def filters(self):
+        return {
+            'remove_keys': self.remove_keys
+        }
+
+    def remove_keys(self, orig_dict, keys_to_remove):
```
jovial, line 13: remove suggests that the original dict is modified. Would filter_keys be a better name?

ansible/roles/basic_users/tasks/main.yml
sjpb: Yes but the key is only generated on the first one (by design), so it is only present in the facts for that host.

jovial (approved):
All comments have now been addressed, looks good :)

PR #107, opened 2021-09-01
Manage block devices

sjpb:
Adds `block_devices` functionality to the `filesystems.yml` playbook to make adding e.g. Cinder volumes easier. See [ansible/block_devices/README](https://github.com/stackhpc/ansible-slurm-appliance/blob/feature/block_devices/ansible/roles/block_devices/README.md) for details.


environments/common/inventory/group_vars/builder/defaults.yml
```diff
@@ -6,3 +6,6 @@ openhpc_slurm_service_started: false
 nfs_client_mnt_state: present
 openhpc_rebuild_reconfigure: false
 update_enable: true
+block_devices_partition_state: skip
```
jovial, line 9: Does this mean that we need to run ansible against any new node we enroll? Even if we boot from the packer created image?

ansible/roles/block_devices/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+This creates 'ext4' partitions on `/dev/sdb1` on `server` and `/dev/sdc1` on `server2`, both mounted at `/mnt/files`:
+
+```yaml
+block_devices_fstype: ext4
+block_devices_path: /mnt/files
+block_devices_number: 1
+block_devices_configurations:
+  - device: /dev/sdb
+    hostnames: server1
```
jovial, line 78: Would it make more sense to use host_vars/group_vars to configure `block_devices_configurations` to be different for a particular host/group of hosts?

environments/common/inventory/group_vars/builder/defaults.yml
sjpb: Well, if we're booting a new node with a packer-built image, that boot process can't create the partition. So the only sensible thing to do is skip, I think?

For cinder volumes that's fine as we can run the ansible first and then the packer-built image just has a directory in the right place and fstab configured to enable the volume to be mounted.

ansible/roles/block_devices/README.md
sjpb: Yes, it would. I guess it means that `block_devices` variables aren't set in the `block_devices` group_vars, but you're right that's much more sensible.

environments/common/inventory/group_vars/builder/defaults.yml
jovial: I see; I guess the alternative would be to some sort of cloud-init script or similar to setup these block devices on first boot, but I'm not sure if there is a requirement to run the ansible against a new node already?

jovial (approved):
Looks sensible.

Edit: just seen the conflicts, but change looks good :+1:

PR #109, opened 2021-09-09
Add podman_cidr to allow changing podman network range

sjpb:
By default rootless podman containers use a network of `10.0.2.0/24` - see `slirp4netns` option in `podman run` [docs](url). If actual host networking overlaps with this, containers can't communicate.

This provides a) validation that the host's (default) IP address isn't in the default podman namespace and b) a `podman_cidr` variable to change it if it is.

If we need to set which IP the podman network is bound to using the default IP for the validation would also have to be more sophisticated, but this change is worthwhile as-is.


jovial (approved):
Nice validation check  :). EDIT: Oops missed the fact that CI failed the check!

sjpb:
@jovial I need to check why this is failing (on the validation) in CI, when we haven't had problems with Kibana in CI....

```shell
Default ipv4 address 10.0.2.15 for testohpc-control-0 is in podman network range 10.0.2.0/24 - set `podman_cidr` to avoid host network address range
```

jovial:
> @jovial I need to check why this is failing (on the validation) in CI, when we haven't had problems with Kibana in CI....
> 
> ```shell
> Default ipv4 address 10.0.2.15 for testohpc-control-0 is in podman network range 10.0.2.0/24 - set `podman_cidr` to avoid host network address range
> ```

Is it because we use another address for the internal communication: https://github.com/stackhpc/ansible-slurm-appliance/blob/main/environments/vagrant-example/inventory/group_vars/all/overrides.yml#L3. That corresponds to private_network in the vagrant file (192.168.56.0/24). So maybe we should be checking that internal_address doesn't conflict?  That is the address services use for internal communication.

sjpb:
Ah it's also `api_address`. But then that doesn't seem to be used anywhere ...

sjpb:
Duh, it is. Normally `api_address` & `internal_address` will be the hostname though so it's no good checking those in general.

sjpb:
65130b6 is maybe a bit nuclear but should avoid hitting hard-to-diagnose problems in the future - I think having podman overlapping with _any_ real networks is asking for trouble.

jovial:
> [65130b6](https://github.com/stackhpc/ansible-slurm-appliance/commit/65130b6d820a9c841f8f6e1b645e233d1bc29869) is maybe a bit nuclear but should avoid hitting hard-to-diagnose problems in the future - I think having podman overlapping with _any_ real networks is asking for trouble.

Nice, I think you are right. Looks good to merge.

jovial (approved):

PR #113, opened 2021-09-17
Fix idempotency when running 'slurm.yml --tags openhpc'

### LINKED ISSUE #112: openhpc tags not idempotent
If you run `ansible-playbook ansible/slurm.yml --tags openhpc` slurm isn't left in the same state it started, as the rebuild/reboot stuff doesn't get run and that modifies slurm.conf. Need to add the appropriate tag

sjpb:
Fixes #112

### DIFF
--- a/ansible/slurm.yml
+++ b/ansible/slurm.yml
@@ -31,6 +31,7 @@
   become: yes
   tags:
     - rebuild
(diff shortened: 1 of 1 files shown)


jovial (approved):
Good catch

PR #120, opened 2021-09-30
Don't add centos vault now openhpc v2.3 released to fix centos 8.4 pa…

sjpb:
Centos Vault was added to when centos 8.4 came out as it changed the `hwloc` package, which was depended on by openhpc packages. 

However now openhpc v2.3 is released any openhpc install will get the rebuild openhpc packages.

### DIFF
--- a/ansible/slurm.yml
+++ b/ansible/slurm.yml
@@ -15,14 +15,6 @@
   tags:
     - openhpc
   tasks:
-    - name: Add CentOS 8.3 Vault repo for OpenHPC hwloc dependency
-      # NB: REMOVE THIS once OpenHPC works on CentOS 8.4
-      yum_repository:
-        name: vault
-        file: CentOS-Linux-Vault8.3
-        description:  CentOS 8.3 packages from Vault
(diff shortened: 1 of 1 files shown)


sjpb:
Openhpc issue: https://github.com/openhpc/ohpc/issues/1324

sjpb:
I'm not convinced that this is OK if running a centos 8.3 or earlier image. Maybe that doesn't matter too much though?

ansible/slurm.yml
```diff
@@ -15,14 +15,6 @@
   tags:
     - openhpc
   tasks:
-    - name: Add CentOS 8.3 Vault repo for OpenHPC hwloc dependency
-      # NB: REMOVE THIS once OpenHPC works on CentOS 8.4
-      yum_repository:
```
jovial, line 20: Consider using: `state: absent` to remove this file on hosts upgraded from a previous version.

sjpb:
Will be done by https://github.com/stackhpc/ansible-slurm-appliance/pull/134 for Rocky 8.5 anyway.

PR #127, opened 2021-10-14
Update docs for prerequisites, adhoc plays and new functionality

### LINKED ISSUE #126: Update prerequisites
Should also list:
- Internet access from all hosts (but note e.g. proxies can be configured by the appliance)
- Ability to inject ssh keys (e.g. via cloud init)
- Root access on the ansible deploy host

sjpb:
Docs are a bit outdated. Also fixes #126

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,12 @@ While it is tested on OpenStack it should work on any cloud, except for node reb
 
 ## Prerequisites
 It is recommended to check the following before starting:
+- You have root access on the "ansible deploy host" which will be used to deploy the appliance.
 - You can create instances using a CentOS 8 GenericCloud image (or an image based on that).
 - SSH keys get correctly injected into instances.
 - Instances have access to internet (note proxies can be setup through the appliance if necessary).
 - DNS works (if not this can be partially worked around but additional configuration will be required).
-- You have root access on the "ansible deploy host" which will be used to deploy the appliance.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,14 +7,20 @@ This repository contains playbooks and configuration to define a Slurm-based HPC
...
 The repository is designed to be forked for a specific use-case/HPC site but can contain multiple environments (e.g. development, staging and production). It has been designed to be modular and extensible, so if you add features for your HPC site please feel free to submit PRs back upstream to us!
 
-## Pre-requisites
+While it is tested on OpenStack it should work on any cloud, except for node rebuild/reimaging features which are currently OpenStack-specific.
 
-- Working DNS so that we can use the ansible inventory name as the address for connecting to services.
-- Bootable images based on Centos 8 Cloud images.
+## Prerequisites
+It is recommended to check the following before starting:
+- You have root access on the "ansible deploy host" which will be used to deploy the appliance.
```
jovial, line 18: Why do we need root on the control host?

jovial (approved):
Seem like good improvements.

README.md
sjpb: - Packer build instructions require root for package installs on localhost.
- [Validation for app creds](https://github.com/stackhpc/ansible_collection_slurm_openstack_tools/blob/main/roles/rebuild/tasks/validate.yml#L28) requires sudo on localhost, doesn't work properly if it's not run with sudo, appears to be an `openstack.cloud.auth` module bug but 🤷.

PR #131, opened 2021-11-30
Fix block_devices to not rely on device path

sjpb:
The `block_devices` role currently uses the device path (e.g. /dev/sdb) to identify the device to create fileystems on and the fileystem (e.g. /dev/sdb1) to mount. This is unsafe as device letters can/do change after reboot.

This PR changes the role to:
1. Use serial numbers to identify the device to create a fileysystem on.
2. Specify mounts using filesystem UUIDs.

Note that for openstack volumes the serial number is set to the volume ID hence can be retrieved from e.g. terraform.


ansible/roles/block_devices/library/block_devices.py
```diff
@@ -0,0 +1,44 @@
...
+
+import json
+
+from ansible.module_utils.basic import AnsibleModule
+
+def run_module():
+    module_args = dict()
+    module = AnsibleModule(argument_spec=module_args, supports_check_mode=True)
+    result = {"changed": False}
+    _, stdout, _ = module.run_command("lsblk --paths --json -O", check_rc=True)
```
sjpb, line 33: NB `json` only works on CentOS 8 but that's ok

sjpb: And so does `-O`

sjpb: Something like `lsblk --paths --list -o name,serial` works on CentOS7 (although obvs format is different)

sjpb:
Reviews pending since December 2021, passes CI so merging anyway.

PR #136, opened 2022-01-12
Enable build of control images

sjpb:
Other things needed for this:
  - [ ] Fix failure on `TASK [stackhpc.openhpc : Fail if openhpc_slurm_control_host or openhpc_cluster_name or openhpc_slurm_partitions are undefined]` (b/c we've emptied openhpc_slurm_partitions) - possibly this is OK if using the `feature/autoscale` branch of this role anyway?


sjpb:
Replaced by #160

PR #137, opened 2022-01-12
Add openstack-based CI

sjpb:
Current Vagrant-based CI is problematic:
- Has to run on OSX to get nested virt, which uses credits
- Uses vagrant boxes rather than genericcloud images, which means CI != production
- Vagrant and Virtualbox can be complex to set up/debug
- Very hard to debug virtualbox VMs on gitlab

This implementation:
- [x] Adds an smslabs-example environment with TF to setup a 2-compute-node cluster (as per vagrant)
- [ ] Adds a workflow to provision, configure and test using this environment

Also requires:
- [x] A `slurm-app-ci` bastion user
- [x] App cred for `slurm-app-ci` user
- [x] OpenStack keypair for `slurm-app-ci` user.


jovial (commented):
Looks good. Couple of points:

- Can we stop CI running on external PRs (they won't have access to the secrets)?
- Is it worth mentioning your suggested workflow for getting an external PR merged somewhere?

sjpb:
> Can we stop CI running on external PRs (they won't have access to the secrets)?

"Fork pull request workflows from outside collaborators" is already set to "Require approval for all outside collaborators", so I think that deals with that. The default is to only require approval for first-time contributors.

jovial (approved):
OK, I'm happy, thanks for answering my questions :)

PR #138, opened 2022-01-18
Add OpenStack-based Packer build

sjpb:
- Adds packer-based build of both login & compute images, simultaneously
- Adds example smslabs config for this
- Add above builds to CI on smslabs
- Tests reimage of compute (via Slurm) and login (via OpenStack) in CI


sjpb:
Superseded by refactored version in #140

PR #140, opened 2022-01-26
Add OpenStack-based Packer build with CI for build + reimage

sjpb:
- Adds packer build using OpenStack of both login & compute images, simultaneously, and docs
- Adds example smslabs config for image build
- Adds CI on smslabs for image build and reimage

Plus various fixes exposed by CI:
- Ensure control node is up, always (needed for both direct ansible deploy and image builds).
- Fix various stages timing out in slow network/compute CI environment.
- Fix image deploy failure by rolling back rebuild role changes introduced for autoscaling (changes are valid, but need changes in appliance too).


jovial (approved):
Looks good, probably worth a squash.

PR #144, opened 2022-02-02
Skip some hpctests when not enough compute nodes

sjpb:
pingpong and pingmatrix don't make sense with < 2 compute nodes and cause an error, so skip them if this is the case.


m-bull:
Works in the single-compute-node case via Azimuth

m-bull (approved):

PR #147, opened 2022-02-15
Add blackbox probing of slurm endpoints

sjpb:
Add probing of slurmctld and slurmd ports using the prometheus blackbox exporter. Will allow monitoring/alerting on slurm{ctld,d} being up/reachable.

"Default" (= `everything` layout) behaviour is to put the blackbox exporter on the prometheus node, on the basis that that removes a failure point. Is that appropriate?

Note there are slurm exporters for prometheus but they seem slightly fragile/dependent on specific versions. This is anyway more at the infra layer so probably still useful.

TODO:
- [ ] Expand slurmd endpoints to all compute/login nodes.
- [ ] Label jobs appropriately.
- [ ] Create dashboard

Could also add (probably as another job) blackbox probing for e.g. DNS?

### DIFF
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -61,6 +61,12 @@
   tasks:
     - import_role: name=cloudalchemy.node_exporter
 
+- name: Deploy blackbox exporter
+  hosts: blackbox_exporter
+  tags: blackbox_exporter
+  tasks:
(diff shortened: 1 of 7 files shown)


elelaysh:
Also need to add alerts for certificate expiry (`expr: probe_ssl_earliest_cert_expiry - time() < 7*24*3600` or similar).

PR #148, opened 2022-02-17
Fix accidently not using configless mode

sjpb:
The smslabs (CI) environment overrode `openhpc_config` which override the appliance default of configless. We always want to operate in this mode, so this PR a) fixes the problem in that environment, b) checks that the environment hasn't turned it off.


sjpb:
Not required in `main`, appliance only uses `openhpc_config` in feature/autoscale

PR #155, opened 2022-03-10
Fix issue with templating hostvars

sjpb:
Currently, this fails:

    ansible all -m debug -a msg="{{ hostvars[inventory_hostname] }}"

whereas this succeeds:

    ansible all -m debug -a msg="{{ <any hostvar> }}"

Turns out this is because of the use of `context['hostvars']` in [this helper](https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/filter_plugins/utils.py#L12) called from the `prometheus_node_exporter_targets` filter [source](https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/filter_plugins/utils.py#L23) used in [environments/common/group_vars/all/prometheus](https://github.com/stackhpc/ansible-slurm-appliance/blob/c7274e91758f8cc10909185acb52185935768653/environments/common/inventory/group_vars/all/prometheus.yml#L16).

The difference is the first one calls for resolving hostvars in the jinja template whereas the second doesn't - presumably this somehow trips up the context as it is required "too early".

The fix is just to pass in the `env` var directly, instead of getting it from hostvars. Results in simpler code too.

### DIFF
--- a/ansible/filter_plugins/utils.py
+++ b/ansible/filter_plugins/utils.py
@@ -9,22 +9,10 @@ import jinja2
 from ansible.module_utils.six import string_types
 import os.path
 
-def _get_hostvar(context, var_name, inventory_hostname=None):
-    if inventory_hostname is None:
-        namespace = context
-    else:
-        if inventory_hostname not in context['hostvars']:
-            raise AnsibleFilterError(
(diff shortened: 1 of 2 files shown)


jovial (approved):
Nice fix

PR #156, opened 2022-03-10
Allow openondemand group to not exist and validation to still run

m-bull:
The `validate.yml` playbook fails with a keyerror when an `openondemand` group doesn't exist. This adds a further check to ensure that the group does exist before checking to see if it contains any hosts.

Tested and working with Azimuth clusters.

### DIFF
--- a/ansible/validate.yml
+++ b/ansible/validate.yml
@@ -59,7 +59,9 @@
     - import_role:
         name: openondemand
         tasks_from: validate.yml
-      when: groups['openondemand'] | length > 0
+      when: 
+        - "'openondemand' in groups"
+        - groups['openondemand'] | length > 0
       tags:
         - openondemand
         - openondemand_server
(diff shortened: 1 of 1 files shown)


sjpb:
I don't understand this, the play runs on the `ondemand` group so if that's empty the validation doesn't happen. The group must always exist as its defined (empty) in environments/common/inventory/groups. This convention of functionallity being enabled by populating the empty group is the same throughout the appliance, eg nfs, grafana, etc.

If this is a CaaS thing then the group should always be defined, empty, in however the CaaS appliance defines groups.

sjpb:
Ah now I can see this properly (not on a mobile) I can see the real issue. We're trying to check that `openondemand_servername != '' on both:
- `openondemand` hosts,
- `grafana` hosts

But the latter is only relevant if `openondemand` is enabled, yet will always run if `grafana` is enabled.

You can't fix this with host patterns (as the grafana and ondemand hosts aren't the same), so your proposal is correct. Would be nice to describe this in a `#` comment though, it is not obvious why the normal pattern of just assuming host groups exist and are non-empty if relevant isn't enough.

sjpb (approved):

PR #160, opened 2022-03-30
Enable build of environment-specific control images

### LINKED ISSUE #133: Building images for control nodes
At least once the changes for autoscaling are merged into the `openhpc` role, it should be able to build images for control nodes as a long as `openhpc_partitions` is empty. The best way would simply be to override `openhpc_partitions` using ansible extra vars in the ansible command line in packer, when `control` is in the group.

sjpb:
Enables building environment-specific control images in the same way as for login/compute builds.

Ticket: https://stackhpc.atlassian.net/browse/DEV-695

Note this does NOT move state off-board from the controller, so reimaging the controller will loose all state. However it does at least provide an image which can be used to create a working controller for an existing cluster (regardless of e.g. upstream package changes).

Note that `ansible/site.yml` will need running after imaging a node with this image, to set:
- Partition information in slurm.conf - even if compute nodes are up during image build they are not in the packer play so this information cannot be templated
- Potentially Prometheus scrape config, depending on which compute nodes were up during image build

Fixes #133. Replaces #136.

Requires:
 - [x] #159
 - [x] https://github.com/stackhpc/ansible-role-openhpc/pull/132
 - [x] Update requirements.yml
 - [x] #177 

Note that with `openondemand` enabled, building a control image BEFORE deploying the login node fails in `smslabs` environment as grafana needs to know the `openondemand_servername` which requires the private IP (defined as `.ansible_host`) for the login node. This doesn't occur in CI as a direct deployment of control/login/2x compute is done first which generates the `hosts` inventory file with this information. 

This limitation could be fixed in a later PR / for other environments by changing the TF to be two stages:
1. Deploy ports to get private IPs and template `hosts` file using this info (instead of from instances).
2. Deploy instances
and only doing step 1 before running the image build. That should also allow login & compute image build without the control node actually existing.

CI does not currently try the built image.

**NB:** CI is failing until requirements.yml updated, but waiting for another PR to merge on openhpc role before bumping version

Dev deployment: vglabs-steveb-ansible-rocky85:/home/rocky/slurm-app-control-images


sjpb:
Think CI ran out of instances, can retry later.

ansible/ci/delete_images.yml
```diff
@@ -2,22 +2,11 @@
...
     - name: Delete images
       shell:
         cmd: |
-          openstack image delete {{ login_build.artifact_id }}
-          openstack image delete {{ compute_build.artifact_id }}
+          openstack image delete {{ item.artifact_id }}
       delegate_to: localhost
-    
\ No newline at end of file
+      loop: "{{ manifest['builds'] }}"
```
m-bull, line 12: Do you need to run the `get_image_ids.yml` tasks if you're not using the `*_build` facts here?

ansible/ci/test_reimage.yml
```diff
@@ -63,3 +69,39 @@
...
+      retries: 5
+      delay: 30
+      until: control_build.artifact_id in openstack_control.stdout
+      changed_when: false
+    
+    - name: Wait for control connection
+      wait_for_connection:
+        timeout: 800
+    
+- name: Run slurm playbook again to add partition info
```
m-bull, line 97: Will this bit be needed when state is offboarded? Wondering out of my own curiosity really!

sjpb: Currently, yes. The [list of things offboarded](https://github.com/stackhpc/ansible-slurm-appliance/pull/173#issue-1210971616) doesn't include /etc/slurm.conf. Maybe it should do though, bit of a blurry line between "state" and "config" ...

ansible/ci/delete_images.yml
sjpb: I think so as I need the `manifest` fact it sets. I guess that fact would actually still be available (?) from the previous run of this playbook but that seems a bit fragile.

m-bull: Of course - didn't clock that properly...

m-bull (approved):

PR #163, opened 2022-04-04
Add Squid DNF proxy to smslabs CI

sjpb:


sjpb:
Timing info **with #166 and using mitogen** shows that this doesn't make much difference:
- with squid: real 8m24.258s
- without squid: real 8m42.779s

Issue #129, opened 2021-10-15
Default monitoring more suitable for large clusters


sjpb:
See [alaska 96d5ab1] for a config more suitable for smaller clusters.

sjpb:
Closed by https://github.com/stackhpc/ansible-slurm-appliance/pull/164

PR #164, opened 2022-04-05
By default scrape all the node exporter bits

JohnGarbutt:
Currently we are using the reduced set of scrapes we had when polling node exporter every second, this makes the storage bits of the dashboard empty. Instead we really should scrape everything by default, I think.


sjpb:
As used on alaska: https://github.com/stackhpc/ansible-slurm-appliance/blob/alaska/environments/common/inventory/group_vars/all/prometheus.yml

Fixes https://github.com/stackhpc/ansible-slurm-appliance/issues/129

sjpb (approved):

PR #165, opened 2022-04-05
Add exporter & dashboards for slurm itself

sjpb:
Add a [prometheus slurm exporter](https://github.com/vpenso/prometheus-slurm-exporter) and corresponding Grafana dashboard. This provides visibility of slurm state e.g. queue length, node state, job state etc. C.f. existing dashboards which provide visibility of node state for jobs.


ansible/roles/slurm_exporter/tasks/install.yml
```diff
@@ -0,0 +1,72 @@
...
+    append: true
+    shell: /usr/sbin/nologin
+    system: true
+    create_home: false
+    home: /
+
+- name: Install GCC
+  ansible.builtin.dnf:
+    name: gcc
+    state: latest
```
m-bull, line 20: Might `latest` upgrade gcc unpredictably? Probably doesn't matter on a control host, but people might be annoyed if it was a login node

ansible/roles/slurm_exporter/tasks/install.yml
```diff
@@ -0,0 +1,72 @@
...
+
+- name: Install go
+  ansible.builtin.unarchive:
+    remote_src: yes
+    src: "https://go.dev/dl/go{{ slurm_exporter_go_version }}.linux-{{ slurm_exporter_go_arch }}.tar.gz"
+    dest: /usr/local
```
m-bull, line 26: Could this go to `/tmp/{{ slurm_exporter_godir }}`, then be removed after being added to the PATH for the `make` stage?
Seems like quite a big thing to leave hanging around in an image just to build the exporter!

environments/common/inventory/group_vars/all/prometheus.yml
```diff
@@ -50,4 +50,11 @@ prometheus_scrape_configs_default:
   scrape_interval: 30s
   scrape_timeout: 20s
 
+- job_name: "slurm_exporter"
+  scrape_interval: 30s
+  scrape_timeout: 30s
+  static_configs:
+    - targets:
+      - "{{ openhpc_slurm_control_host }}:{{ slurm_exporter_port }}"
```
m-bull, line 58: If `--listen-address` is configurable, then `{{ openhpc_slurm_control_host }}` would need to be updated too.

ansible/roles/slurm_exporter/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+slurm_exporter_go_version: 1.15.15 # as per https://github.com/vpenso/prometheus-slurm-exporter, 1.16 doesn't work due to module changes
+slurm_exporter_go_arch: amd64
+slurm_exporter_user: slurm-exp
+slurm_exporter_group: "{{ slurm_exporter_user }}"
+slurm_exporter_builddir: /tmp/prometheus-slurm-exporter
+slurm_exporter_binary_install_dir: /usr/local/bin
+slurm_exporter_state: started
+slurm_exporter_port: 8080 # as per https://github.com/vpenso/prometheus-slurm-exporter/blob/master/DEVELOPMENT.md
```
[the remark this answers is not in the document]
sjpb: Yes, it is (and they [refuse](https://github.com/vpenso/prometheus-slurm-exporter/issues/44) to fix it). I'd thought it was better for the role default to follow the actual binary default (to document it if nothing else), then override in the appliance (where I need to define it in any case as the prometheus group also needs to know it). Comments?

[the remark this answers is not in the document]
m-bull: Justified!

ansible/roles/slurm_exporter/tasks/install.yml
sjpb: Agreed, good catch

ansible/roles/slurm_exporter/tasks/install.yml
m-bull: Hopefully you should be able to use https://github.com/stackhpc/prometheus-slurm-exporter/releases/download/0.20/prometheus-slurm-exporter-0.20-2.el8.x86_64.rpm to install the built exporter now, along with a user and listening on 9341.

environments/common/inventory/group_vars/all/prometheus.yml
sjpb: listen address is not configurable now we're using the unit file from the rpm

sjpb:
@m-bull ready for re-review please. The last commit has failed on smslabs due to hosts, but previous commit passed and is functionally no different.

m-bull (approved):

PR #166, opened 2022-04-07
Support using images with binaries pre-installed and use for smslabs CI

sjpb:
[stackhpc.slurm_image_builder](https://github.com/stackhpc/slurm_image_builder) can build an image which includes nearly all binaries required for all nodes in a default slurm appliance.

This PR:
- Updates the smslabs environment to use an image from [`slurm_image_builder`](https://github.com/stackhpc/slurm_image_builder) for both direct-deploy and as the base for image build, to try to reduce CI cycle time.
- Uses a `cloudalchemy.node_exporter` fork which copes with binaries pre-existing (see https://github.com/cloudalchemy/ansible-node-exporter/pull/265). This avoids the normal "localhost download + propagate to nodes" approach taken by this role.
- Disables prometheus binary installs when it finds binaries already exist and no specific version is required.

Current deployment: `vglabs-steveb-ansible-rocky85:/home/rocky/slurm-app-prebuilt`

### DIFF
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -76,6 +76,18 @@
   hosts: prometheus
   tags: prometheus
   tasks:
+    - name: Check for existing prometheus binaries
+      stat:
+        path: /usr/local/bin/{{ item }}
+      register: prometheus_binaries
+      loop:
+        - prometheus
+        - promtool
+    - name: Skip prometheus install if prometheus binaries exist and prometheus_version not defined
+      # i.e. if prometheus_version isn't defined we don't care, so use what's already there
+      set_fact:
+        prometheus_skip_install: "{{ false if prometheus_version is defined else true }}" 
+      when: "{{ prometheus_binaries.results | map(attribute='stat') | map(attribute='exists') }}"
     - import_role:
         name: cloudalchemy.prometheus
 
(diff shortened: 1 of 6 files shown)


ansible/monitoring.yml
```diff
@@ -76,6 +76,18 @@
   hosts: prometheus
   tags: prometheus
   tasks:
+    - name: Check for existing prometheus binaries
+      stat:
+        path: /usr/local/bin/{{ item }}
```
m-bull, line 81: Might binaries ever be installed somewhere that wasn't `/usr/local/bin`?

ansible/monitoring.yml
```diff
@@ -76,6 +76,18 @@
...
+      register: prometheus_binaries
+      loop:
+        - prometheus
+        - promtool
+    - name: Skip prometheus install if prometheus binaries exist and prometheus_version not defined
+      # i.e. if prometheus_version isn't defined we don't care, so use what's already there
+      set_fact:
+        prometheus_skip_install: "{{ false if prometheus_version is defined else true }}"
```
m-bull, line 89: Could add `when: prometheus_skip_install is not defined` to allow to it to be punched through directly from `group_vars` if its desirable for this to be able to be set directly?

ansible/monitoring.yml
sjpb: The install path is a "private" role variable (prefixed _) in the `cloudalchemy.prometheus` role so I think we can assume not, when using that role at least.

m-bull: Sounds sensible to me

m-bull (approved):

PR #173, opened 2022-04-21
Support moving state to persistent storage

sjpb:
- Allows defining a hostvar `appliances_state_dir` to put control node state onto persistent storage.
- Changes the default TF + group_vars to put /home on persistent storage (for all except control node)

Ticket: https://stackhpc.atlassian.net/browse/DEV-834

In the skeleton TF (as used for `arcus`):
- volumes `home` and `state` are created and attached to the control node
- cloud-init userdata is used to partition/format/mount thes to `/var/lib/state` and `/exports/home` on the control node
- the inventory template defines `appliances_state_dir` for the control *group* as `/var/lib/state`. **NB** this must be defined on the group, not the host, so that the Packer builds for `control` images also get this set.

Currently the persistent state on the slurm control node covers:
- slurmctld state
- MySQL database for slurmdbd
- Prometheus database
- Opendistro data
- Grafana data

It also adds documentation for this feature, and modifies the `block_devices` docs to explain why this isn't the right way to use volumes.

# CI (`arcus` environment)
This now:
- defines volumes on the control node, as described above
- runs hpctests before reimaging the cluster
- runs the `slurm.yml` playbook after the reimage to regenerate partition information (this is excluded from the control image build info)
- checks that the job info for the `hpctests` runs is still present after the reimage:
  a) from`sacct`, which checks mysql state has persisted and slurmdbd restarted
  b) from opendistro proxied through grafana, which checks the opendistro state has been persisted and the datasource works

# Manual checks
I have (manually) checked that after a reimage as above:
- `hpctests` works
- FAILED - OOD shell works - TODO: RECHECK now /home is persistent
- Monitoring works

# Caveats
- `ansible/slurm.yml` must be rerun after reimaging the control node to redefine partition information
- The default nfs config now assumes `/exports/home` exists, which assumes the default TF (or similar) is used.
- The new `mysql` role can't change the mysql root password after initialisation.

# Requires/TODOs:
- [x] Update requirements.yml after merging above
- [x] #160 to build environment-specific state
- [x] Move smslabs state to volume
- [x] Check AlaSKA disk usage and suggest volume size in comment
- [x] Add tests in CI on upgrade?
- [x] Fix https://github.com/stackhpc/ansible-slurm-appliance/issues/183
- [x] resolve mysql problem (below)
- [x] resolve slurm-stats problem (below)
- [x]  update for opendistro data
- [x] un-hack-ify fixes for unit dependencies to avoid starting before statedir mounted
- [ ] stop deleting compute images??


sjpb:
TODO: check AlaSKA disk usage and suggest volume size (ideally in comment)

sjpb:
Using this on a deployment, the recreation of a cluster fails with

```
TASK [geerlingguy.mysql : Disallow root login remotely] *****************************************************************************
failed: [dev-control] (item=DELETE FROM mysql.user WHERE User='root' AND Host NOT IN ('localhost', '127.0.0.1', '::1')) => {
    "ansible_loop_var": "item",
    "changed": false,
    "cmd": [
        "mysql",
        "-NBe",
        "DELETE FROM mysql.user WHERE User='root' AND Host NOT IN ('localhost', '127.0.0.1', '::1')"
    ],
    "delta": "0:00:00.004478",
    "end": "2022-07-14 11:24:54.025773",
    "item": "DELETE FROM mysql.user WHERE User='root' AND Host NOT IN ('localhost', '127.0.0.1', '::1')",
    "rc": 1,
    "start": "2022-07-14 11:24:54.021295"
}

STDERR:

ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)
```

sjpb:
Despite the CI there also *appears* to be a problem with 9893c35, where the firewalld install task fails with a directory error on `/home/rocky`, which should be `/var/lib/rocky`.

sjpb:
the [slurm-stats role generates](https://github.com/stackhpc/ansible_collection_slurm_openstack_tools/blob/main/roles/slurm-stats/tasks/main.yml#L13) `/var/log/slurm-stats/finished_jobs.json`  - this needs to be moved onto state too.

sjpb:
Agreed with @m-bull that moving the podman volume root is the wrong approach here. Will mount in host directories (owned by `podman` user) to containers instead (using `:U` switch). Will provide more flexibility to deal with upgrades/changes/etc.

use `:U` switch on volume mounts

ansible/roles/block_devices/README.md
```diff
@@ -1,17 +1,19 @@
...
+Manage filesystems on block devices (such as OpenStack volumes), including creating partitions, creating filesystems and mounting filesystems.
 
 This is a convenience wrapper around the ansible modules:
 - community.general.parted
 - community.general.filesystem
 - ansible.buildin.file
 - ansible.posix.mount
 
-It includes logic to handle OpenStack-provided volumes appropriately both for appliance instances and the Packer build VM.
+To avoid issues with device names changing after e.g. reboots, devices are identified by serial number and mounted by filesystem UUID.
```
m-bull, line 12: Is this true still? `control.userdata.tpl` doesn't have any UUIDs in it.

environments/skeleton/{{cookiecutter.environment}}/terraform/control.userdata.tpl
```diff
@@ -0,0 +1,21 @@
+#cloud-config
+disk_setup:
```
m-bull, line 2: If there's just a single partition per-device, should we just create the filesystem directly on the device rather than partitioning first?

m-bull (changes requested):

ansible/ci/check_sacct_hpctests.yml
```diff
@@ -0,0 +1,27 @@
+- hosts: control
+  gather_facts: false
+  become: true
+  vars:
+    sacct_stdout_expected: |- # based on CI running hpctests as the first job - NB note no trailing newline
+      JobID,JobName,State
```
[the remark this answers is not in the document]
sjpb: The first JobID on a new openhpc cluster always seems to be 1, I've noticed that before (unless you set a higher starting nunber). Not sure whether its a slurm or openhpc thing.

ansible/roles/mysql/README.md
```diff
@@ -0,0 +1,54 @@
...
+Requirements
+------------
+
+None.
+
+Role Variables
+--------------
+
+- `mysql_root_password`: Required str. Password to set for `root` mysql user. **NB** This cannot be changed by this role once mysql server has initialised.
+- `mysql_tag`: Optional str. Tag for version of `mysql` container image to use. Default `8.0.30`.
```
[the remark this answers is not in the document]
sjpb: This does actually use mysql: https://hub.docker.com/_/mysql. The previous non-containerisered version used mariadb, despite calling everything mysql.

ansible/roles/mysql/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+    dest: /etc/systemd/system/mysql.service
+    src: mysql.service.j2
+  register: _mysql_unitfile
+
+- name: Create environment file for mysql server root password
+  # NB: This doesn't trigger a restart on changes as it will be ignored once mysql is initialised
+  copy:
+    dest: /etc/sysconfig/mysqld
+    content: |
+      MYSQL_INITIAL_ROOT_PASSWORD='{{ mysql_root_password }}'
```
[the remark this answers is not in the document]
sjpb: No. As per the [docs](https://hub.docker.com/_/mysql):
> `MYSQL_ROOT_PASSWORD`
This variable is mandatory and specifies the password that will be set for the MySQL root superuser account

environments/skeleton/{{cookiecutter.environment}}/terraform/control.userdata.tpl
```diff
@@ -0,0 +1,21 @@
...
+  /dev/vdb:
+    table_type: gpt
+    layout: true
+  /dev/vdc:
```
m-bull, line 6: How can we control for device names here? `/dev/vda` vs `/dev/sda` etc - these are properties of the SCSI device I think...

sjpb: You can't really. It is supposed to depend on the flags set on the image, so setting `--property hw_scsi_model=virtio-scsi` should I think give you `sd*` although in my test it didn't. For the use of this TF I think it's ok - in CI we control both the device name here and the image tags. If you're using this TF in your own deployment you may need to change it but that goes for anything

ansible/roles/block_devices/README.md
sjpb: I haven't seen swaps after reboot in the same way that I did with the original block_devices role (the above isn't actually a change, just a clarification in the docs). I think part of the problem originally was doing explicit volume attachments in TF, i.e. post-boot. They were definitely liable to change ordering. The TF used here creates the volumes by specifying multiple ephemeral disks for the server, which is more tedious wrt syntax but seems to ensure the device ordering follows the definition ordering. The other part is mounts; looking a node created this way it appears the filesystem gets a label and that is used for the mount, which is safe:

```
[rocky@offboardstate-control ~]$ lsblk -o NAME,TYPE,LABEL,FSTYPE,MOUNTPOINT
NAME   TYPE LABEL FSTYPE MOUNTPOINT
vda    disk              
└─vda1 part       xfs    /
vdb    disk state ext4   /var/lib/state
vdc    disk home  ext4   /exports/home
[rocky@offboardstate-control ~]$ cat /etc/fstab 

#
# /etc/fstab
# Created by anaconda on Sat Jul  2 12:41:01 2022
#
# Accessible filesystems, by reference, are maintained under '/dev/disk/'.
# See man pages fstab(5), findfs(8), mount(8) and/or blkid(8) for more info.
#
# After editing this file, run 'systemctl daemon-reload' to update systemd
# units generated from this file.
#
UUID=1cff4c58-02ed-41c1-b5fc-b29f12aae17a /                       xfs     defaults        0 0
LABEL=state     /var/lib/state  auto    defaults,nofail,x-systemd.requires=cloud-init.service,comment=cloudconfig     02
LABEL=home      /exports/home   auto    defaults,nofail,x-systemd.requires=cloud-init.service,comment=cloudconfig     02
offboardstate-control:/exports/home /home nfs defaults 0 0
```

sjpb:
TODO: do I need to delay nfs server start till after /home mounted from /etc/exports?

m-bull:
> TODO: do I need to delay nfs server start till after /home mounted from /etc/exports?

This might be useful: https://www.freedesktop.org/software/systemd/man/systemd.mount.html#x-systemd.wanted-by=

ansible/roles/block_devices/README.md
m-bull: This was more a comment on the docs should probably `s/UUID/LABEL` here.

m-bull (approved):
I think this looks okay once that UUID/LABEL docs change is complete

ansible/roles/block_devices/README.md
sjpb: In `block_devices` they are mounted by UUUID: https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/roles/block_devices/tasks/main.yml#L47

Issue #72, opened 2021-05-11
Can't override environment-defined config in packer build


sjpb:
Because `builder` is near the start of the alphabet it is impossible to override group_vars defined in a concrete environment (rather than `common/.../all`).

e.g.:
`environments/myenv/inventory/group_vars/compute/defaults.yml` is not overriden by `myenv/.../builder/overrides.yml`

Not sure how to fix this. Changing `builder` group to be near the end of the alphabet, e.g. `xtrabuilder` does not do it,

jovial:
Potential workaround for such issues:

`environments/myenv/inventory/group_vars/compute/defaults.yml`:
```
some_value: "{{ some_value_override | default('default-value') }}"
```

`myenv/.../builder/overrides.yml`:
```
some_value_override: my-override
```

But yeah, a little bit hairy...

PR #175, opened 2022-04-27
Fix openondemand_servername default

sjpb:
Currently `openondemand_servername` is set to `''` in `common/all/group_vars` to permit templating on grafana host even when openondemand is not configured.

However setting it here means it can't[^1] be overriden by the `[all:vars]` section in an environment's (TF-templated) `inventory/hosts` file, which is the natural place to define it given it a) depends on infrastructure, at least when it is a FIP rather than a fully-qualified hostname and b) needs to be defined on at least the ood and grafana hosts.  

This PR removes it from group_vars/all and fixes templating to cope with the fact it may be undefined.

[^1]: The relevant [precedence rules](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable) are items 3. and 4.

### DIFF
--- a/environments/common/inventory/group_vars/all/grafana.yml
+++ b/environments/common/inventory/group_vars/all/grafana.yml
@@ -14,7 +14,7 @@ grafana_api_url: "http://{{ grafana_api_address }}:{{ grafana_port }}"
 # Configure external address, with external URL depending on whether we are using Open Ondemand as a proxy
 grafana_address: "{{ hostvars[groups['grafana'].0].api_address }}"
 grafana_url_direct: "http://{{ grafana_address }}:{{ grafana_port }}"
-grafana_url_openondemand_proxy: "https://{{ openondemand_servername }}/node/{{ groups['grafana'].0 }}/{{ grafana_port }}"
+grafana_url_openondemand_proxy: "https://{{ openondemand_servername | default('') }}/node/{{ groups['grafana'].0 }}/{{ grafana_port }}"
 grafana_url: "{{ grafana_url_openondemand_proxy if groups['openondemand'] | count > 0 else grafana_url_direct }}"
 grafana_serve_from_sub_path: "{{ groups['openondemand'] | count > 0 }}"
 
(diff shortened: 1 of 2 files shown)


jovial (commented):
It does make me wonder if we should migrate our defaults to be: `inventory file or script group vars` i.e move groups_vars/all/grafana.yml to grafana with:

```
[all:vars]
grafana_address = ....
```
Does the more specific inventory always win in this case? Or is there some sort of alphabetical ordering (I seem to remember a quirk like that)? In which case, you might want some prefix that always sorts to the beginning.

sjpb:
If you're suggesting moving `common/inventory/group_vars/all/grafana` to `common/inventory/group_vars/grafana` then the more-specific one always wins.

What seems to happen is that the merging happens in precedence ordering _first_, with inventory ordering only important when precedence in multiple inventories is the same. Not sure how lexicographical ordering plays in here, there was this #72 which had some notes.

There is also `ansible_group_priority` (described under [here](https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#how-variables-are-merged)) which we've never used yet.

jovial:
> If you're suggesting moving `common/inventory/group_vars/all/grafana` to `common/inventory/group_vars/grafana` then the more-specific one always wins.
> 
> What seems to happen is that the merging happens in precedence ordering _first_, with inventory ordering only important when precedence in multiple inventories is the same. Not sure how lexicographical ordering plays in here, there was this #72 which had some notes (ETA: so that wouldn't work).
> 
> There is also `ansible_group_priority` (described under [here](https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#how-variables-are-merged)) which we've never used yet.

I was thinking we could make a hosts file with the the variables defined:

inventory/<hosts file>

with:

```
[all:vars]
```

i.e so that we are using the lowest precedence level we can. I think this might break the caas appliance though...

jovial (approved):

sjpb:
Ah I see. That's not a bad idea actually and I think should work. CaaS looks like [this](https://github.com/stackhpc/caas-slurm-appliance/tree/main/group_vars), so you're right it would break it but we've been discussing changing this anyway.

PR #177, opened 2022-04-28
Packer build fixes

sjpb:
Couple of fixes for packer builds from experience in non-smslabs environment.

**NB:** This disables block device configuration. This is probably correct, but needs discussing.


sjpb:
Need to add some documentation to:
- Note that `block_devices` can't work during image build
- Suggest using cloud-init instead.
- Point to TF example in https://github.com/stackhpc/ansible-slurm-appliance/pull/173/commits/aaddf02487d8be6d68444bd358fbf8b2707c53a6

PR #184, opened 2022-05-11
Disable updates and test OFED build in arcus CI

sjpb:
This uses a `slurm_image_builder` produced image containing OFED in arcus CI.

As the OFED install replaces some system packages, updates are no longer done in the packer build, and the update play has been removed from site.yml. Updates could still be done manually on a running cluster using the ad-hoc playbook.

TODO:
- [ ] Merge OFED build into pipeline
- [ ] Check rdma-core in image


sjpb:
Replaced by #254 using in-appliance packer build

PR #197, opened 2022-07-21
Replace opendistro

### LINKED ISSUE #70: Versions of docker images not easily configurable
For example: https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/roles/kibana/templates/kibana.service.j2#L12

But I think the same is true for other podman services. Would be nice to factor this out into a role variable.

sjpb:
Ticket: https://stackhpc.atlassian.net/browse/DEV-855

OpenDistro is EOL.

This PR:
- [x] Replaces OpenDistro with OpenSearch.
- [x] Updates `filebeat` to the [newest-supported version](https://opensearch.org/docs/2.1/clients/agents-and-ingestion-tools/index/#compatibility-matrices).
- [x] Adds the [required version faking](https://opensearch.org/docs/2.1/clients/agents-and-ingestion-tools/index/) to enable filebeat.
- [x] Configures [important opensearch settings](https://opensearch.org/docs/latest/opensearch/install/important-settings) for production use.
- [x] Updates Grafana version
- [x] Updates the opensearch Grafana datasource plugin definition
- [x] Removes the appliances `grafana-datasources` role as we can use grafana's provisioning mode with the `cloudalchemy.grafana` rather than requiring a customised API-based approach. <-- TODO CHECK: think this was merged already.
- [x] Adds a test in CI that expected jobs from the hpctests runs are found via Grafana (NB: for slurm-stats this has to be 5 mins past job completion, so may add some delay)
- [x] Changes the storage used for open{distro,search} from a podman volume to a host directory to enable easier future upgrades/migration/backups.
- [x] Adds a playbook `ansible/adhoc/migrate-opendistro.yml` to migrate opendistro data to opensearch (checked by upgrading a running cluster from `main` `7bcacb0`)
- [x] Uses a new "prebuilt" image in arcus with the updated Grafana version (note actually CI was using this before this PR, so grafana's been getting downgraded during CI deployments)
- [x] Merge workaround for `ohpc-base-compute` dependency on singularity.
- [x] Update image build with correct grafana version, in a PR, and test that.
- [x] TODO: fix/document migration - currently it will always run if opendistro service even *exists*.
- [ ] 
Once merged and passed:

- [ ] Move appropriate image to release bucket

Closes #70.


sjpb:
It is leaving massive gaps between jobs:
image

sjpb:
3a25308 fails on checking that the hpctests jobs exist in grafana/opensearch. Turns out neither dashboard nor datasources have been provisioned in grafana after rebuilding control node with packer-built image (although grafana is running). Possibly this has always been broken, just not checked for until this PR.

direct configuration:
```
2022-08-11T12:48:19.5059996Z TASK [cloudalchemy.grafana : Create/Update datasources file (provisioning)] ****
2022-08-11T12:48:19.5062754Z task path: /home/runner/work/ansible-slurm-appliance/ansible-slurm-appliance/ansible/roles/cloudalchemy.grafana/tasks/datasources.yml:26
2022-08-11T12:48:23.6044928Z NOTIFIED HANDLER cloudalchemy.grafana : restart grafana for ci2839370468-control
2022-08-11T12:48:23.6047794Z changed: [ci2839370468-control] => {
2022-08-11T12:48:23.6048876Z     "changed": true,
2022-08-11T12:48:23.6050009Z     "checksum": "c3246446b05c316d4a9dfde60badfa068fd3936c",
2022-08-11T12:48:23.6051407Z     "dest": "/etc/grafana/provisioning/datasources/ansible.yml",
2022-08-11T12:48:23.6052598Z     "gid": 979,
2022-08-11T12:48:23.6053560Z     "group": "grafana",
2022-08-11T12:48:23.6054671Z     "md5sum": "ed1ebc1e5e73ffca64f835dea1145202",
2022-08-11T12:48:23.6055784Z     "mode": "0640",
2022-08-11T12:48:23.6056768Z     "owner": "root",
2022-08-11T12:48:23.6057824Z     "secontext": "system_u:object_r:etc_t:s0",
2022-08-11T12:48:23.6058901Z     "size": 569,
2022-08-11T12:48:23.6060242Z     "src": "/var/lib/rocky/.ansible/tmp/ansible-tmp-1660222100.111777-5114-195439064785952/source",
2022-08-11T12:48:23.6061512Z     "state": "file",
2022-08-11T12:48:23.6062449Z     "uid": 0
2022-08-11T12:48:23.6063363Z }
```

control image build:
```
2022-08-11T12:57:56.6329342Z     openstack.control: TASK [cloudalchemy.grafana : Create/Update datasources file (provisioning)] ****
2022-08-11T12:57:56.6339799Z     openstack.control: task path: /home/runner/work/ansible-slurm-appliance/ansible-slurm-appliance/ansible/roles/cloudalchemy.grafana/tasks/datasources.yml:26
2022-08-11T12:57:56.6530631Z     openstack.control: skipping: [default] => {
2022-08-11T12:57:56.6532331Z     openstack.control:     "changed": false,
2022-08-11T12:57:56.6533879Z     openstack.control:     "skip_reason": "Conditional result was False"
2022-08-11T12:57:56.6535269Z     openstack.control: }
2022-08-11T12:57:56.7564322Z     openstack.control:```

**Comment by sjpb** [comment]
Various problems:
- [ ]: opensearch container failed, looks like this: https://access.redhat.com/solutions/6904591
- [ ] grafana dashboards aren't provisioned in the image, check logic but presumably b/c this is being done via the API and grafana is not up in the builder (check what the logic is/what actually happens)
- [ ] grafana admin password is not set after rebuild - is at default `admin` value despite correctly being in `/etc/grafana/grafana.ini` - this is the error which is currently failing CI (although then the others will cause failure too)

**Comment by sjpb** [comment]
The image `openhpc-220811-0842.qcow2` (here)[https://github.com/stackhpc/slurm_image_builder/pull/5/files] installs 9.0.3, then the upgrade task bumps to 9.0.7, then the pinned to 9.0.3 at `environments/common/inventory/group_vars/all/grafana.yml:grafana_version`, means the grafana role downgrades it again.

**Comment by sjpb** [comment]
A note on certs:

- The version faking for filebeat requires a custom `opensearch.yml` config file to be mounted into the container.
- As we want to add an admin user password we have to enable the security plugin, which means certs have to be configured.
- The opensearch container will by default use hardcoded certs and configure opensearch.yml - in our case that means its modified the one mounted in from the host.
- This is problematic as the certs it uses are copied into the config directory (next to the `opensearch.yml`) - and are therefore lost when the container is deleted, which happens as we use `podman run --remove` but would anyway happen on reimage (how I found the problem). When opensearch starts again it sees the certs listed in opensearch.yml and dies as they do not exist.
- We cannot mount in the entire config directory as the one in the container has files which are required. We cannot mount it as an overlay as we need the `:U` flag on the volume which is not compatible with other options. We cannot mount the default certs to somewhere writeable on the host, as they don't exist on first startup.

The solution is to generate self-signed certs ourselves, mount them into the container, and add them to opensearch.yml ourselves. This appears to be fine; a minor problem is that runnning `monitoring.yml` again changes the owner/group of some files which are mounted in, which kills the service. It restarts almost instantaneously though and recovers.

Finally note that this uses the `plugins.security.allow_default_init_securityindex: true` to avoid having to wait for opensearch container to finish initialising and then run `securityadmin.sh`. I note that the docs state:
> The opensearch.yml file also contains the plugins.security.allow_default_init_securityindex property. When set to true, the security plugin uses default security settings if an attempt to create the security index fails when OpenSearch launches. Default security settings are stored in YAML files contained in the opensearch-project/security/config directory. By default, this setting is false.

I think the only potentially-problematic part of this is [opensearch-project/security/config/internal_users.yml](https://github.com/opensearch-project/security/blob/main/config/internal_users.yml), but we create our own internal users anyway which will override this.

**Comment by sjpb** [comment]
Note certs have a hardcoded 2yr life.

**Comment by sjpb** [comment]
@m-bull I tried using `community.crypto.x509_certificate_info` to extract validity and delete if necessary, but as podman chowns everything in `certs/` the ansible loops/logic were just getting really messy. Put that on hold as not best use of effort at the moment.

File: ansible/adhoc/migrate-opendistro.yml [hunk 1]
```diff
@@ -0,0 +1,18 @@
...
+# This does not automatically remove the opendistro podman volume. That can be done manually once happy data has transferred ok.
+
+- hosts: opensearch
+  become: no
+  gather_facts: no
+  tasks:
+    - name: Export opendistro volume
+      import_role:
+        name: opensearch
+        tasks_from: migrate-opendistro.yml
```
m-bull, line 18: Can this be run as part of the role/tasks/main.yml rather than an adhoc playbook? Existing Azimuth-deployed clusters would need this ad-hoc playbook to run as part of an upgrade.

ansible/roles/opensearch/tasks/certs.yml
```diff
@@ -0,0 +1,44 @@
...
+  command:
+    cmd: "openssl req -new -x509 -sha256 -key root-ca-key.pem -out root-ca.pem -days {{ opensearch_certs_duration }} -subj '/C=XX'"
+    chdir: "{{ opensearch_config_path }}/certs"
+    creates: root-ca.pem
+  become_user: "{{ opensearch_podman_user }}"
+
+- name: Generate node cert
+  # FIXME: need a way to regenerate a new one on expiry?
+  shell:
+    cmd: |
```
m-bull, line 27: Is there any way this can be done with ansible modules rather than a multiline shell command?

ansible/roles/opensearch/templates/opensearch.service.j2
```diff
@@ -0,0 +1,39 @@
...
+# see also https://opensearch.org/docs/2.0/opensearch/install/important-settings/
+ExecStart=/usr/bin/podman run \
+    --network=host \
+    --sdnotify=conmon --cgroups=no-conmon -d --replace --name opensearch --restart=no --user opensearch \
+    --ulimit memlock=-1:-1 --ulimit nofile=65536:65536 \
+    --volume {{ opensearch_data_path }}:/usr/share/opensearch/data:U \
+    --volume {{ opensearch_config_path }}/internal_users.yml:/usr/share/opensearch/config/opensearch-security/internal_users.yml:U \
+    --volume {{ opensearch_config_path }}/opensearch.yml:/usr/share/opensearch/config/opensearch.yml:U \
+    --volume {{ opensearch_config_path}}/certs:/usr/share/opensearch/config/certs:U \
+    --env node.name=opensearch --env discovery.type=single-node --env bootstrap.memory_lock=true --env "OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m" \
```
m-bull, line 22: nit: Split each `--env` onto a newline

sjpb:
FIXME: that merge won't be right as we need an image using updated grafana etc.

ansible/monitoring.yml
```diff
@@ -2,18 +2,12 @@
 # # NOTE: Requires slurmdbd
 
 - name: Setup elasticsearch
```
m-bull, line 4: nit: Setup OpenSearch

ansible/roles/filebeat/templates/filebeat.service.j2
```diff
@@ -12,7 +12,12 @@ After=network-online.target
 [Service]
 Environment=PODMAN_SYSTEMD_UNIT=%n
 Restart=always
-ExecStart=/usr/bin/podman run --network slirp4netns:cidr={{ podman_cidr }} --sdnotify=conmon --cgroups=no-conmon --replace --name filebeat --user root --restart=always --security-opt label=disable --volume /var/log/:/logs:ro --volume /etc/filebeat/filebeat.yml:/usr/share/filebeat/filebeat.yml:ro --detach=True docker.elastic.co/beats/filebeat-oss:7.9.3 -e -strict.perms=false -d "*"
+ExecStart=/usr/bin/podman run \
+    --network=host \
+    --sdnotify=conmon --cgroups=no-conmon --replace --name filebeat --user root --restart=always --security-opt label=disable \
+    --volume /var/log/:/logs:ro \
+    --volume /etc/filebeat/filebeat.yml:/usr/share/filebeat/filebeat.yml:ro \
+    --detach=True docker.elastic.co/beats/filebeat-oss:{{ filebeat_version }} -e -strict.perms=false -d "*"
```
m-bull, line 20: nit: split env var arg onto newline

ansible/roles/filebeat/templates/filebeat.service.j2
```diff
@@ -12,7 +12,12 @@ After=network-online.target
 [Service]
 Environment=PODMAN_SYSTEMD_UNIT=%n
 Restart=always
-ExecStart=/usr/bin/podman run --network slirp4netns:cidr={{ podman_cidr }} --sdnotify=conmon --cgroups=no-conmon --replace --name filebeat --user root --restart=always --security-opt label=disable --volume /var/log/:/logs:ro --volume /etc/filebeat/filebeat.yml:/usr/share/filebeat/filebeat.yml:ro --detach=True docker.elastic.co/beats/filebeat-oss:7.9.3 -e -strict.perms=false -d "*"
+ExecStart=/usr/bin/podman run \
+    --network=host \
+    --sdnotify=conmon --cgroups=no-conmon --replace --name filebeat --user root --restart=always --security-opt label=disable \
```
m-bull, line 17: nit: split these onto newlines

ansible/roles/mysql/templates/mysql.service.j2
```diff
@@ -15,7 +15,7 @@ EnvironmentFile=/etc/sysconfig/mysqld
 ExecStartPre=+install -d -o {{ mysql_podman_user }} -g {{ mysql_podman_user }} -Z container_file_t {{ mysql_datadir }}
 ExecStartPre=+chown -R {{ mysql_podman_user }}:{{ mysql_podman_user }} {{ mysql_datadir }}
 ExecStart=/usr/bin/podman run \
-    --network slirp4netns:cidr={{ podman_cidr }} \
+    --network=host \
     --sdnotify=conmon --cgroups=no-conmon \
```
m-bull, line 19: nit: split over lines

ansible/roles/mysql/templates/mysql.service.j2
```diff
@@ -15,7 +15,7 @@ EnvironmentFile=/etc/sysconfig/mysqld
...
     --detach --replace --name mysql --restart=no \
```
m-bull, line 20: nit: split over lines

ansible/roles/opensearch/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+# Used to set passwords
+#opensearch_internal_users_path:
+
+opensearch_podman_user: "{{ ansible_user }}"
+opensearch_version: '2.4.0' # https://hub.docker.com/r/opensearchproject/opensearch/tags
+opensearch_config_path: /usr/share/opensearch/config
+opensearch_data_path: /usr/share/opensearch/data
+opensearch_state: started # will be restarted if required
+opensearch_systemd_service_enabled: true
+opensearch_certs_duration: "{{ 365 * 10 }}" # days validity for self-signed certs
```
m-bull, line 11: This is clever and readable :)

ansible/roles/opensearch/tasks/certs.yml
```diff
@@ -0,0 +1,70 @@
...
+  file:
+    path: "{{ opensearch_config_path }}/certs"
+    state: directory
+    owner: "{{ opensearch_podman_user }}"
+    group: "{{ opensearch_podman_user }}"
+    mode: ug=rwx,o=
+
+# Cert generation based on https://opensearch.org/docs/latest/security-plugin/configuration/generate-certificates/
+
+- name: Generate root private key
```
m-bull, line 11: This whole process is loads nicer now, thanks for updating!

ansible/roles/opensearch/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
...
+- name: Ensure required opensearch host directories exist
+  file:
+    state: directory
+    path: "{{ item }}"
+    owner: "{{ opensearch_podman_user }}"
+    group: "{{ opensearch_podman_user }}"
+    mode: 0770
+  become: true
+  loop:
+    - "{{ opensearch_config_path }}/"
```
m-bull, line 31: Is the trailing `/`  needed?

ansible/roles/opensearch/templates/opensearch.service.j2
```diff
@@ -0,0 +1,42 @@
...
+After=network-online.target
+
+[Service]
+Environment=PODMAN_SYSTEMD_UNIT=%n
+Restart=always
+# paths below based on https://opensearch.org/docs/latest/opensearch/configuration/ and https://opensearch.org/docs/latest/security-plugin/configuration/yaml
+# see also https://opensearch.org/docs/2.0/opensearch/install/important-settings/
+ExecStart=/usr/bin/podman run \
+    --network=host \
+    --sdnotify=conmon --cgroups=no-conmon -d --replace --name opensearch --restart=no --user opensearch \
```
m-bull, line 16: nit: split over lines

ansible/roles/opensearch/templates/opensearch.yml.j2
```diff
@@ -0,0 +1,34 @@
+### This section copies a running container's default config ###
```
m-bull, line 1: Should this be a template if it doesn't have any template strings in?

docs/monitoring-and-logging.README.md
```diff
@@ -11,11 +11,10 @@ Parses log files and ships them to elasticsearch. Note we use the version shippe
 Visualisation tool that supports multiple different datasources. In our stack,
 we use it to visualise prometheus and elasticsearch data.
 
-### [opendistro](https://opendistro.github.io/)
+### [opensearch](https://opendistro.github.io/)
```
m-bull, line 14: Update URL

m-bull (changes requested):

ansible/roles/opensearch/tasks/main.yml
sjpb: Nope, hangover from using `ansible.builtin.copy` here which does need it

ansible/roles/opensearch/templates/opensearch.yml.j2
sjpb: Hmm not really, but I'm not sure if the `copy` module searches inside a role?

m-bull (dismissed):

m-bull (approved):

PR #200, opened 2022-07-27
Plot pingpong results

sjpb:
Retrieve pingpong results to localhost and produce a combined latency/bandwidth plot to allow for easy review of shape etc:

image

image


m-bull (approved):
Nice!

PR #203, opened 2022-08-04
Use volume-backed instances for smslabs

sjpb:
Currently CI on smslabs is flaky due to a lack of disk capacity for VMs. This PR adds a variable to the skeleton terraform `volume_backed_instances` which is set to `true` in `smslabs` to use volume-backed instances.

However this also requires the use of `raw`-format images (otherwise glance times out when launching >1 instance). So the image also needs to be converted from the arcus-built `qcow2` image.


sjpb:
SMSLabs no longer in CI

PR #205, opened 2022-08-16
Update grafana

sjpb:
This:
- Bumps grafana from 8.5.1 to 8.5.9 (latest on 8.5.x series, which is compatible with opendistro).
- Excludes grafana from yum update to fix the packer-built control image:
  - The current "fat" image used (`openhpc-220810-0839`) as the base image contains the grafana 8.x package.
  - The yum upgrade during packer build installs grafana 9.x. The package's script detects there's been an "upgrade" & starts grafana. As only the package is present with no config, grafana initialises with the default admin password. This means that once reimaging with the packer build image, the desired admin password cannot be set.
- Defines grafana datasources using provisioning mode rather than through the API, to allow these to be "baked in" to the packer-built control image
- Adds a check that the jobs run by hpctests are visible in the Grafana slurm-stats datasource.


ansible/roles/grafana-dashboards/tasks/main.yml
```diff
@@ -108,17 +108,51 @@
...
-        path: "{{ _tmp_dashboards.path }}/{{ ((item.dashboard_id | string) + '.json') if 'dashboard_id' in item else item.dashboard_file }}"
-        # message renamed to commit_message (Be aware if using old ansible)
-        # https://github.com/ansible/ansible/pull/60051
-        #commit_message: Updated by ansible
-        state: present
-        overwrite: true
-      #no_log: true
-      with_items: "{{ grafana_dashboards }}"
-      when: grafana_state | default('started') != 'stopped'
+- name: Create/Update dashboards file (provisioning)
```
sjpb, line 111: @m-bull this is /basically/ https://github.com/cloudalchemy/ansible-grafana/blob/master/tasks/dashboards.yml#L108, modified to run on the grafana node rather than on the deploy host.

m-bull (approved):

PR #206, opened 2022-08-18
Support generic compute & login images

sjpb:
Builds and tests compute and login images without environment-specific config or secrets. In brief this enables reuse of compute and login images built in CI here for any CaaS or cookiecutter created environment - but see further comments below.

The approach for this is that:
a) In Packer builds, the templated config/secrets are replaced by either a) dummy values or b) values which reference an environment variable.
b) Systemd units are modified via drop-ins to run after `cloud-init.final`, and if necessary to add an EnvironmentFile directive.
d) An ansible adhoc command templates out cloud_init userdata containing `write_file` options which create files or EnvironmentFiles using values from the active environment's group_vars.

On boot the services therefore get values read from the cloud-init generated files, i.e. environment-specific values rather than the dummy values in the image.

To make this moderately convenient, some machinery has been added:
- Extravars to override the actual environment values with the dummy/shell expansion values - see `packer/{compute,login}.yml`
- A role `systemd` which can create dropins for units
- A role `cloud_init` which can create userdata for hosts from service- and group-specific fragments, plus an adhoc to template out the above.
- An adhoc to template out the ba

# Support and Limitations:
- This can provide generalised compute and control images for appliances using the following groups/roles
  - `openhpc` - always required
  - `openondemand` - optional but enabled in `everything` layout (requires additional configuration)
  - `etc_hosts` - optional, used in arcus/CI
  - `rebuild` - optional, used in arcus/CI, desirable for CaaS
  - `basic_users` - optional, used in arcus/CI
  - `nfs` - optional but enabled in `everything` layout with default configuration in `common` environment - NB: `compute` and `login` hosts cannot be NFS servers.

# TODO:
- [ ] describe possible workflows
- [ ] upload images to s3


sjpb:
replaced by compute-init approach

PR #207, opened 2022-08-22
Fix packer build defaults

sjpb:
The packer build contains some variables used to proxy through a jumphost - this is unlikely to be needed for real site deployments but is required for CI. This PR adds defaults so that the packer build works when these aren't set.

This is a cherry pick from NREL


m-bull (approved):
LGTM

PR #209, opened 2022-08-22
Create role to manage /etc/hosts and use in arcus

sjpb:
- Currently the `arcus` environment for CI (and other client ones) use a pre-hook to template out `/etc/hosts` to workaround a lack of internal DNS.
- When rebuilding compute nodes with "generic" compute images (without environment-specific config/secrets), clearly a pre-hook won't work and `/etc/hosts` needs to be injected during boot.
- For multi-homed hosts, how this is generated from inventory needs to be configurable.

This PR adds an `etc_hosts` role to support the above. It is done via a fairly-generic `cloud_init` role to provide support for:
- persistent directories on volumes (added by #173)
- future cloud_init userdata planned for "generic" or "no-config" images.

**NB** This means that the playbook `ansible/adhoc/template-cloud-init.yml` MUST be run before running terraform. However as IPs are required to template inventory, a `create_servers` var has been added to the default TF, and inventory templating is now done using ports instead. This will actually help for the generic images work but does change the required deploment workflow - see `.github/workflows/stackhpc.yml` for details

**NB** As nodes are now ALWAYS launched with userdata (if they need to be), the slurm-driven rebuild should also work without having to specifically set the userdata during the rebuild (as openstack will reuse previous userdata for a rebuild unless told otherwise). This is handy as it avoids having to change the `rebuild` script.

TODO:
- [x] Add support for this in rebuild adhoc
- ~~[ ] Add support for this in slurm-driven rebuild~~ No longer required.


sjpb:
Closing as this is just a part of #206 really.

PR #213, opened 2022-09-20
Make multiple OOD nodes safe

### LINKED ISSUE #212: OOD proxying assumes only single login node
OOD defaults are potentially broken for >1 login node.

1. Everything layout defines openondemand as all login, so could have multiple openondemand nodes
```
environments/common/layouts/everything:
[openondemand:children]
# Host to run Open Ondemand server on - subset of login
login
```

1. Clearly grafana can only proxy a single OOD node, so this proxy config assumes `openondemand_servername` is the same on all hosts:
```
environments/common/inventory/group_vars/all/grafana.yml:grafana_url_openondemand_proxy: "https://{{ openondemand_servername | default('') }}/node/{{ groups['grafana'].0 }}/{{ grafana_port }}"
```

1. The `arcus` definition is actually ok for this:
```
environments/arcus/inventory/group_vars/all/openondemand.yml:openondemand_servername: "{{ hostvars[ groups['openondemand'] | first].ansible_host }}" # Use a SOCKS proxy to acccess
```

but this will be really easy to mess up

sjpb:
Fixes #212.

### DIFF
--- a/environments/common/inventory/group_vars/all/grafana.yml
+++ b/environments/common/inventory/group_vars/all/grafana.yml
@@ -10,15 +10,21 @@ grafana_port: 3000
 # Define where state is stored
 grafana_data_dir: "{{ appliances_state_dir | default('/var/lib') }}/grafana"
 
-# Configure internal address & URL - note "api" means "internal" to cloudalchemy.grafana but "external" to appliance:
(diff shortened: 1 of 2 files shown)


sjpb:
Not the solution, see comments in #212

PR #214, opened 2022-09-20
Fix templating

sjpb:
Currently `ansible/filter_plugins` is only found from inside roles, which means trying to access hostvars from a top-level playbook (e.g. a debug playbook) fails as it can't find prometheus_node_exporter_targets.


m-bull (approved):
LGTM

PR #215, opened 2022-09-20
Supply OOD defaults for cookie-cutter produced environments

sjpb:
Currently, actually getting a working environment by following the README instructions needs specifying some variables for OOD, and setting up some users. These variables are actually documented OK, but its not clear they do not have defaults.

This changes:
- [x] common env to make it clear when/which OOD variables are required
- [ ] cookiecutter env to add default OOD values (i.e. assuming no FIP on login) and add a testuser
- [x] arcus env to use these defaults (instead of including the same values as arcus-specific group vars)
- [ ] README


sjpb:
Replaced by #345

PR #216, opened 2022-09-26
Support multiple partitions in MPI tests

sjpb:
Adds new `hpctests_partition` var which defaults to default partition, to support running `hpctests` on multiple partitions.

Adds a new non-default partition in CI to check hpctests can cope with multiple partitions.

Also removes the "hpctests" group as that didn't really do anything, as a simplification.


sjpb:
Closing as history a mess.

PR #217, opened 2022-09-27
Support multiple partitions in MPI tests

sjpb:
Ticket: https://stackhpc.atlassian.net/browse/DEV-680

- [x] Adds new `hpctests_partition` var which defaults to default partition, to support running hpctests on multiple partitions.
- [x] Adds a new non-default partition in arcus CI environment to check hpctests can cope with multiple partitions.
- [x] Removes the unnecessary "hpctests" group (as a simplification).
- [x] Shows partition name in hpctests outputs

### DIFF
--- a/ansible/roles/hpctests/tasks/hpl-solo.yml
+++ b/ansible/roles/hpctests/tasks/hpl-solo.yml
@@ -42,8 +42,8 @@
 - debug:
     msg: "Using {{ hpctests_hplsolo_ntasks }} process per node with P={{ hpctests_hplsolo_pq.grid.P }}, Q={{ hpctests_hplsolo_pq.grid.Q }} targeting {{ (hpctests_hpl_mem_frac | float)  * 100 }}% of {{ hpctests_nodeinfo.info['MEMORY'][0] }} MB memory per node, block size (NB) = {{ hpctests_hpl_NB }}, problem size (N) = {{ hpctests_hplsolo_N }}"
 
-- name: Get all nodes
-  shell: "sinfo --Node --noheader --format %N" # TODO: assumes only one partition, although actually excluding nodes not in the default partition should be fine.
+- name: Get all nodes in partition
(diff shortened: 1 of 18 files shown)


ansible/roles/hpctests/tasks/hpl-solo.yml
```diff
@@ -42,8 +42,8 @@
 - debug:
     msg: "Using {{ hpctests_hplsolo_ntasks }} process per node with P={{ hpctests_hplsolo_pq.grid.P }}, Q={{ hpctests_hplsolo_pq.grid.Q }} targeting {{ (hpctests_hpl_mem_frac | float)  * 100 }}% of {{ hpctests_nodeinfo.info['MEMORY'][0] }} MB memory per node, block size (NB) = {{ hpctests_hpl_NB }}, problem size (N) = {{ hpctests_hplsolo_N }}"
 
-- name: Get all nodes
-  shell: "sinfo --Node --noheader --format %N" # TODO: assumes only one partition, although actually excluding nodes not in the default partition should be fine.
+- name: Get all nodes in partition
+  shell: "sinfo --Node --noheader --format %N --partition={{ hpctests_partition }}"
```
m-bull, line 46: Does `sinfo` still truncate outputs in certain circumstances?

m-bull (changes requested):

ansible/roles/hpctests/tasks/hpl-solo.yml
sjpb: Using `--format` is [safe](https://slurm.schedmd.com/sinfo.html#OPT_format):
> Minimum field size. If no size is specified, whatever is needed to print the information will be used.

Using `--Format` is [not](https://slurm.schedmd.com/sinfo.html#OPT_Format)
> The minimum field size. If no size is specified, 20 characters will be allocated to print the information.

PR #219, opened 2022-09-27
Refactor CI

sjpb:
This tidies up the current CI to:
- reduce duplication in workflow file
- make reimaging checks more understandable by promoting them to separate workflow steps, rather than being hidden in some nested ansible. Now much more like what you'd run from the command-line.

This is a preliminary refactoring ahead of bigger changes when generic images are possible.


m-bull (approved):
LGTM

PR #221, opened 2022-10-03
Always skip updating grafana

sjpb:
Grafana is supposed to be pinned at a version. `update_exclude` was set appropriately for the packer builds (tested in CI) but not for a direct update (not tested in CI, but done in Azimuth CaaS deployments).


m-bull (approved):
LGTM

PR #226, opened 2022-10-04
Make pingpong plotting independent of role path

sjpb:
Removes use of `APPLIANCES_REPO_ROOT` which is not available in CaaS appliance.


sjpb:
Fixes #227

PR #229, opened 2022-10-10
Remove env lookup for environ root for hooks

sjpb:
Remove unnecessary duplicate definition of `appliances_environment_root`  in hook- play vars in `ansible/site.yml`. This is already defined in `environments/common/inventory/group_vars/all/defaults.yml`.

Having it defined in play vars is confusing and makes it hard to override e.g. in CaaS.


m-bull (approved):

sjpb:
Interestingly this fails with:
```
ERROR! {{ appliances_environment_root }}/hooks/pre.yml: 'appliances_environment_root' is undefined
```

Presumably inventory has not been read by the time the `import_playbook` is done in site.yml.

PR #231, opened 2022-10-19
Fix dashboard idempotency

sjpb:
Currently running `ansible-playbook ansible/monitoring.yml --tags grafana` repeatedly flip-flops between deploying and removing the grafana dashboards. This PR fixes that.

Not picked up by CI as CI cannot check the grafana UI directly - there is no easy way to scrape pages due to their use of javascript.

Note that the logic in `grafana-dashboards/tasks/main.yml` is based on https://github.com/cloudalchemy/ansible-grafana/blob/master/tasks/dashboards.yml#L108, but modified to also allow dashboards to be deployed from local files. However the result of `_dashboards_copied` in the current code is not the same as in that "upstream" version due to changes to paths, which has caused this problem.

### DIFF
--- a/ansible/roles/grafana-dashboards/tasks/main.yml
+++ b/ansible/roles/grafana-dashboards/tasks/main.yml
@@ -127,14 +127,14 @@
     mode: 0640
   notify: restart grafana
 
-- name: Register previously copied dashboards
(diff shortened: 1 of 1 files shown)


m-bull (approved):
Good spot

PR #238, opened 2022-11-23
Template terraform using a basic inventory

sjpb:
Currently the "inputs" to the cluster are split between TF and ansible variables. Moving everything to ansible - by templating TF from ansible - unifies inputs, avoiding problems where e.g. userdata really needs inputs from both ansible and terraform.

Two differences from the CaaS approach (which I'd envisage moving to this once AWX goes and CaaS is merged with this):
1. The files `environments/{common,arcus}/cluster.yml` define the infrastructure, and are actual ansible inventory files. This:
  - Gives a simple/single place to define the cluster for an environment. 
  - Means hosts are defined _before_ any provisioning, which means group and hostvars can be used when templating terraform (rather than just `all` groupvars). This enables simple variation between hosts e.g. for volumes, userdata, networks/ports. It will in future (with some changes to the OpenHPC role) allow environment-specific image build without having to provision a cluster first to define e.g. the slurm control nodename.
  - Allows using ansible hostpatterns to define sequential compute names, while still also allowing arbitrary names.
2. The `community.general.terraform` role is wrapped to (by-default) require user confirmation before making changes.

This `refactor/templatef2` branch makes modifications so that inventory_hostname != hostname. Note prom scrapes are still broken/not fixed for this.


sjpb:
Superseded by #240 which uses inventory_hostname and inventory_hostname.clustername.tld to keep inventory_hostnames as resolvable names.

PR #239, opened 2022-11-24
Add freeipa using container

sjpb:
Test deployment: `/home/rocky/slurm-app-noetchosts`


sjpb:
Closed as currently RockyLinux 8.x can't run rootless podman from a system unit with cgroupsv2 as needed for containers with systemd in them

PR #242, opened 2022-12-14
Return pingpong sbatch output if job fails

sjpb:
Currently if hpctests fails the user has to ssh into a node and retreive the sbatch output to find out why. This PR returns that info to the deployhost automatically for pingpong.


m-bull (approved):
LGTM

PR #243, opened 2022-12-20
Automatically generate openondemand_host_regex

sjpb:
Automatically generate `openondemand_host_regex` ([docs](https://osc.github.io/ood-documentation/latest/app-development/interactive/setup/enable-reverse-proxy.html)) defining the nodes which should be proxied by Open Ondemand server.

The default will still need adjusting for non-standard nodename patterns or host configurations but this removes one piece of non-obvious config needed when enabling Open Ondemand using the cookiecutter environment.


environments/common/inventory/group_vars/all/openondemand.yml
```diff
@@ -7,6 +7,11 @@
 
 # openondemand_servername: '' # Must be defined when using openondemand
 
+# regex defining hosts which openondemand can proxy; the default regex is compute nodes (for apps) and grafana host.
+# the autogenerated regex may need overriding if compute nodes names do not contain numbers in a consistent position or
+# or include regex special characters.
```
m-bull, line 12: Could you include an example of what the regex looks like in its final state here as well as in the docstring on the filter? Might be helpful for debugging. Not a showstopper though...

m-bull (approved):

sjpb:
Deployed at this commit, confirmed OOD monitoring proxying, shell, jupyter and desktop all work.

PR #247, opened 2023-01-25
Support configuring nameservers and proxies

sjpb:
Adds support to:
- Configure nameservers in `/etc/resolv.conf` via new group/role `resolv_conf`. See [ansible/roles/resolv_conf/README.md](ansible/roles/resolv_conf/README.md).
- Configure proxies for http/https including dnf/yum and processes running in systemd unit files via new group/role `proxy`. See [ansible/roles/proxy/README.md](ansible/roles/proxy/README.md).


sjpb:
In draft pending resolution of https://github.com/stackhpc/ansible-slurm-appliance/issues/257 - this PR will need to ensure that NM control of resolv.conf is disabled and resolv.conf is NOT deleted during cleanup.

ansible/roles/proxy/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
...
+    - key: http_proxy
+      value: "{{ proxy_http_proxy }}"
+    - key: https_proxy
+      value: "{{ proxy_https_proxy }}"
+    - key: no_proxy
+      value: "{{ proxy_no_proxy }}"
+
+- name: Define dnf/yum proxy
+  ini_file:
+    path: /etc/yum.conf
```
m-bull, line 22: Should this be `/etc/dnf/dnf.conf`?

ansible/roles/proxy/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Define http/s proxy configuration.
+
+## Role variables
+
+- `proxy_http_proxy`: Required. Address of http proxy. E.g. "http://10.1.0.28:3128" for a Squid proxy on default port.
+- `proxy_https_proxy`: Optional. Address of https proxy. Default is `{{ proxy_http_proxy }}`.
+- `proxy_no_proxy`: Optional. Comma-separated list of addresses not to proxy. Default is to concatenate `inventory_hostname` (for hostnames) and `ansible_host` (for host IPs) for all Ansible hosts.
+- `proxy_dnf`: Optional bool. Whether to configure yum/dnf proxying through `proxy_http_proxy`. Default `true`.
+- `proxy_systemd`: Optional bool. Whether to give processes started by systemd the above http, https and no_proxy configuration. **NB** Running services will need restarting if this is changed. Default `true`.
```
m-bull, line 11: Should this role have a handler that restarts running services detected by `service_facts`? Could put it behind a switch?

m-bull (changes requested):

ansible/roles/proxy/tasks/main.yml
sjpb: oops, yes! /etc/yum.conf is normally symlinked to /etc/dnf/dnf.conf; apparently changing the former works.

ansible/roles/proxy/README.md
sjpb: - `service_facts.services` has a length of 256 on a running system, which seems very suspicious to me, as if it is not returning all services.

- I don't really want to restart ALL services, that seems horrible. I could reboot?

But really, the issue is that the monitoring podman services may need a proxy working to be able to pull the containers. These services aren't started yet at this point, hence why I didn't actually need a restart. I think it's unlikely that someone would configure a proxy and set `proxy_systemd: false`, run site.yml/fatimage.yml, then set `proxy_systemd: true` and rerun site.yml/fatimage.yml.


Can we discuss?

m-bull: Sounds fine to me, agree that sounds a bit suspect.

m-bull (approved):

PR #248, opened 2023-01-26
Add script to retrieve CI inventory

sjpb:
Enables running ansible against clusters provisioned by CI to help with debugging.

Has no impact on cluster, CI does not need to run/pass.


ansible/ci/retrieve_inventory.yml
```diff
@@ -0,0 +1,32 @@
+# Retrieve inventory from a deployed CI arcus environment by reversing arcus/inventory/hooks/pre.yml
+# Usage example:
+#   ansible-playbook ansible/ci/retrieve_inventory.yml -e cluster_prefix=ci4005969475
+#
+- hosts: localhost
+  become: no
+  gather_facts: no
+  vars:
+    cluster_prefix: "{{ undef(hint='cluster_prefix must be defined') }}" # e.g. ci4005969475
+    cluster_network: WCDC-iLab-60
```
m-bull, line 10: Should this value be a variable?

m-bull (changes requested):

ansible/ci/retrieve_inventory.yml
sjpb: Well, it is. However CI is only running on Arcus so writing it into the playbook makes sense. When we go to ansible driving the TF it could automagically collect it from the cluster config.

m-bull (approved):

PR #250, opened 2023-02-17
Build fat image in appliance

sjpb:
- Builds the "fat" image containing binaries for all nodes (but no configuration) in the appliance itself, making https://github.com/stackhpc/slurm_image_builder redundant.
- Removes environment-specific control and login node builds - use the fat image instead.
- Compute node build is now intended to be performed from the latest fat image, hence removed the `yum update *`.
- New github workflow for "fat" image build.
- Built images now include a shortened git hash in their name to make tracking images easier.
- Minor fixes:
   - Fixed error when `appliances_state_dir` is not defined.
   - Updated flavor names for CI environment following cloud changes.
   - Fixed error in slurm exporter handler when required state is `stopped`.
   - Improved clean-up of build-VM-specific logs etc. in built image.
 
Due to the above changes, CI no longer builds environment-specific images, although the appliance still has the capability to build environment-specific compute images. Instead, in CI a reimage is performed using the same "fat" image and ansible is run (`site.yml`), to check that the cluster's state survives a reimage.


.github/workflows/fatimage.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Get created image name from manifest
+        id: manifest
+        run: |
+          . venv/bin/activate
+          IMAGE_ID=$(jq --raw-output '.builds[-1].artifact_id' packer-manifest.json)
+          while ! openstack image show -f value -c name $IMAGE_ID; do
+            sleep 30
+          done
+          IMAGE_NAME=$(openstack image show -f value -c name $IMAGE_ID)
+          echo "::set-output name=IMAGE_ID::$IMAGE_ID"
```
m-bull, line 63: Think `set-output` is deprecated, might be worth looking for the replacement.

.github/workflows/stackhpc.yml
```diff
@@ -8,12 +8,7 @@ on:
   pull_request:
 jobs:
```
m-bull, line 9: Not directly related to this PR, but we should consider setting up jobs so that they don't run in forks: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository

.github/workflows/stackhpc.yml
```diff
@@ -27,13 +22,13 @@ jobs:
         run: |
           set -x
           mkdir ~/.ssh
-          echo "${${{ matrix.cloud }}_SSH_KEY}" > ~/.ssh/id_rsa
+          echo "${arcus_SSH_KEY}" > ~/.ssh/id_rsa
```
m-bull, line 25: This replacement is `arcus` and the next is `.stackhpc` - is that correct?

m-bull (commented):
Couple of general comments, but think this looks good.

.github/workflows/stackhpc.yml
sjpb: Yes

m-bull (approved):

PR #252, opened 2023-02-24
Support EESSI

JohnGarbutt:
Add support for https://eessi.github.io/docs/, providing additional `lmod` modules.

This will be enabled in the "fat" image and in the `everything` template provided by the cookiecutter template.

Note deployments of more than a handful of nodes should consider providing and configuring a squid proxy for EESSI.

Install is based off this script:
https://github.com/EESSI/eessi-demo/blob/main/scripts/install_cvmfs_eessi_RHEL.sh

### DIFF
--- a/ansible/bootstrap.yml
+++ b/ansible/bootstrap.yml
@@ -99,6 +99,7 @@
 - name: Setup EESSI
   hosts: eessi
   tags: eessi
+  become: true
   tasks:
     - name: Install and configure EESSI
       import_role:
--- a/ansible/bootstrap.yml
+++ b/ansible/bootstrap.yml
@@ -100,6 +100,7 @@
   hosts: eessi
   tags: eessi
   become: true
+  gather_facts: false
   tasks:
     - name: Install and configure EESSI
       import_role:
--- a/ansible/bootstrap.yml
+++ b/ansible/bootstrap.yml
@@ -96,6 +96,14 @@
         tasks_from: config.yml
       tags: config
 
+- name: Setup EESSI
+  hosts: eessi
+  tags: eessi
+  tasks:
+    - name: Install and configure EESSI
+      import_role:
+        name: eessi
+
 - hosts: update
   gather_facts: false
   become: yes
--- a/ansible/ci/check_eessi.yml
+++ b/ansible/ci/check_eessi.yml
@@ -1,6 +1,8 @@
 ---
 - name: Run EESSI test job
   hosts: login # TODO: Limit this to single node when there are multiple?
+  become: true
+  become_user: testuser
   tasks:
     - name: Clone eessi-demo repo
       ansible.builtin.git:
--- a/ansible/ci/check_eessi.yml
+++ b/ansible/ci/check_eessi.yml
@@ -1,20 +1,28 @@
 ---
 - name: Run EESSI test job
-  hosts: login # TODO: Limit this to single node when there are multiple?
-  become: true
-  become_user: testuser
+  hosts: login[0]
+  vars:
+    eessi_test_rootdir: /home/eessi_test
   tasks:
+    - name: Create test root directory
+      file:
+        path: "{{ eessi_test_rootdir }}"
+        state: directory
+        owner: "{{ ansible_user }}"
+        group: "{{ ansible_user }}"
+      become: true
+      
     - name: Clone eessi-demo repo
       ansible.builtin.git:
         repo: "https://github.com/eessi/eessi-demo.git"
-        dest: ~/eessi-demo
+        dest: "{{ eessi_test_rootdir }}/eessi-demo"
 
     - name: Run test job
       ansible.builtin.shell:
         cmd: |
           source /cvmfs/pilot.eessi-hpc.org/latest/init/bash
           srun ./run.sh
-        chdir: ~/eessi-demo/TensorFlow
+        chdir: "{{ eessi_test_rootdir }}/eessi-demo/TensorFlow"
         executable: /bin/bash
       register: job_output
 
--- a/ansible/ci/check_eessi.yml
+++ b/ansible/ci/check_eessi.yml
@@ -1,0 +1,24 @@
+---
+- name: Run EESSI test job
+  hosts: login # TODO: Limit this to single node when there are multiple?
+  tasks:
+    - name: Clone eessi-demo repo
+      ansible.builtin.git:
+        repo: "https://github.com/eessi/eessi-demo.git"
+        dest: ~/eessi-demo
+
+    - name: Run test job
+      ansible.builtin.shell:
+        cmd: |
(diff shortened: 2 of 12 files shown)


environments/common/inventory/groups
```diff
@@ -13,6 +13,10 @@ login
 control
 compute
```
sjpb, line 15: Do we really want EESSI to be non-optional? If we want it to be optional, but deployed "by default" then it should be an empty group here, and its children set as openhpc in `environments/common/layouts/everything`

JohnGarbutt: don't know, that wasn't my intent I guess.

JohnGarbutt: ah, I see, I need a separate group to make it optional. yeah, I should do that.

ansible/bootstrap.yml
```diff
@@ -96,6 +96,15 @@
         tasks_from: config.yml
       tags: config
 
+- name: Setup EESSI
```
sjpb, line 99: Do we actually need facts? If not use `gather_facts: no`  for speed.

ansible/roles/eessi/tasks/main.yaml
```diff
@@ -0,0 +1,22 @@
...
+    disable_gpg_check: true
+- name: Install CVMFS
+  yum:
+    name: cvmfs
+- name: Install EESSI CVMFS config
+  yum:
+    name: https://github.com/EESSI/filesystem-layer/releases/download/latest/cvmfs-config-eessi-latest.noarch.rpm
+    # NOTE: Can't find any docs on obtaining gpg key
+    disable_gpg_check: true
+- name: Add base CVMFS config
```
sjpb, line 15: From previous experience with the openhpc role this will be painful - someone will instantly want to configure something. I suggest we make it more flexible, maybe use the `community.general.ini_file` module to dump a dict with some indirection via a _default and _extra dict to allow easily adding and overriding things (we use this pattern elsewhere), maybe something like:
```yaml
cvmfs_config: "{{ cvmfs_config_default | combine(cvmfs_config_extra) }}"

#where:
cvmfs_config_default:
  CVMFS_CLIENT_PROFILE: single
  CVMFS_QUOTA_LIMIT: "{{ cvmfs_quota_limit_mb }}"
cvmfs_config_extra: {}
```

sjpb: Or maybe its an _overrides given it uses combine(), your call.

ansible/roles/eessi/templates/default.local.j2
```diff
@@ -0,0 +1,2 @@
+CVMFS_CLIENT_PROFILE="single"
```
sjpb, line 1: This looks a bit worrying: https://github.com/cvmfs/cvmfs/blob/10c74bfb3a383709ca2de3e1ed676e6801eafdcb/mount/default.conf#L25

It wouldn't be hard to add a squid proxy (we've already got a containerised role), can we ask what JG thinks.

ansible/roles/eessi/tasks/main.yaml
```diff
@@ -0,0 +1,22 @@
...
+  template:
+    src: default.local.j2
+    dest: /etc/cvmfs/default.local
+# NOTE: Not clear how to make this idempotent
```
sjpb, line 19: Agree :-(

ansible/roles/eessi/tasks/main.yaml
```diff
@@ -0,0 +1,22 @@
+---
+- name: Install CVMFS repo
+  yum:
```
sjpb, line 3: for new code, can we use `dnf` instead of `yum` please?

ansible/roles/eessi/tasks/main.yaml
```diff
@@ -0,0 +1,22 @@
...
+    name: https://ecsft.cern.ch/dist/cvmfs/cvmfs-release/cvmfs-release-latest.noarch.rpm
+    # NOTE: Can't find any docs on obtaining gpg key
+    disable_gpg_check: true
+- name: Install CVMFS
```
sjpb, line 7: Add spaces between tasks (again, old code doesn't necessarily follow this)

ansible/ci/check_eessi.yml
```diff
@@ -0,0 +1,26 @@
+---
+- name: Run EESSI test job
+  hosts: login # TODO: Limit this to single node when there are multiple?
```
sjpb, line 3: I think you can just use `login[0]` to do this, see https://docs.ansible.com/ansible/latest/inventory_guide/intro_patterns.html#using-group-position-in-patterns. Definitely worth adding to avoid running multiple times.

ansible/ci/check_eessi.yml
```diff
@@ -0,0 +1,26 @@
...
+  become: true
+  become_user: testuser
```
sjpb, line 5: TBH I think I'd be happier decoupling this from the CI-defined `testuser`. We can take the same approach as `hpctests` and just explicitly create a /home/eessi_tests directory owned by `{{ ansible_user }}`, so no become needed. Its easier to run on other environments then.

ansible/bootstrap.yml
sjpb, line 1: I'm not totally certain about the location of this - it feels a bit "early" in the setup process, `bootstrap` was really intended to be about getting a "working" instance, not about generic s/w installs. But it is after proxies and nameservers will be set by PR https://github.com/stackhpc/ansible-slurm-appliance/pull/247, and it automatically does the right thing in the fat image build, so I think it makes sense for now at least.

sjpb (changes requested):

ansible/roles/eessi/tasks/main.yaml
```diff
@@ -0,0 +1,22 @@
...
+  yum:
+    name: cvmfs
+- name: Install EESSI CVMFS config
+  yum:
+    name: https://github.com/EESSI/filesystem-layer/releases/download/latest/cvmfs-config-eessi-latest.noarch.rpm
+    # NOTE: Can't find any docs on obtaining gpg key
```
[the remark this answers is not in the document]
sd109: This link does give us a GPG key for the CVMFS repo which helps, but still can't find one for the actual EESSI package. Maybe it's not so bad here since it's coming directly from GitHub?

sjpb (changes requested):
Role needs a README.md, that's all.

sjpb (dismissed):

sjpb:
TODO:
- needs fat image build running
- needs fat image testing in CI

sjpb:
Fat image build WIP: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/4863217900/jobs/8670617599

sjpb:
Fat image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/4870492195

sjpb (approved):

PR #255, opened 2023-03-16
Use environment hooks when building "fat" image

sjpb:
This allows for customisation of the "fat" image pipeline by running environment-specific hooks. This is useful for cases where deployments want to build their own fat image, and have some specific requirements.

If hooks need to be prevented from running during *any* build, use e.g.:
```yaml
- hosts: compute:!builder 
```

If hooks should only run in the environment-specific compute image build, not the fat image build, use e.g.:
```yaml
- hosts: compute:!control:!login
```
(as the fat image build host is in all 3 groups)

### DIFF
--- a/ansible/fatimage.yml
+++ b/ansible/fatimage.yml
@@ -1,6 +1,20 @@
 # Builder version of site.yml just installing binaries
 
+- name: Run pre.yml hook
+  vars:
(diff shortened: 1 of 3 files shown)


m-bull (approved):
LGTM

PR #264, opened 2023-05-02
Support changing the podman user's uid

sjpb:
Needed for client using DIB-built base image, where the current default `1001` uid is not available.


m-bull (approved):
LGTM

m-bull:
CI fail unrelated.

PR #268, opened 2023-05-04
Allow specifying the packer manifest output path

m-bull:


packer/openstack.pkr.hcl
```diff
@@ -101,6 +101,11 @@ variable "floating_ip_network" {
   default = null
 }
 
+variable "manifest_output_path" {
+  type = string
+  default = null
```
sjpb, line 106: Can we not set a default for this, so we don't have to specify it in every environment? As it is, this PR will break CI.

sjpb (changes requested):

packer/openstack.pkr.hcl
m-bull: Updated.

sjpb (approved):

PR #269, opened 2023-05-04
Support using volume-backed instances for building

m-bull:


packer/openstack.pkr.hcl
```diff
@@ -118,10 +138,14 @@ source "openstack" "openhpc" {
 # NB: build names, split on "-", are used to determine groups to add build to, so could build for a compute gpu group using e.g. `compute-gpu`.
 build {
   source "source.openstack.openhpc" {
```
sjpb, line 140: I'd prefer to put things which are the same in the two build source blocks into the base `source "openstack" "openhpc"` block. I don't think there's any way to pass different values for the same variable for the different builds (i.e. you'd need different variables for the two builds, like with `source_image_name` and `fatimage_source_image_name`), so the way its done currently is just code duplication.

packer/openstack.pkr.hcl
```diff
@@ -142,10 +166,14 @@ build {
 # The "fat" image build with all binaries:
 build {
   source "source.openstack.openhpc" {
+    volume_size = "${var.volume_size}"
+    use_blockstorage_volume = "${var.use_blockstorage_volume}"
+    image_disk_format = "${var.image_disk_format}"
+    metadata = "${var.metadata}"
     floating_ip_network = "${var.floating_ip_network}"
```
sjpb, line 173: Should this be set in the compute build block too?

sjpb (changes requested):

packer/openstack.pkr.hcl
m-bull: Fair - updated.

packer/openstack.pkr.hcl
m-bull: I didn't think it made much sense to have a FIP attached to a compute node in the context of imaging a whole cluster, so I only added to the fatimage builder. I've not used the compute image builder, so don't really know if that is right or not - happy to update if needed, but will add it in a separate PR.

sjpb: The builder isn't really a compute node (just in that group, for this build), so if you need a FIP to access your builder you'll need a FIP here too. But I'm happy to skip that here as I think it's client-specific.

sjpb (dismissed):

sjpb:
Testing fat image build here: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/4892239063

Note this needs to pass before merging this PR, but the appliance does NOT need to be updated to use the result of this fatimage build (and in fact that image should be deleted once built).

sjpb:
Deploy CI failed on:
```
2023-05-05T09:52:04.5929957Z TASK [cloudalchemy.grafana : Enable and start Grafana systemd unit] ************
2023-05-05T09:53:36.0726174Z [0;31mfatal: [ci4892004543-control]: FAILED! => {[0m
2023-05-05T09:53:36.0728396Z [0;31m    "changed": false[0m
2023-05-05T09:53:36.0729041Z [0;31m}[0m
2023-05-05T09:53:36.0729626Z [0;31m[0m
2023-05-05T09:53:36.0730208Z [0;31mMSG:[0m
2023-05-05T09:53:36.0730776Z [0;31m[0m
2023-05-05T09:53:36.0731788Z [0;31mUnable to start service grafana-server: Job for grafana-server.service failed because a timeout was exceeded.[0m
2023-05-05T09:53:36.0735842Z [0;31mSee "systemctl status grafana-server.service" and "journalctl -xe" for details.[0m
2023-05-05T09:53:36.0736703Z [0;31m[0m
```
Going to rerun it.

sjpb (approved):

PR #270, opened 2023-05-05
Default to no_proxy localhost in proxy role

sjpb:
Changes defaults for `proxy` role so that localhost isn't proxied.


m-bull (approved):
LGTM

PR #271, opened 2023-05-05
Add debug logging options for opensearch & filebeat

sjpb:
Set variables `opensearch_debug` and `filebeat_debug` to enable debug-level logging for these containerised services.


m-bull (approved):
LGTM

PR #272, opened 2023-05-05
Always delete resources on deploy failure in CI

sjpb:
The "non enough hosts" check script seems to fail in CI (although it works fine when run against a failed deployment locally). So this just always deletes the CI cluster if deployment fails, to avoid manually having to clean up resources.


m-bull (approved):
LGTM

PR #274, opened 2023-05-10
Use of ephemeral SSH keys when building Packer images

m-bull:
This can save an extra step of creating an SSH key outside of Packer, particularly useful in ephemeral image-build environments.


sjpb (approved):
Fine, looks useful.

PR #276, opened 2023-05-17
Bump ohpc role to allow multiple empty partitions

sjpb:
See https://github.com/stackhpc/ansible-role-openhpc/pull/156


m-bull (dismissed):
LGTM

PR #285, opened 2023-06-21
Install out of tree openstack builder plugin

m-bull:
Hashicorp OpenStack builder plugin is now out-of-tree and doesn't get shipped with Packer by default. List it as an explicit dependency.


sjpb (approved):
LGTM

PR #290, opened 2023-07-21
Switch to the Prometheus community collection

mkarpiarz:
Cloud Alchemy's role has been deprecated in favour of https://github.com/prometheus-community/ansible


sjpb:
Hi @mkarpiarz I'm not totally ignoring this! Clearly all the cloudalchemy roles need to be replaced at some point, but as it is not broken for us it is hard to prioritise on our time. I'll discuss the requirement with @m-bull . However I'd rather replace these roles with containerised services as we use for other Azimuth applications, e.g. for [prometheus here](https://github.com/stackhpc/azimuth-images/blob/main/ansible/roles/linux-monitoring/tasks/prometheus.yml).

PR #292, opened 2023-08-01
Fix opensearch grafana plugin at last working version

sjpb:
Without this PR the grafana plugin installed on a new deployment passes the GUI "test" button but can't actually talk to opensearch. Grafana returns a 500 error and the relevant tile in explore mode shows a JS trace containing `TypeError: pa[t] is undefined`.

Matt A worked out the last working version manually.

This also incorporates https://github.com/stackhpc/ansible-slurm-appliance/pull/293 to set the query type correctly. As noted there, this is because of changes to how querytype is handled by the plugin.

Note that:
- Idempotency was broken before this PR as [this difference](https://github.com/cloudalchemy/ansible-grafana/blob/b77b25607003c4cd9fd03091d03119d1dc8c2700/tasks/plugins.yml#L14) doesn't do what's intended - elements in `.files` are not plugin names but dicts including e.g. file stats.
- At this plugin version given here, the datasource still complains (in the edit DS GUI) about needing a version, but this does not prevent it working.


mkarpiarz:
@sjpb: Could you please check if https://github.com/stackhpc/ansible-slurm-appliance/pull/293 fixes this problem?

Pinning this plugin to a specific version would still be useful to have but, in my opinion, should be handled by the collection.

sjpb:
Cancelled CI as requires #294 to merge first before CI will complete

environments/common/inventory/group_vars/all/grafana.yml
```diff
@@ -74,12 +75,15 @@ grafana_datasources:
...
+      # Have to set flavor and version, but ansible/roles/opensearch/templates/opensearch.yml.j2 fakes version for filebeat
+      # so need to set to fake version here:
+      version: '7.10.2'
+      flavor: elasticsearch
     editable: true
     # readOnly: false
 
 grafana_plugins:
-  - grafana-opensearch-datasource
+  - grafana-opensearch-datasource 2.8.1
```
m-bull, line 86: Does it make sense to expose this as its own variable: `grafana_opensearch_datasource_version` or something?

m-bull (changes requested):

environments/common/inventory/group_vars/all/grafana.yml
sjpb: Personally, I don't think so. It'd just more templating and vars in common, when really we want less there, without actually adding making it more customisable.

sjpb:
@ af1c633492e3dd6c5cd8cfb2a21cdf50fe1a3ce9, confirmed slurm jobs dashboard works.

image

m-bull (approved):

PR #294, opened 2023-08-02
Use Python3.9 for jupyter notebook server

sjpb:
Upgrades the venv used for the OOD jupyter notebook server app to Python3.9 and uses a pip-freeze generated requirements file for the jupyter install. This fixes an issue where pip gets stuck in a dependency resolution loop and times-out (from Ansible's point of view at least) during the jupyter install.


m-bull (approved):
LGTM

PR #296, opened 2023-08-04
Fix SMS ssh

sjpb:
ControlPath was missing from the original smslabs SSH config.

TODO: change environments/skeleton/{{cookiecutter.environment}}/ansible.cfg as well


sjpb:
Without this PR: 2h 12m 47s ([run](https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/5759967085/job/15615018827))

@bae7255:
First run : 57m
Second run: RUNNING

PR #297, opened 2023-08-04
Fix ssh ControlPath in skeleton

sjpb:
Applies fix for CI ssh persistence configuration from https://github.com/stackhpc/ansible-slurm-appliance/pull/296 to skeleton ansible.cfg too.


sjpb:
CI cancelled as it doesn't actually effect CI, the equivalent change has already run successfuly.

sjpb:
CI cancelled as it doesn't actually affect CI, the equivalent change has already run successfuly.

m-bull (approved):
LGTM

PR #298, opened 2023-08-04
New fatimage build on SMS

sjpb:
- Builds new fat image incorporating:
  - #292
  - #294
  - #299
- Note this now only requires an instance with a 12GB root disk.
- Updates `fatimage` CI workflow and `.stackhpc` environment configuration to be able to build images on SMS labs (i.e uses same CI_CLOUD variable as `stackhpc` workflow).
- Updates SMS-labs `.stackhpc` environment CI configuration to use new image. Reverts to non-volume-backed instances, as new image is qcow and small enough for instance root disks.

TODO: upload to arcus S3 once built/pass CI
TODO: merge PR to auto upload to S3 before merging this
TODO: remove the `rebuild_image` now its not needed

PR #300, opened 2023-08-09
Add image upload CI workflow

sjpb:
Allows images built on either CI cloud to be uploaded to Arcus S3. Either triggered manually, or via successful completion of main CI run after merging a PR to `main` branch.

TODO: add secrets to repo before merging

### DIFF
--- a/.github/workflows/image_upload.yml
+++ b/.github/workflows/image_upload.yml
@@ -1,0 +1,53 @@
+name: Upload CI-tested image to Arcus S3
+on:
+  workflow_dispatch:
+  workflow_run:
+    workflows:
+      - Test deployment and reimage on OpenStack
+    types:
+      - complete
+    branches:
+      - main # only want to run after merge has successfully passed CI
+jobs:
+  image_upload:
+    runs-on: ubuntu-20.04
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    concurrency: ${{ github.ref }} # to branch/PR
+    env:
+      ANSIBLE_FORCE_COLOR: True
+      OS_CLOUD: openstack
+      CI_CLOUD: ${{ vars.CI_CLOUD }}
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Record which cloud CI is running on
+        run: |
+          echo CI_CLOUD: ${{ vars.CI_CLOUD }}
+
+      - name: Add bastion's ssh key to known_hosts
+        run: cat environments/.stackhpc/bastion_fingerprints >> ~/.ssh/known_hosts
+        shell: bash
+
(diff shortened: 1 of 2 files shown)


sjpb:
Cancelled CI as doesn't test this PR.

m-bull (changes requested):
I think this needs equivalent manifest build/upload and then manifest promote steps from https://github.com/stackhpc/azimuth-images/pull/28.

Builds on `main` may still fail the functional tests even if they have passed in a feature branch, because package versions can shift underneath. Adding the "promote-manifest-on-tag" functionality means that we can tag only commits on `main` that pass the functional tests, then  the manifest can be consumed by azimuth-ops.

I think its just a few more steps in the `upload_image.yml` playbook to checksum the downloaded image, write a manifest.json that contains the checksum and the image URL and upload the manifest to the same bucket.

Maybe @mkjpryor has thoughts too?

sjpb:
> I think this needs equivalent manifest build/upload and then manifest promote steps from [stackhpc/azimuth-images#28](https://github.com/stackhpc/azimuth-images/pull/28).
> 
> Builds on `main` may still fail the functional tests even if they have passed in a feature branch, because package versions can shift underneath. Adding the "promote-manifest-on-tag" functionality means that we can tag only commits on `main` that pass the functional tests, generate a manifest, then the manifest can be consumed by azimuth-ops.
> 
> I think its just a few more steps in the `upload_image.yml` playbook to checksum the downloaded image, write a manifest.json that contains the checksum and the image URL and upload the manifest to the same bucket.
> 
> Maybe @mkjpryor has thoughts too?

Ah - the entire point here is that we do NOT rerun the build on main. The sequence is:
- In a PR branch:
  - *optionally* the `fatimage` workflow is run, manually triggered, to create a new image, and the image defined for the .stackhpc environment is manually bumped.
  - The `stackhpc` workflow runs on push and tests whatever image is defined
  - PR is approved and merges
 - On main, the `stackhpc` workflow runs again and tests the defined image
 - If that passes, the `image_upload` workflow runs and uploads the image, if not already present, to s3.

So the image build only ever runs in the branch, so it runs once, but it gets tested on main.

sjpb:
I guess I should point out - the above / this PR is just replicating what I'm going to have to do manually before #298 is usable. So I'm open to making subsequent changes for better azimuth integration, but this seems like a good first step to me.

sjpb:
Personally I think it's ok; the image *build* happens on a merge checkout; the image *test* happens both on the merge checkout (in the PR) and then again actually on `main`. And then that same image is used in "production". So we do test what we use. Its a tradeoff re. the already-lengthy 1hr stackhpc CI + whatever the image build is, and trying to avoid redoing the latter.

PR #302, opened 2023-08-16
Pin Terraform in CI to MPL licenced version

sjpb:
See https://www.hashicorp.com/license-faq#products-covered-by-bsl for last MPL version


m-bull (approved):
:(

PR #304, opened 2023-08-29
Pin python package versions and update pips

sd109:
Pin pip-installed versions of pymysql, websockify and pip itself to avoid breakage from new upstream package versions.


sd109:
Manually tested (via CaaS Slurm appliance branch [here](https://github.com/stackhpc/caas-slurm-appliance/compare/main...fix/ood-pip-update)) but happy to wait for SMS to be fixed so that proper CI can run before merging if preferred.

ansible/roles/openondemand/tasks/vnc_compute.yml
```diff
@@ -27,13 +27,15 @@
   # Requires separate step so that the upgraded pip is used to install packages
   pip:
     name: pip
+    state: latest
```
JohnGarbutt, line 30: shouldn't we pin this to a specific version as well? pip can break things.

ansible/roles/openondemand/tasks/vnc_compute.yml
```diff
@@ -27,13 +27,15 @@
   # Requires separate step so that the upgraded pip is used to install packages
   pip:
     name: pip
+    state: 23.2.1
```
sjpb, line 30: Passing a version to `state` is [not documented](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pip_module.html#parameter-state) as working

sjpb (changes requested):

sjpb:
@sd109 which makes me wonder if there should be a `pip_version` default in environments/common/inventory/group_vars/defaults which everything could use?

sd109:
Yeah probably, I'll make sure this CI actually passes then make that change.

sjpb:
Turns out this was fixed by https://github.com/stackhpc/ansible-slurm-appliance/pull/294 but CaaS hadn't been updated to include that.

PR #306, opened 2023-09-05
Use portal-internal network for Arcus CI

sjpb:
Swap Arcus to use portal-internal network as ilab-60 network appears to have a broken router on the default gateway. This means direct-mode ports are no longer tested.


sd109 (approved):
LGTM

PR #315, opened 2023-10-04
Make nvidia-driver install idempotent

sjpb:
Makes the install of the `nvidia-driver` module by the CUDA role idempotent, so that the module's packages are not updated by rerunning the role.

If using the `nvidia-driver:latest-dkms` module stream, running `dnf module install nvidia-driver` will update drivers if more recent ones become available. This is in contrast to non-modular `dnf install $package_name` behaviour, where installed packages are *not* updated by rerunning the command.

This PR changes the role to check if a nvidia-driver module is enabled, and only if not, search for the highest-numbered dkms-variant module stream and enable that.

### DIFF
--- a/ansible/roles/cuda/README.md
+++ b/ansible/roles/cuda/README.md
@@ -2,6 +2,8 @@
 
 Install NVIDIA CUDA. The CUDA binaries are added to the PATH for all users, and the [NVIDIA persistence daemon](https://docs.nvidia.com/deploy/driver-persistence/index.html#persistence-daemon) is enabled.
 
+To avoid unwanted package updates which break functionality, on first use this role enables the dkms-flavour `nvidia-driver` DNF module stream which has the current highest version number. The `latest-dkms` stream is not enabled, and subsequent runs of the role will *not* switch which stream is enabled even if later version streams become available. If an upgrade of the `nvidia-driver` module is required, the currently-enabled stream and all packages should be manually removed.
+
 ## Prerequisites
 
 Requires OFED to be installed to provide required kernel-* packages.
--- a/ansible/roles/cuda/tasks/main.yml
+++ b/ansible/roles/cuda/tasks/main.yml
@@ -17,22 +17,40 @@
     dest: "/etc/yum.repos.d/cuda-{{ cuda_distro }}.repo"
     url: "{{ cuda_repo }}"
 
-- name: Enable nvidia driver module
-  ansible.builtin.command: dnf module enable -y nvidia-driver:latest-dkms
-  register: nvidiadriver_enable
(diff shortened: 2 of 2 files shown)


ansible/roles/cuda/README.md
```diff
@@ -2,6 +2,8 @@
 
 Install NVIDIA CUDA. The CUDA binaries are added to the PATH for all users, and the [NVIDIA persistence daemon](https://docs.nvidia.com/deploy/driver-persistence/index.html#persistence-daemon) is enabled.
 
+To avoid unwanted package updates which break functionality, on first use this role enables the dkms-flavour `nvidia-driver` DNF module stream which has the current highest version number. The `latest-dkms` stream is not enabled, and subsequent runs of the role will *not* switch which stream is enabled even if later version streams become available. If an upgrade of the `nvidia-driver` module is required, the currently-enabled stream and all packages should be manually removed.
```
m-bull, line 5: Should we always install latest? Should the version be parameterised and latest installed if the version variable isn't set?

ansible/roles/cuda/tasks/main.yml
```diff
@@ -17,22 +17,40 @@
...
+  # "nvidia-driver 418-dkms   default [d], fm, ks    Nvidia driver for 418-dkms branch "
+  changed_when: false
+  register: _cuda_driver_module_streams
+  when: "'No matching Modules to list' in _cuda_driver_module_enabled.stderr"
+
 - name: Enable nvidia driver module
-  ansible.builtin.command: dnf module enable -y nvidia-driver:latest-dkms
-  register: nvidiadriver_enable
-  changed_when: "'Nothing to do' not in nvidiadriver_enable.stdout"
+  ansible.builtin.command: "dnf module enable -y nvidia-driver:{{ _cuda_driver_module_streams.stdout_lines | last }}"
```
m-bull, line 37: How does `_cuda_driver_module_streams.stdout_lines | last` match up to the output in the comment above? Doesn't this need to be trimmed?

```
# "nvidia-driver 418-dkms   default [d], fm, ks    Nvidia driver for 418-dkms branch "
```

m-bull (changes requested):

ansible/roles/cuda/README.md
sjpb: Added, needed some logic to avoid it trying to change that, b/c that appears to be unsupported.

ansible/roles/cuda/tasks/main.yml
sjpb: Comment was a bit misleading, ^ is actually ONE line from `dnf module list --enabled ...`

m-bull (approved):
LGTM

PR #323, opened 2023-10-25
DO NOT MERGE: PoC for RL9

sjpb:
This mostly works, but it is not production-ready. Key things:
- podman now won't allow you (by default at least) to use unqualified image names. Fixed by changing to fully-qualified names (and fixing the arcus registry to mirror docker.io)
- couldn't reset the podman database after adding the tmp directory config (see roles/podman/config.yml) - fixed by simply not doing these tasks, I haven't convinced myself this is OK. The original code was specifying `tmp_dir`, which is documented in [containers.conf](https://github.com/containers/common/blob/main/docs/containers.conf.5.md) as having to be on a tmpfs, but that isn't mentioned in [podman docs](https://docs.podman.io/en/stable/markdown/podman.1.html#tmpdir-path). The [rootless tutorial](https://github.com/containers/podman/blob/main/docs/tutorials/rootless_tutorial.md) doesn't mention it but states that "$XDG_RUNTIME_DIRdefaults on most systems to /run/user/$UID", which doesn't exist (with the containers up) for the `podman` user (and woudn't be a tmpfs).

  `podman info` shows
  ```
    graphRoot: /var/lib/podman/.local/share/containers/storage
  ...
    runRoot: /tmp/containers-user-1001/containers
  ```
both of which are mounted on `/`

- podman systemd units complain:
  ```
  time="2023-10-13T13:57:06Z" level=warning msg="The cgroupv2 manager is set to systemd but there is no systemd user session available"
  time="2023-10-13T13:57:06Z" level=warning msg="For using systemd, you may need to login using an user session"
  time="2023-10-13T13:57:06Z" level=warning msg="Alternatively, you can enable lingering with: `loginctl enable-linger 1001` (possibly as root)"
  time="2023-10-13T13:57:06Z" level=warning msg="Falling back to --cgroup-manager=cgroupfs"
  time="2023-10-13T13:57:06Z" level=error msg="unlinkat /run/podman/libpod/tmp: permission denied"
  ```

- the openhpc role has its own PR: https://github.com/stackhpc/ansible-role-openhpc/pull/164. There is some incomplete stuff here (e.g. this PR won't work on RL8) but it also needs the "generic slurm" PR merging so we can define cgroups.conf properly which appears to be necessary. Really I'd like to move the plugin defaults to use cgroups too, but this doens't work in a container (although see OpenHPC slack for a possible workaround).
- monitoring.yml fails b/c there's no `prometheus-slurm-exporter` build for RL9. This is our repo, so it'd presumably be an easy fix.

### DIFF
--- a/ansible/roles/mysql/tasks/install.yml
+++ b/ansible/roles/mysql/tasks/install.yml
@@ -1,6 +1,13 @@
-- name: Install python mysql client
+- name: Install pip
+  dnf:
+    name: python3-pip
+
+- name: Install python mysql client and dependencies
+  # TODO: FIXME: doing this in the system python doesn't seem right
   pip:
-    name: pymysql
+    name:
+      - pymysql
+      - cryptography # avoids error "'cryptography' package is required for sha256_password or caching_sha2_password auth methods"
     state: present
 
 - name: Create systemd mysql container unit file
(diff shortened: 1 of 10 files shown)


sjpb:
Replaced by https://github.com/stackhpc/ansible-slurm-appliance/pull/353

PR #329, opened 2023-11-24
Fix #73: Fails late if no secrets defined

### LINKED ISSUE #73: Fails late if no secrets defined
If `environments/<env>/inventory/group_vars/all/secrets.yml` isn't defined it fails in geerlingguy.mysql/tasks/users.yml with an error which is hidden due to the `no_log: true` on this task. Would be better to fail during validation and more obviosuly - and I think we can require a munge key to be defined for the appliance (even though the role doesn't)

sjpb:
fixes #73

### DIFF
--- a/ansible/roles/passwords/tasks/validate.yml
+++ b/ansible/roles/passwords/tasks/validate.yml
@@ -1,0 +1,4 @@
+- name: Assert secrets created
+  assert:
+    that: (hostvars[inventory_hostname].keys() | select('contains', 'vault_') | length) > 1 # 1 as may have vault_testuser_password defined in dev
+    fail_msg: "No inventory variables 'vault_*' found: Has ansible/adhoc/generate-passwords.yml been run"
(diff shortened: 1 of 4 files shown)


ansible/roles/passwords/tasks/validate.yml
```diff
@@ -0,0 +1,4 @@
+- name: Assert secrets created
+  assert:
+    that: (hostvars[inventory_hostname].keys() | select('contains', 'vault_') | length) > 1 # 1 as may have vault_testuser_password defined in dev
+    fail_msg: "No inventory variables 'vault_*' found: Has ansible/adhoc/generate-passwords.yml been run"
```
sd109, line 4:
```suggestion
    fail_msg: "No inventory variables 'vault_*' found: Has ansible/adhoc/generate-passwords.yml been run?"
```

sd109 (approved):
LGTM

sd109 (approved):

PR #332, opened 2023-11-29
Cope with stale NFS file handles

sjpb:
Updates the `stackhpc.nfs` role to a version which can handle client mountpoints being stale filehandles, e.g. after NFS server reboot while client has directory mounted.


sjpb:
Sorry got overexcited after CI pass. nfs role needs review first.

Issue #327, opened 2023-11-24
Should fix dependency on volume order


sjpb:
Currently, the slurm app TF (and caas) relies ([here](https://github.com/stackhpc/ansible-slurm-appliance/blob/ae1c4d912b7d771d83d9271b107a805eb9dc6adb/environments/skeleton/%7B%7Bcookiecutter.environment%7D%7D/terraform/nodes.tf#L100), [here](https://github.com/stackhpc/ansible-slurm-appliance/blob/ae1c4d912b7d771d83d9271b107a805eb9dc6adb/environments/skeleton/%7B%7Bcookiecutter.environment%7D%7D/terraform/nodes.tf#L127)) on the fact that block devices are attached in the order they are defined; so if labels are applied in that order they end up on the right devices/filesystems. This works for RL8.6 and RL8.8 genericcloud images, actually I don't think it did for RL8.7 and it doesn't for RL9.2. For that, labels are getting applied to the expected devices, it's just that the device ordering isn't consistent with the definition order.

There is a fix at https://github.com/stackhpc/ansible-slurm-appliance/commit/93c70c0175bbfb058e23f51a39e4db174058d46b, which would also mean we could get rid of caas's inspection of image properties to define the correct volume path.

PR #333, opened 2023-11-29
Make volume mounting robust

sjpb:
Use the OpenStack volume ID to find the appropriate block device when labeling filesystem volumes.

Previously labeling relied on the ordering in `/dev` matching the order of the block device definitions in the terraform instance resource. This was the case for RockyLinux 8.8 images but does not appear to be the case for RockyLinux 8.9 images.

Notes:
- The filesystem volume label is taken from the first word of the OpenStack volume description.
- `cloud-init's` `runcmd` happens after mounts, so `bootcmd` must be used to create filesystems.
- The mount options used are the cloud-init `mount` module's defaults, except that `nofail` is *not* specified to avoid hard-to-diagnose problems and the home volume mount declares it is required by the NFS server.
- If the image has `hw_disk_bus='scsi'` and `hw_scsi_model='virtio-scsi'` properties set, entries in `/dev/disk-by-id` contain the full openstack volume ID (plus a prefix). Without these properties, the entries only contain the first 20 characters of the openstack volume ID.

Fixes https://github.com/stackhpc/ansible-slurm-appliance/issues/327

TODO:
- [ ] Fix for RL9.3 images - see https://github.com/systemd/systemd/issues/30246
- [ ] Implement same for ansible-templated-tf in caas environment

### DIFF
--- a/.github/workflows/fatimage.yml
+++ b/.github/workflows/fatimage.yml
@@ -48,6 +48,8 @@ jobs:
           cd packer/
           packer init .
           PACKER_LOG=1 packer build -only openstack.openhpc -on-error=${{ vars.PACKER_ON_ERROR }} -var-file=$PKR_VAR_environment_root/${{ vars.CI_CLOUD }}.pkrvars.hcl openstack.pkr.hcl
+        env:
+          TESTUSER_PASSWORD: ${{ secrets.TEST_USER_PASSWORD }}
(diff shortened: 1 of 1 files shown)


sjpb:
Replaced by https://github.com/stackhpc/ansible-slurm-appliance/pull/341

PR #334, opened 2023-11-29
Don't ignore image changes in skeleton terraform lifecycle

sjpb:
Allows the skeleton terraform to act on image changes. This is because:
- Now state is not in the root disk, being able to reimage with terraform is useful, especially for development.
- No one is using slurm-controlled reimage (which conflicts with terraform tracking compute node usage)
- For the control node, there are issues around lifecycle due to the root disk block device definition also referencing the image.
- For production non-Azimuth use, any change to lifecycle needs some careful consideration anyway.


sd109 (approved):
LGTM

PR #338, opened 2023-11-29
Workaround turbovnc repo bug

sjpb:
Temporary workaround for https://github.com/TurboVNC/turbovnc/issues/387:
- Disable turbovnc repo, if present, very early to allow other dnf uses to succeed.
- Assumes that if the repo exists, host is using the fat image in which case vnc_compute.yml was run during image build and can be skipped.

Note that `dnf clean all` didn't fix the issue.


sjpb:
Closed by https://github.com/stackhpc/ansible-slurm-appliance/pull/339 instead

PR #339, opened 2023-12-01
Use new TurboVNC repofile

sjpb:
TurboVNC repofile has moved. See https://github.com/TurboVNC/turbovnc/issues/387 and the [current install instructions](https://turbovnc.org/Downloads/YUM).


m-bull (approved):
LGTM

PR #346, opened 2024-01-02
Fix removal of packer bundled ansible plugin

sjpb:
Fixes packer builds using packer >= v1.10. In earlier versions ansible plugin was bundled with the packer binary but in https://github.com/hashicorp/packer/releases/tag/v1.10.0 this was removed. This PR adds this as an explicit plugin.


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/7384686816

m-bull (approved):
LGTM

PR #350, opened 2024-01-12
Use most-recent image in skeleton terraform if multiple found

sjpb:
Sometimes terraform apply in the Arcus CI project with the stackhpc environment fails with
```
│ Error: Your query returned more than one result. Please try a more specific search criteria, or set `most_recent` attribute to true.
```
This fixes this by a) making all image parameters use the image uid and b) finding those images using terraform's `most_recent` option.


sd109 (approved):
LGTM

PR #352, opened 2024-01-19
WIP/DNM: Make fat image usable on boot for compute nodes

sjpb:
Things which need to be done on boot to make the fat image usable for a compute node immediately (i.e. no ansible):
- munge key: written to file by ansible
- control name/addr for slurm - written to file by ansible
- FreeIPA: We run the renrolement stuff from Ansible at the moment.
- services start:
   - cuda role (not enabled fat image anyway) would run in full during fat image build sets nvidia-persistenced enabled
   - munge
   - slurmd
   - node-exporter: is already enabled in image build
- etc_hosts: At least for arcus where internal dns doesn't work. And this is hard cause we don’t know the IPs until we have created the hosts (or at least ports)
- also potentially resolv_conf, proxy
- also need to think about systemd role / common defaults

See also https://github.com/stackhpc/ansible-slurm-appliance/compare/main...experiment/bootable-compute


sjpb:
Replaced by compute-init implementation.

PR #355, opened 2024-01-29
Remove the cluster_previous_image workaround

JohnGarbutt:
The caas operator always sends the correct image via
extra_vars, so we can remove the cluster_previous_image
workaround.


sjpb (approved):
LGTM

PR #356, opened 2024-02-07
Require user SSH key to be available in ui-meta

JohnGarbutt:
explanation by @sjpb:
- AWX driver defaulted this true. CaaS operator defaults it to false.
- Hit on new cloud. Hadn't been hit as users/CI user always had a key defined.


sd109 (approved):
LGTM

mkjpryor (approved):
LGTM

PR #357, opened 2024-02-09
Add manila /home specific UI for Arcus

sjpb:
Current Arcus appliances have non-manila home, so don't want to wipe out user data on patch.

### DIFF
--- a/environments/.caas/ui-meta/slurm-infra-manila-home.yml
+++ b/environments/.caas/ui-meta/slurm-infra-manila-home.yml
@@ -1,0 +1,127 @@
+# Exactly as for slurm-infra.yml but to allow for separate manila/non-manila home appliances
+name: "slurm-manila-home"
+label: "Slurm (CephFS home)"
+description: >- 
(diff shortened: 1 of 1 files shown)


environments/.caas/ui-meta/slurm-infra-manila-home.yml
```diff
@@ -0,0 +1,127 @@
...
+    immutable: true
+    options:
+      count_parameter: compute_count
+      min_ram: 2048
+      min_disk: 20
+  
+  - name: home_volume_size
+    label: Home volume size (GB)
+    description: The size of the cloud volume or share to use for home directories.
+    kind: cloud.volume_size
```
mkjpryor, line 58: This is now always a share when we use this ui-meta. We should update the description and make this `integer` so it doesn't contribute to quota calculations.

mkjpryor (changes requested):
Just a small tweak

mkjpryor (approved):
LGTM

PR #358, opened 2024-02-09
Improve CaaS UI descriptions

sjpb:
NB: defaulting to not mentioning manila home in default UI, as how to turn this on is not yet documented .


mkjpryor (approved):
LGTM

PR #359, opened 2024-02-13
Use project manila share user from Azimuth

sjpb:
Changes CaaS project manila shares to use the share user provided by Azimuth (in https://github.com/stackhpc/azimuth/pull/148/files), rather assuming there is only a single access rule. Fixes cases where multiple access rules are defined on the project share.

### DIFF
--- a/environments/.caas/inventory/group_vars/all/manila.yml
+++ b/environments/.caas/inventory/group_vars/all/manila.yml
@@ -8,6 +8,7 @@ caas_manila_home:
 cluster_project_manila_share_name: azimuth-project-share
 caas_manila_project:
   share_name: "{{ cluster_project_manila_share_name }}"
+  share_user: "{{ cluster_project_manila_share_user }}"
   mount_path: /project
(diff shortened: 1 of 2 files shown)


environments/.caas/inventory/group_vars/all/manila.yml
```diff
@@ -8,6 +8,7 @@ caas_manila_home:
 cluster_project_manila_share_name: azimuth-project-share
 caas_manila_project:
   share_name: "{{ cluster_project_manila_share_name }}"
+  share_user: "{{ cluster_project_manila_share_user }}"
```
JohnGarbutt, line 11: I think we need to default these two to None or similar? Because when we only pass them when cluster_project_manila_share = true.

JohnGarbutt (approved):
@sjpb I am hoping this one gets through the testing. This looks OK to me now. And its now consistent with:
https://github.com/stackhpc/caas-workstation/pull/17/files

PR #361, opened 2024-02-17
Fix failure when setting basic_users state: absent

sjpb:
At some point loginctl format changed to include extra fields, some of which may be empty. Luckily we only care about the first two so don't have to parse this properly.

```shell
[rocky@demo-login-0 ~]$ loginctl list-sessions
SESSION  UID USER  SEAT TTY   IDLE SINCE
   1789 1000 rocky      pts/0 no        
   1796 1000 rocky      n/a   no        
   1797 1006 lab0       pts/1 no  
```

### DIFF
--- a/ansible/roles/basic_users/library/terminate_user_sessions.py
+++ b/ansible/roles/basic_users/library/terminate_user_sessions.py
@@ -56,13 +56,11 @@ def run_module():
 
     _, sessions_stdout, _ = module.run_command("loginctl --no-legend list-sessions", check_rc=True)
     for line in sessions_stdout.splitlines():
-        try:
-            session, uid, user = line.split()
-        except ValueError:
-            raise ValueError('failed to split "%s"' % line)
-            
+        session_info = line.split()
+        user = session_info[1]
+        session_id = session_info[0]
         if user == module.params['user']:
-            _, sessions_stdout, _ = module.run_command("loginctl terminate-session %s" % session, check_rc=True)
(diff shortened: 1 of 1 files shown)


ansible/roles/basic_users/library/terminate_user_sessions.py
```diff
@@ -56,13 +56,11 @@ def run_module():
 
     _, sessions_stdout, _ = module.run_command("loginctl --no-legend list-sessions", check_rc=True)
```
m-bull, line 57: `loginctl` supports `--output=json-pretty` - parsing JSON might prevent a breaking format change happening again in the future?

sd109 (approved):
LGTM

m-bull (approved):
This works as-is, so I don't see the point in fussing about the methodology, though I've a minor comment on the use of JSON.

Approved, but also happy to re-review if you want to update to address the comment.

ansible/roles/basic_users/library/terminate_user_sessions.py
sjpb: Should have commented - I found that but it doesn't seem to do anything though for list-sessions at least:
```
[rocky@patches-control ~]$ loginctl --output=json-pretty list-sessions
SESSION  UID USER  SEAT TTY   IDLE SINCE
      1 1000 rocky      pts/0 no        

1 sessions listed.
```

sjpb: The weird thing is that it's not listed in the options for `loginctl --help` but trying something like `loginctl --otuput=madeup` errors, so clearly it is parsing it.

PR #363, opened 2024-02-21
Don't link manila for /home to manila for /project in caas

sjpb:
Enables making the use of manila for /home in CaaS clusters an explicit opt-in via azimuth-ops configuration. This is to avoid the current situation where:

1. Deploy vanilla Azimuth
2. Deploy a Slurm cluster, gets Cinder home
3. Enable the project share support in Azimuth
4. Patch the Slurm cluster - existing data in /home gets deleted

### DIFF
--- a/environments/.caas/inventory/group_vars/all/cluster.yml
+++ b/environments/.caas/inventory/group_vars/all/cluster.yml
@@ -23,4 +23,4 @@ appliances_state_dir: /var/lib/state
 
 # Defaults for caas-provided extravars:
 cluster_project_manila_share: false
-cluster_home_manila_share: "{{ cluster_project_manila_share }}"
+cluster_home_manila_share: false


mkjpryor (approved):
LGTM

sjpb:
Testing at this commit:
- Created azimuth without setting `azimuth_openstack_manila_project_share_gb`
- Created slurm had no /project and /home was cinder - correct
- Reprovisioned azimuth with `azimuth_openstack_manila_project_share_gb: 10`
- New slurm had /project but /home was still cinder - correct

mkjpryor:
I have also verified that the following works:

1. Created azimuth without setting `azimuth_openstack_manila_project_share_gb`
2. Created slurm had no /project and /home was cinder - correct
3. Reprovisioned azimuth with `azimuth_openstack_manila_project_share_gb: 10`
4. Update of existing slurm picks up /project but /home was still cinder - correct

PR #366, opened 2024-02-22
WIP: Add support for appliances_task_mode

sjpb:
Adds a new variable `appliances_task_mode` which controls which taskfiles are run from roles. This takes options `main` and `configure` at present, with the latter usable running on a "fat" image and skipping most install tasks.

Earlier versions provided a ~25% speedup in site.yml, reducing it from 7m15s to 5m21s.

It currently affects:
- eessi
- os-manila-mount (could be neater with some role refactoring)
- opensearch
- filebeat
- slurm_exporter
- mysql
- openhpc
- grafana
- podman
- firewalld
- fail2ban
- cuda
- OOD desktop
- OOD jupyter
- freeipa client

Currently does not affect:
- nfs
- slurm_openstack_tools
- openondemand server
- prometheus
- nodeexporter

This is set to `configure` for the stackhpc CI environment.
 It could be used in caas to make platform creation faster.

Note currently an `appliances_task_mode: install` is NOT supported, and `ansible/fatimage.yml` determines what portions of roles are run.


sjpb:
CI failure above was b/c grafana repos (which are still enabled, despite not doing an install) started failing GPG keys.

sjpb:
This is at least partially supported in the appliance.

PR #367, opened 2024-02-23
Reduce CI resources

sjpb:
Downsize to 2x compute / single partition and smaller control node. Still shows 15 of 20 Gi mem free at idle.

Hoping to reduce contention for CI resources.


sjpb:
Replaced by #369, which also did this

Issue #370, opened 2024-02-28
FreeIPA server fails with referer error


sjpb:
Trying to connect to the web UI or add hosts fails with the following in the httpd logs:
```
ipa: ERROR: Rejecting request with bad Referer https://10.1.112.227/ipa/ui/
```

PR #372, opened 2024-03-01
Fix freeipa server referer errors

sjpb:
Latest freeipa server has additional checks for HTTP_REFERER which means the current workaround isn't sufficient. This PR uses mod_headers to add the correct HTTP_REFERER instead, which is a more-robust/cleaner solution.


sjpb:
Fixes #370

m-bull (approved):
LGTM

PR #375, opened 2024-03-06
Add hpctests_pre_cmd option

sjpb:
Allow adding a command to hpctests sbatch templates before module load, to enable working around situations where paths or module paths are customised.


m-bull (approved):
LGTM

PR #383, opened 2024-03-27
Give caas azimuth user sudo

sjpb:
For the caas environment:
- The user's ssh key is only injected for the `azimuth` user (not `rocky`)
- The deploy key is only injected for the `rocky` user
- The `azimuth` user has passwordless sudo on all nodes
- The `azimuth` user is permitted to ssh into compute nodes even without jobs running

Note that:
- The `azimuth` user is setup via the `basic_users` role instead of cloud-init as ssh keys need to be created on a shared `$HOME` and propagated across the cluster.
- To allow the `azimuth` user to login to the control node, the default shell logic in the `basic_users` role is overriden and NFS configuration is modified so the control node has `/home` mounted.


sjpb:
Checks using Azimuth user:
- can ssh into compute and sudo
- can ssh into control and sudo
- can sudo on login

sjpb:
Redone checks after merge from main changed image.

PR #385, opened 2024-04-05
Use RockyLinux 9 for CI by default

sjpb:
Makes RL9 the default for image builds and test and makes RL9 optional.


m-bull (approved):
LGTM

PR #386, opened 2024-04-18
Fix ControlPath for cookiecutter ansible.cfg

sjpb:
Current cookiecutter ansible.cfg has an error in the ssh options which means that ssh doens't actually work.


m-bull (approved):
LGTM

sd109 (approved):

PR #389, opened 2024-05-14
Fix '400 HTTP Bad Request' when adding IPA hosts to in-cluster dev FreeIPA server

sjpb:
Something has changed in upstream packages. CI not relevant here as the development FreeIPA server is not deployed in CI.


m-bull:
If it works elsewhere then LGTM.

m-bull (approved):

PR #390, opened 2024-05-14
Define compute groups in Tofu

sjpb:
Allow homogenous compute groups to be defined in the skeleton Tofu more naturally and with different vnic types, bindings and image IDs between groups. This makes cookiecutter-produced environments more production-ready.

Also changes volume attach logic to use name rather than description, as description is not available on volume data resources.


m-bull (approved):
LGTM

PR #391, opened 2024-05-15
add storage VLAN interface on all slurm nodes

scrungus:
Additional note from @sjpb :
> The `.caas` environment uses `etc_hosts` to template out /etc/hosts. By default (= .caas environment) it uses `ansible_host` as the IP. This is safe after this IP as `stackhpc.terraform.infra/inventory_adopt` uses the terraform output `node.ip` for this, and this is set in `roles/cluster_infra/templates/outputs.tf` to be network[0] on all nodes. This PR adds the storage network as the *2nd* network so it will not be in /etc/hosts.

### DIFF
--- a/ansible/roles/cluster_infra/templates/resources.tf.j2
+++ b/ansible/roles/cluster_infra/templates/resources.tf.j2
@@ -116,10 +116,12 @@ data "openstack_networking_network_v2" "cluster_external_network" {
   name = "{{ cluster_external_network }}"
 }
 
-# Always get the SRIOV storage network 
-data "openstack_networking_network_v2" "portal_storage_direct" {
-  tags = ["portal-storage-direct"]
+# SRIOV storage network 
+{% if cluster_storage_network is defined %}
+data "openstack_networking_network_v2" "cluster_storage_direct" {
+  name = "{{ cluster_storage_network }}"
 }
+{% endif %}
 
 data "openstack_networking_subnet_ids_v2" "cluster_external_subnets" {
   network_id = "${data.openstack_networking_network_v2.cluster_external_network.id}"
@@ -186,7 +188,7 @@ data "openstack_networking_subnet_v2" "cluster_subnet" {
 # Login node
 ###
 
-# VXLAN
+# Primary network
 resource "openstack_networking_port_v2" "login" {
   name = "{{ cluster_name }}-login-0"
   network_id = "${data.openstack_networking_network_v2.cluster_network.id}"
@@ -212,9 +214,10 @@ resource "openstack_networking_port_v2" "login" {
 }
 
 # Storage VLAN
-resource "openstack_networking_port_v2" "login_portal_storage" {
-  name           = "{{ cluster_name }}-login-0-portal-storage"
-  network_id     = data.openstack_networking_network_v2.portal_storage_direct.id
+{% if cluster_storage_network is defined %}
+resource "openstack_networking_port_v2" "login_storage" {
+  name           = "{{ cluster_name }}-login-storage-0"
+  network_id     = data.openstack_networking_network_v2.cluster_storage_direct.id
   admin_state_up = "true"
 
   # not sure if needed here
@@ -225,15 +228,16 @@ resource "openstack_networking_port_v2" "login_portal_storage" {
 
   # for now, until we set up flavors
   binding {
-    vnic_type = "normal"
+    vnic_type = "{{ cluster_storage_vnic_type | default('normal') }}"
   }
 }
+{% endif %}
 
 ###
 # Control node
 ###
 
-# VLXAN
+# Primary network
 resource "openstack_networking_port_v2" "control" {
   name = "{{ cluster_name }}-control-0"
   network_id = "${data.openstack_networking_network_v2.cluster_network.id}"
@@ -258,10 +262,10 @@ resource "openstack_networking_port_v2" "control" {
 }
 
 # Storage VLAN
-
-resource "openstack_networking_port_v2" "control_portal_storage" {
-  name           = "{{ cluster_name }}-control-0-portal-storage"
-  network_id     = data.openstack_networking_network_v2.portal_storage_direct.id
+{% if cluster_storage_network is defined %}
+resource "openstack_networking_port_v2" "control_storage" {
+  name           = "{{ cluster_name }}-control-storage-0"
+  network_id     = data.openstack_networking_network_v2.cluster_storage_direct.id
   admin_state_up = "true"
 
   # not sure if needed here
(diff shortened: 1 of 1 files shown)


ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -224,7 +257,29 @@ resource "openstack_networking_port_v2" "control" {
...
+  admin_state_up = "true"
+
+  # not sure if needed here
+  security_group_ids = [
+    "${openstack_networking_secgroup_v2.secgroup_slurm_cluster.id}"
+  ]
+
+  # for now, until we set up flavors
+  binding {
+    vnic_type = "normal"
```
JohnGarbutt, line 274: lets add cluster_vnic_type a bit like the existing one:

`vnic_type = "{{ cluster_vnic_type | default('normal') }}"`

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -116,6 +116,11 @@ data "openstack_networking_network_v2" "cluster_external_network" {
   name = "{{ cluster_external_network }}"
 }
 
+# Always get the SRIOV storage network 
+data "openstack_networking_network_v2" "portal_storage_direct" {
+  tags = ["portal-storage-direct"]
```
JohnGarbutt, line 121: Lets make this something like cluster_storage_network, and we skip attaching to it if its not defined via extra vars?

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -201,6 +211,29 @@ resource "openstack_networking_port_v2" "login" {
...
 
+# Storage VLAN
+resource "openstack_networking_port_v2" "login_portal_storage" {
+  name           = "{{ cluster_name }}-login-0-portal-storage"
+  network_id     = data.openstack_networking_network_v2.portal_storage_direct.id
+  admin_state_up = "true"
+
+  # not sure if needed here
+  security_group_ids = [
+    "${openstack_networking_secgroup_v2.secgroup_slurm_cluster.id}",
```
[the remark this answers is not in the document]
sjpb: If we're on OVN we need stateless secgroups to get decent performance. Not sure if that's relevant here?

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -201,6 +211,29 @@ resource "openstack_networking_port_v2" "login" {
   }
 }
 
+# Storage VLAN
+resource "openstack_networking_port_v2" "login_portal_storage" {
+  name           = "{{ cluster_name }}-login-0-portal-storage"
```
sjpb, line 216: what's the `-portal-` bit of the name here?

ansible/roles/cluster_infra/templates/resources.tf.j2
sjpb: Same for other port. And in fact it'd be nice to make it clear that this doesn't apply to the non-storage ports, so something like `cluster_storage_vnic_type`.

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -224,7 +257,29 @@ resource "openstack_networking_port_v2" "control" {
   }
 }
 
+# Storage VLAN
+
+resource "openstack_networking_port_v2" "control_portal_storage" {
+  name           = "{{ cluster_name }}-control-0-portal-storage"
+  network_id     = data.openstack_networking_network_v2.portal_storage_direct.id
```
sjpb, line 264: again why `portal`? `storage` I think makes some sense c.f. something more generic, implies it probably does not provide external connectivity, and hopefully high perf.

scrungus:
`cluster_storage_network` changes here https://github.com/stackhpc/azimuth/pull/236

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -201,6 +213,30 @@ resource "openstack_networking_port_v2" "login" {
...
+  name           = "{{ cluster_name }}-login-storage-0"
+  network_id     = data.openstack_networking_network_v2.cluster_storage_direct.id
+  admin_state_up = "true"
+
+  # not sure if needed here
+  security_group_ids = [
+    "${openstack_networking_secgroup_v2.secgroup_slurm_cluster.id}",
+  ]
+
+  # for now, until we set up flavors
```
JohnGarbutt, line 228: lets remove the comment here, I think what you have is the end state anyway.

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -116,6 +116,13 @@ data "openstack_networking_network_v2" "cluster_external_network" {
   name = "{{ cluster_external_network }}"
 }
 
+# SRIOV storage network 
+{% if cluster_storage_network is defined %}
+data "openstack_networking_network_v2" "cluster_storage_direct" {
```
JohnGarbutt, line 121: lets just call this cluster_storage for now, so we match the cluster_storage_network parameter please?

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -201,6 +213,30 @@ resource "openstack_networking_port_v2" "login" {
...
 }
 
+# Storage VLAN
+{% if cluster_storage_network is defined %}
+resource "openstack_networking_port_v2" "login_storage" {
+  name           = "{{ cluster_name }}-login-storage-0"
+  network_id     = data.openstack_networking_network_v2.cluster_storage_direct.id
+  admin_state_up = "true"
+
+  # not sure if needed here
```
JohnGarbutt, line 223: I think we need it, for the moment.

We might have to eventually make this option, when we test with VF-LAG, etc.

JohnGarbutt (changes requested):
Nice, its looking good.

Mostly just a few nits around removing the comments we don't need, and renaming the network to be called "storage" rather than "storage_direct", just so we are consistent with the param.

scrungus:
`cluster_vnic_profile` also removed here from the primary network as it is no longer used

JohnGarbutt (approved):
I think this looks good now. @sjpb are you happy with this too?

sjpb:
> `cluster_vnic_profile` also removed here from the primary network as it is no longer used

@scrungus why not?

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -193,14 +205,31 @@ resource "openstack_networking_port_v2" "login" {
 
   binding {
     vnic_type = "{{ cluster_vnic_type | default('normal') }}"
-    {% if cluster_vnic_profile is defined %}
```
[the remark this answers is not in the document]
JohnGarbutt: That was from me, we shouldn't ever use vnic_profile any more, its fixed in yoga.

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -193,14 +205,31 @@ resource "openstack_networking_port_v2" "login" {
...
+  name           = "{{ cluster_name }}-login-storage-0"
+  network_id     = data.openstack_networking_network_v2.cluster_storage.id
+  admin_state_up = "true"
+
+  security_group_ids = [
+    "${openstack_networking_secgroup_v2.secgroup_slurm_cluster.id}",
+  ]
+
+  binding {
+    vnic_type = "{{ cluster_storage_vnic_type | default('normal') }}"
```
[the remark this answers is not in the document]
JohnGarbutt: So this one is specific to the storage network. We need normal ports on the cluster network and direct mode ports on the storage network. Not sure this has been tested yet though.

PR #398, opened 2024-06-18
Extend caas manila /home

sjpb:
Mount caas environment's manila `/home` share on control node to permit azimuth user to ssh to the control node.


sjpb:
Closed as not desirable, we should fix the NFS configuration to not mount /home on the controller (as it was before adding sudo)

PR #403, opened 2024-06-27
Allow extending fat images with site-specific groups

sjpb:
Adds a new packer build "source" and makes the groups this is added into controllable. Can be used to just extend an existing fat image, rather than running the entirety of ansible/fatimage.yml.

Extracted from WIP Lustre PR, so it has been tested.


sd109 (approved):
LGTM

PR #412, opened 2024-07-17
Allow items in compute mapping to have different keys 

sjpb:
Without this e.g. specifying `image_id` only in one compute group will raise an error.


m-bull (changes requested):
nit: Does the compute module need a test to check that we're not passing in a string or something that it doesn't know how to interpret?

sjpb:
> nit: Does the compute module need a test to check that we're not passing in a string or something that it doesn't know how to interpret?

That'd be lovely but the terraform here seems mad - `any` is clearly not useful here but I can't work out what the "correct" option is, any suggestions appreciated!

PR #417, opened 2024-07-24
Enable setting per-compute group availability zones

sjpb:
**NB:** This won't work as is, the availability zone host needs to be the node UUID, not the node name.

Adds a new option `availability_zone_prefix` to the compute groups in the skeleton terraform. If set, this is prefixed to the last portion of each compute node name split on "-" and used to set the instance availability zone

This enables scheduling baremetal instances to specific hosts. For example if `openstack baremetal node list` shows baremetal nodes `compute-[0,5]` then the following OpenTofu configuration will result in instances `${cluster_name}-hpc-[0,2]` landing on the first 3x baremetal nodes. This can simplify troubleshooting.

```
# environments/$env/terraform/terraform.tfvars:
...
compute = {
  baremetal = {
   flavor = "baremetal.xlarge"
   availability_zone_prefix = "nova::compute-"
   nodes = ["hpc-0", "hpc-1", "hpc-2"]
  }
}
...
```

### DIFF
--- a/environments/skeleton/{{cookiecutter.environment}}/terraform/compute/nodes.tf
+++ b/environments/skeleton/{{cookiecutter.environment}}/terraform/compute/nodes.tf
@@ -26,6 +26,7 @@ resource "openstack_compute_instance_v2" "compute" {
   image_id = var.image_id
   flavor_name = var.flavor
   key_pair = var.key_pair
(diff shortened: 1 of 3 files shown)


sjpb:
Replaced by #485

PR #420, opened 2024-08-07
Avoid python-openstackclient v7 due to rebuild bug

sjpb:
With the current latest v7.0.0 python-openstackclient the rebuild command in CI fails, see https://bugs.launchpad.net/python-openstackclient/+bug/2076232

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 ansible==6.0.0
 openstacksdk
-python-openstackclient
+python-openstackclient==6.6.1 # v7.0.0 has a bug re. rebuild
 python-manilaclient
 jmespath
 passlib[bcrypt]==1.7.4


sd109 (approved):

requirements.txt
```diff
@@ -1,6 +1,6 @@
 ansible==6.0.0
 openstacksdk
-python-openstackclient
+python-openstackclient==6.6.1 # v7.0.0 has a bug re. rebuild
```
m-bull, line 3: Not important for here, but someone smarter than me once told me to pin versions like this:
```
python-openstackclient<7.0.0
```

because it makes complex environments easier for pip to solve...

m-bull (approved):

requirements.txt
sjpb: good to know, thanks. Looks like the bug report has some traction so I guess we will be changing this soon anyway.

PR #421, opened 2024-08-07
Update hpctests to obey UCX_NET_DEVICES when RoCE devices present

bertiethorpe:
Adds openmpi runtime options to prevent openmpi's own RoCE detection overriding UCX_NET_DEVICES, if set. See https://github.com/openucx/ucx/issues/10049 for more information.


sjpb (changes requested):
Can we add a comment here explaining why these magic works, and pointing to your UCX thread please? Otherwise this will be a head-scratcher for someone!

bertiethorpe (commented):
Given more info

sjpb (approved):

PR #423, opened 2024-08-14
Correct the -only options in the Packer README

MoteHue:
The options need to be prefixed with ``openstack.``


sjpb (approved):
Correct, thanks.

PR #427, opened 2024-08-14
Build RL8+OFED image in CI

MoteHue:
Uses OFED for both RockyLinux 8 and 9 image builds.

- Fixes fact gathering in the ofed role; facts were only gathered once due to smart gathering which happens before updating host packages. As such, the `ofed_distro_version` was e.g. 8.9 due to the base image despite the dnf update step having upgraded the distro to something later.
- Changes both RL8 and RL9 to LTS version of OFED (now this is supported for RL9).
- Fixes bug where package update wouldn't correctly trigger a reboot.
- Image build workflow now always runs R8+OFED, R9+OFED, RL9+OFED+CUDA versions.
- Test workflow always runs on RL8+OFED and RL9+OFED images.
- Changes skeleton, caas and stackhpc ansible.cfg to fail if any inventory can't be parsed.


MoteHue:
Might want to switch to forcing a reboot. /var/run/reboot-required isn't populated, despite a slightly newer kernel getting installed:
```
[rocky@openhpc-ofed-rl8-240814-1513-eeb1f9fc ~]$ dnf list --installed kernel
Installed Packages
kernel.x86_64                                                                                                                                         4.18.0-553.el8_10                                                                                                                                              @anaconda
kernel.x86_64                                                                                                                                         4.18.0-553.16.1.el8_10                                                                                                                                         @baseos  
```

jovial:
> Might want to switch to forcing a reboot. /var/run/reboot-required isn't populated, despite a slightly newer kernel getting installed:
> 
> ```
> [rocky@openhpc-ofed-rl8-240814-1513-eeb1f9fc ~]$ dnf list --installed kernel
> Installed Packages
> kernel.x86_64                                                                                                                                         4.18.0-553.el8_10                                                                                                                                              @anaconda
> kernel.x86_64                                                                                                                                         4.18.0-553.16.1.el8_10                                                                                                                                         @baseos  
> ```



> Might want to switch to forcing a reboot. /var/run/reboot-required isn't populated, despite a slightly newer kernel getting installed:
> 
> ```
> [rocky@openhpc-ofed-rl8-240814-1513-eeb1f9fc ~]$ dnf list --installed kernel
> Installed Packages
> kernel.x86_64                                                                                                                                         4.18.0-553.el8_10                                                                                                                                              @anaconda
> kernel.x86_64                                                                                                                                         4.18.0-553.16.1.el8_10                                                                                                                                         @baseos  
> ```

If I understood your point correctly, you were suggesting that why might want to reboot prior to building ofed? I don't think this is necessary as we can compile the kernel modules for a kernel we are not running (as long as we have the kernel headers). The new kernel and modules will be loaded when we boot the new image.

Or is this  during the ansible run of site.yml afterwards?

sjpb:
> necessary as we can compile the kernel modules for a kernel we are not running (as long as we have the kernel headers). The new kernel and modules will be loaded when we boot the new image.

Hmm, possibly, but to me that seems too risky/confusing. Especially given we may be e.g. installing CUDA afterwards which has kernel-* and ofed dependencies. We have to reboot on selinux changes anyway, so we are expecting a reboot somewhere during the build process. We just don't want to do it twice, if possible.

I think doing this is appropriate:
1. rebooting if the update step has changed, rather than just if the reboot sentinel file exists (i.e. don't trust latter as it's been wrong at least once, probably a packaging bug/oversight I guess)
2. always dumping facts (b/c meta: can't be conditional)
3. always rerunning setup (b/c 2.)

sjpb:
Actually there's another reason: boostrap.yml runs both from site.yml and from fatimage.yml, which is good (as it reduces code duplication). So really, if we need a reboot we should do it, rather than having to determine we're running in a build and therefore can ignore that as we will get a "reboot" when creating an instance with the image ...

Make sense?

jovial:
> Actually there's another reason: boostrap.yml runs both from site.yml and from fatimage.yml, which is good (as it reduces code duplication). So really, if we need a reboot we should do it, rather than having to determine we're running in a build and therefore can ignore that as we will get a "reboot" when creating an instance with the image ...
> 
> Make sense?

Yep, I missed that you run this against live nodes too. That makes more sense. Isn't /var/run/reboot-required a Debian/Ubuntuism? Redhat seem to suggest using: https://access.redhat.com/solutions/27943

ansible/bootstrap.yml
```diff
@@ -226,25 +226,22 @@
...
-        msg: "setstatus:{{ (sestatus.reboot_required | default(false)) }} packages: {{ (update_reboot_required.stat.exists | bool) }}"
-    - name: Reboot if required from SELinux state change or package upgrades
+    - name: Reboot to cover SELinux state change or package upgrades
       reboot:
         post_reboot_delay: 30
-      when: (sestatus['reboot_required'] | default(false)) or (update_reboot_required.stat.exists | bool)
     - name: Wait for hosts to be reachable
       wait_for_connection:
         sleep: 15
-    - name: update facts
```
jovial, line 242: Do we still need to refresh the facts? Not sure if they contain stuff like running kernel version?

environments/skeleton/{{cookiecutter.environment}}/ansible.cfg
```diff
@@ -13,3 +13,7 @@ filter_plugins = ../../ansible/filter_plugins
 [ssh_connection]
 ssh_args = -o ControlMaster=auto -o ControlPath=~/.ssh/%r@%h-%p -o ControlPersist=240s -o PreferredAuthentications=publickey -o UserKnownHostsFile=/dev/null
 pipelining = True
+
+[inventory]
+# Fail when any inventory source cannot be parsed.
+any_unparsed_is_failed = True
```
jovial, line 19: Dumb question. Looks reasonable, but how does this relate to OFED? Is it so we fail earlier in the build?

ansible/bootstrap.yml
MoteHue: I don't _think_ we need to, because either the ofed or the ansible_init tasks will gather facts anyway?

jovial: I think the [new tasks ](https://github.com/stackhpc/ansible-slurm-appliance/pull/427/files#diff-7bc925a3ec3c73e00c9a56a23ae1ebaf83905d08b86cd39569091e1f89b32e42R240) only run if the host is in the ofed group. Should we run them against any host that has rebooted?

MoteHue: Ah I see what you mean, yes that sounds sensible.

environments/skeleton/{{cookiecutter.environment}}/ansible.cfg
MoteHue: It doesn't relate to OFED tbh, I added this to catch the case when you have encrypted secrets in your inventory but don't have ANSIBLE_VAULT_PASSWORD_FILE set. Otherwise, it would fail to parse the inventory and skip a bunch of roles. 

I could split this into a separate PR?

ansible/bootstrap.yml
MoteHue: Made the change

environments/skeleton/{{cookiecutter.environment}}/ansible.cfg
jovial: I'm fine with it here - cheers for clarifying

jovial (dismissed):
Seems to match what steve specified to me :+1:

sjpb:
@MoteHue @jovial this has been approved but the first comment in the PR makes it look like this is draft, can someone clarify/set status/comments appropriately please?

MoteHue:
> @MoteHue @jovial this has been approved but the first comment in the PR makes it look like this is draft, can someone clarify/set status/comments appropriately please?

It's no longer a draft

sjpb:
Build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/10736952422

sjpb:
TODO: check reboot did happen during image build

ansible/bootstrap.yml
```diff
@@ -227,24 +227,24 @@
     - update
   tasks:
     - name: Check for pending reboot from package updates
-      stat: 
-        path: /var/run/reboot-required
+      command:
+        cmd: dnf needs-restarting
```
jovial, line 231: Do we need -r here E.g:
```
[rocky@cclr-dev-control ~]$ sudo dnf needs-restarting -r
No core libraries or services have been updated since boot-up.
Reboot should not be necessary.
```
No idea what it outputs when you do need a reboot mind :D

jovial: OK, man reveals all:
```
DNF-NEEDS-RESTARTING(8)                                                                        dnf-plugins-core                                                                        DNF-NEEDS-RESTARTING(8)

NAME
       dnf-needs-restarting - DNF needs_restarting Plugin

       Check for running processes that should be restarted.

SYNOPSIS
       dnf needs-restarting [-u] [-r]

DESCRIPTION
       needs-restarting  looks  through  running  processes and tries to detect those that use files from packages that have been updated after the given process started. Such processes are reported by this
       tool.

       Note that in most cases a process should survive update of its binary and libraries it is using without requiring to be restarted for proper operation. There are however specific cases when this does
       not apply. Separately, processes often need to be restarted to reflect security updates.

OPTIONS
       All general DNF options are accepted, see Options in dnf(8) for details.

       -u, --useronly
          Only consider processes belonging to the running user.

       -r, --reboothint
          Only report whether a reboot is required (exit code 1) or not (exit code 0).
```

jovial: Just looks like mine has some false positives in stdout:
```
[rocky@cclr-dev-control ~]$ sudo dnf needs-restarting
785 : /sbin/auditd 
1054 : /usr/bin/dbus-daemon --system --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only 
1152 : /usr/lib/polkit-1/polkitd --no-debug 
1367 : /usr/sbin/rpc.statd 
6476 : (sd-pam) 
6559 : /usr/bin/dbus-daemon --session --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only
``` 

or maybe they did get updated :thinking:

jovial: But that conflicts with:
```
[rocky@cclr-dev-control ~]$ sudo dnf needs-restarting
785 : /sbin/auditd 
1054 : /usr/bin/dbus-daemon --system --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only 
1152 : /usr/lib/polkit-1/polkitd --no-debug 
1367 : /usr/sbin/rpc.statd 
6476 : (sd-pam) 
6559 : /usr/bin/dbus-daemon --session --address=systemd: --nofork --nopidfile --systemd-activation --syslog-only 
[rocky@cclr-dev-control ~]$ sudo dnf needs-restarting -r
No core libraries or services have been updated since boot-up.
Reboot should not be necessary.
```

jovial: I'd be tempted to trust the reboot hint - what do you reckon?

jovial (approved):
Logic looks good to me

PR #429, opened 2024-08-20
Automated PRs for version bumps

bertiethorpe:


bertiethorpe:
tested in https://github.com/stackhpc/ansible-slurm-appliance-ci-cd-testing

README.md
```diff
@@ -144,3 +144,22 @@ Please contact us for specific advice, but in outline this generally involves:
...
+Workflow uses [create-pull-request](https://github.com/peter-evans/create-pull-request) to handle the pull request action. See for action inputs.
+
+In order for GitHub actions to fetch workflow changes in `.github/workflows`, a PAT for each deployment must be provided.
+
+The following repository permissions must be set for the PAT:
+ - `Workflows: Read and write`
+ - `Actions: Read and write`
+ - `Pull requests: Read and write`
+
+The PAT should then be copied into an Actions repository secret in the downstream repo with the title `WORKFLOW_TOKEN`.
```
sd109, line 165: I think I'd prefer to make this section just a bullet pointed list of the workflows available and then move these details about workflow permissions and tokens / secrets into the top-level comment of each workflow e.g.

```suggestion
## CI/CD automation

The `.github` directory contains a set of sample workflows which can be used by downstream site-specific configuration repositories to simplify ongoing maintainence tasks. These include:

- An [upgrade check](.github/workflows/upgrade-check.yml.sample) workflow which automatically checks this upstream stackhpc/ansible-slurm-appliance repo for new releases an propose a pull request to the downstream site-specific repo when a new release is published.
```

This is more consistent with the monitoring and logging section above.

.github/workflows/upgrade-check.yml.sample
```diff
@@ -0,0 +1,67 @@
...
+    #   actions: write
+
+    steps:
+      - name: Checkout the config repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+
+      # Based on equivalent GitLab CI job
```
sd109, line 30: Let's change this to say 'based on equivalent azimuth-config job' to make it clear what we based this workflow on. Might also be worth adding a link to that azimuth-config sample workflow.

.github/workflows/upgrade-check.yml.sample
```diff
@@ -0,0 +1,67 @@
...
+  schedule:
+    - cron: "0 9 * * *"
+  workflow_dispatch:
+jobs:
+  check_for_update:
+    runs-on: ubuntu-22.04
+    # permissions:
+    #   contents: write
+    #   pull-requests: write
+    #   actions: write
```
sd109, line 21: I guess we can remove these completely.

README.md
```diff
@@ -144,3 +144,22 @@ Please contact us for specific advice, but in outline this generally involves:
...
+
+A GitHub Actions workflow which checks for new upstream version release tags and creates a PR to update the downstream repo, can be found at:
+
+        .github/workflows/upgrade-check.yml.sample
+
+If activated, the workflow is scheduled by default to run every day at 9 AM UTC and can be triggered manually via the `workflow_dispatch` event. How to activate the workflow is detailed at the top of the file.
+
+Workflow uses [create-pull-request](https://github.com/peter-evans/create-pull-request) to handle the pull request action. See for action inputs.
+
+In order for GitHub actions to fetch workflow changes in `.github/workflows`, a PAT for each deployment must be provided.
```
sd109, line 158: I don't think this is true. The PAT isn't required to *fetch* workflow changes, it's to create PRs which may involve modifying workflows (as far as I understand it anyway). Otherwise, the standard token provided to the workflow by GitHub actions would be sufficient.

We should also rephrase the discussion of tokens to state that both fine-grained tokens and user PAT tokens should work but that it was tested with a PAT token. Fine-grained repo-scoped token being preferred if possible.

sd109 (changes requested):
Technical bits look good to me, just some documentation improvements please.

sd109 (approved):

PR #431, opened 2024-09-04
WIP: Active directory roles

jovial:
Usage is:

- Add relevant hosts to inventory groups: sssd, kerberos_client (if using kerberos), ldap_client (if using ldap)

- If using kerberos, place kerberos configuration here: `environments/<env>/templates/krb5.conf.j2`

- Define sssd config in inventory `(/environments/<env>/inventory/group_vars/all/sssd.yml`):

Note: This needs to be in dictionary format. E.g for LDAP:

```
      sssd_config:
        'sssd':
          'config_file_version': '2'
          'debug_level': '5'
          'reconnection_retries': '3'
          'services': 'nss, pam'
          'domains': 'cam'
        'domain/example':
          'auth_provider': 'ldap'
          'ldap_id_use_start_tls': 'False'
          'chpass_provider': 'ldap'
          'cache_credentials': 'True'
          'krb5_realm': 'EXAMPLE.COM'
          'ldap_search_base': "dc=example,dc=com"
          'id_provider': 'ldap'
          'ldap_uri': "ldaps://ldap.example.com"
          'krb5_kdcip': 'kerberos.example.com'
          'ldap_enumeration_refresh_timeout': '43200'
          'ldap_purge_cache_timeout': '0'
          'enumerate': 'true'

```

If using kerberos, you will need to join the node into AD via one of:

- realm join 
- `adcli preset-computer` and `adcli join`. 
- `ktadd`
- some other means

You can then place the kerberos keytab in the following path (one per host):

```
./environments/<env>/files/<inventory hostname>/krb5.keytab
```

There is a convenience playbook to collect these keytabs if joining on the host itself (ansible/adhoc/collect-kerberos-keytabs.yml). the workflow is:

- Run site.yml
- It will fail because keytab is missing
- Run realm join
- Collect keytabs
- Re-run site.yml


ansible/bootstrap.yml
```diff
@@ -111,6 +111,22 @@
       register: sestatus
 
 # --- tasks after here require access to package repos ---
+- hosts: cacerts
```
jovial, line 114: Active directory was using a custom CA

ansible/roles/sshd/tasks/install.yml
```diff
@@ -0,0 +1,9 @@
+---
+
+- name: Install dependencies
+  ansible.builtin.package:
```
sjpb, line 4: can use dnf with a list for name to allow proper dependency resolution

ansible/roles/sssd/tasks/install-RedHat.yml
sjpb, line 1: We only support Rocky, how come this is here?

ansible/roles/sssd/vars/Rocky.yml
```diff
@@ -0,0 +1,14 @@
+---
+sssd_packages:
+  - 'sssd'
```
sjpb, line 3: On RL9, package providing sssd appears to be called `sssd-common`. There is no sssd package. And sssd-common, sssd-client, sssd-kcm, sssd-nfs-idmap appear to be installed by default in genericcloud image

ansible/roles/ldap/tasks/install.yml
```diff
@@ -0,0 +1,11 @@
...
+
+
+- name: Install dependencies
+  ansible.builtin.package:
+    name: "{{ item }}"
+    state: present
+  become: true
+  with_items:
+    - sssd-ldap
+    - openldap-clients
```
sjpb, line 11: actually a dep of sssd-ldap, doesn't need to be explicitly specified

environments/.gitignore
sjpb, line 1: Nope, don't want to gitignore these generally, deployments might want to commit these.

ansible/roles/ldap/tasks/install.yml
sjpb: and actually this isn't for ldap generally, only for sssd using ldap

ansible/roles/sssd/tasks/configure.yml
```diff
@@ -0,0 +1,24 @@
...
+  template:
+    src: "{{ sssd_conf_template }}"
+    dest: "{{ sssd_conf_path }}"
+    owner: root
+    group: root
+    mode: 0600
+  when: sssd_config is defined
+  notify: "Restart sssd"
+
+- name: "Check if authconfig needs to be run to configure pam/nsswitch"
```
sjpb, line 12: @jovial do you understand what this is actually doing?

ansible/roles/sssd/tasks/configure.yml
```diff
@@ -0,0 +1,24 @@
...
+  # FIXME(wszumski) --test is no longer supported. This probably isn't doing what we think it is doing
+  # Error: option --test is no longer supported and we cannot continue if it is set.
+  # authconfig is also deprecated. We should switch to authselect.
+  shell: "/usr/bin/test \"$(authconfig {% if sssd_enable_mkhomedir | bool %}--enablemkhomedir {% endif %}--enablesssd --enablesssdauth --test)\" = \"$(authconfig --test)\""
```
sjpb, line 16: Think this is the only part of this role not provided by alternative PR (TBD)

ansible/roles/sssd/vars/Rocky.yml
sjpb: Hmm that isn't right, both packages are available on both OS. Installing sssd installs sssd-common. However at least on RL9 installing sssd-common appears to be enough for ldap auth.

RL9: installing sssd = 49M download
RL9: installing sssd-common = 3.9M download

```
[root@9bb7a8d6a6a5 /]# cat /etc/redhat-release 
Rocky Linux release 8.9 (Green Obsidian)
[root@9bb7a8d6a6a5 /]# dnf info sssd
Last metadata expiration check: 0:00:35 ago on Thu Sep 12 12:58:10 2024.
Available Packages
Name         : sssd
Version      : 2.9.4
Release      : 4.el8_10
...
Description  : Provides a set of daemons to manage access to remote directories and
             : authentication mechanisms. It provides an NSS and PAM interface toward
             : the system and a plug-gable back-end system to connect to multiple different
             : account sources. It is also the basis to provide client auditing and policy
             : services for projects like FreeIPA.
             : 
             : The sssd sub-package is a meta-package that contains the daemon as well as all
             : the existing back ends.
[root@9bb7a8d6a6a5 /]# dnf info sssd-common
Last metadata expiration check: 0:00:39 ago on Thu Sep 12 12:58:10 2024.
Available Packages
Name         : sssd-common
Version      : 2.9.4
Release      : 4.el8_10
...
Description  : Common files for the SSSD. The common package includes all the files needed
             : to run a particular back end, however, the back ends are packaged in separate
             : sub-packages such as sssd-ldap.

File: environments/common/inventory/groups [hunk 7]
```diff
@@ -134,4 +134,27 @@ freeipa_client
...
+
+[sssd:children]
+kerberos_client
+ldap_client
+
+[cacerts:children]
+# Adds certificates under APPLIANCES_ENVIORNMENT_ROOT/cacerts to system trust store.
+openhpc
+
+[sshd:children]
```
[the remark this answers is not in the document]
sjpb: Also needs adding to the everything template, empty

ansible/roles/sshd/tasks/runtime.yml
```diff
@@ -0,0 +1,12 @@
+---
+
+- name: Disallow SSH password authentication
+  lineinfile:
+    dest: /etc/ssh/sshd_config
```
sjpb, line 5: This doesn't work on RL9 - there's a /etc/ssh/sshd_config.d/50-cloud-init.conf which wins.

jovial: darn it - a drop in config file does sound nicer

sjpb:
Replaced by https://github.com/stackhpc/ansible-slurm-appliance/pull/438

PR #435, opened 2024-09-09
Dev script - Extract fatimage.yml logs to analyse packer build times

bertiethorpe:
Python script for dev:
Extracts packer build ansible tasks from fatimage.yml logs to CSV.

TODO: Retrieve logs with GH CLI so we don't have to download logs manually.

### DIFF
--- a/.github/workflows/stackhpc.yml
+++ b/.github/workflows/stackhpc.yml
@@ -6,20 +6,20 @@ on:
     branches:
       - main
     paths:
-      - **
-      - !dev/**
-      - dev/setup-env.sh
-      - !docs/**
-      - !README.md
-      - !.gitignore
+      - '**'
+      - '!dev/**'
+      - 'dev/setup-env.sh'
+      - '!docs/**'
+      - '!README.md'
+      - '!.gitignore'
   pull_request:
     paths:
-      - **
-      - !dev/**
-      - dev/setup-env.sh
-      - !docs/**
-      - !README.md
-      - !.gitignore
+      - '**'
+      - '!dev/**'
+      - 'dev/setup-env.sh'
+      - '!docs/**'
+      - '!README.md'
+      - '!.gitignore'
 jobs:
   openstack:
     name: openstack-ci
--- a/.github/workflows/stackhpc.yml
+++ b/.github/workflows/stackhpc.yml
@@ -7,12 +7,14 @@ on:
       - main
     paths-ignore:
       - dev/**
+      - !dev/setup-env.sh
       - docs/**
       - README.md
       - .gitignore
   pull_request:
     paths-ignore:
       - dev/**
+      - !dev/setup-env.sh
       - docs/**
       - README.md
       - .gitignore
--- a/.github/workflows/stackhpc.yml
+++ b/.github/workflows/stackhpc.yml
@@ -5,19 +5,21 @@ on:
   push:
     branches:
       - main
-    paths-ignore:
-      - dev/**
-      - !dev/setup-env.sh
-      - docs/**
-      - README.md
-      - .gitignore
+    paths:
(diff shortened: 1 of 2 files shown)


bertiethorpe:
Checks cancelled because don't test changes

dev/extract_logs.py
```diff
@@ -0,0 +1,65 @@
+import csv
```
sd109, line 1: Could we add a comment to the top of this file explaining what the script does and how to use it please? Similar to the format of the other python scripts in this folder. Other than that, LGTM and is probably a useful script to have in the repo.

dev/extract_logs.py
```diff
@@ -0,0 +1,65 @@
...
+    with open(output_csv_path, 'w', newline='') as csvfile:
+        csvwriter = csv.writer(csvfile)
+        csvwriter.writerow(['Task Name', 'Task Path', 'Time to Complete'])
+        csvwriter.writerows(data)
+
+    print(f"Data extracted, sorted, and saved to {output_csv_path}")
+    
+log_file_path = './RL9-ofed-fatimage-177.txt' # Input workflow log name
+output_csv_path = 'RL9-ofed-fatimage-177.csv' # Output CSV name
+target_directory = '/ansible/' # Shared directory for task path
```
sd109, line 63: Probably slightly nicer if the file path can be provided as a command line argument to the script using `sys.argv` and preferably a helpful error message if you don't give it an arg. So something like
```suggestion
if len(sys.argv) != 2:
    print("Path to workflow log plain text file should be provided as the only arg to this script")
    sys.exit(1)
log_file_path = sys.argv[1] # Input workflow log name
output_csv_path = log_file_path.replace('.txt.', '.csv') # Output CSV name
target_directory = '/ansible/' # Shared directory for task path
```
You'll also need to add `import sys` to top of script for this.

sd109 (changes requested):

.github/workflows/stackhpc.yml
```diff
@@ -5,7 +5,17 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - dev/**
```
sjpb, line 9: I don't think this is correct as dev/setup-env.sh it is used in CI, deliberately, so we test setting up the cluster. See https://github.com/stackhpc/ansible-slurm-appliance/blob/main/.github/workflows/stackhpc.yml#L47

.github/workflows/stackhpc.yml
```diff
@@ -5,7 +5,17 @@ on:
...
+      - docs/**
+      - README.md
+      - .gitignore
   pull_request:
+    paths-ignore:
+      - dev/**
```
sjpb, line 15: as above

sjpb (changes requested):

.github/workflows/stackhpc.yml
sd109: Nice catch, thanks

sd109: @bertiethorpe Would you mind updating this to use something like:
```suggestion
     paths:
       - !dev/**
       - dev/setup-env.sh
```
to capture this requirement correctly? Relevant docs are [here](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) and a useful example [here](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths).

sd109 (approved):
LGTM, nice work

sjpb:
@sd109 @bertiethorpe this still has a TODO in the first comment? If this isn't going to be done can you just rephrase this as a limitation please?

PR #437, opened 2024-09-13
Bump Zenith client to latest from azimuth-cloud namespace

m-bull:


sd109 (approved):
LGTM assuming CI passes, though I guess Slurm appliance CI isn't relevant here right? Assuming Azimuth CI includes CaaS Slurm test then it's that CI that we care about.

PR #441, opened 2024-09-19
Install k3s cluster with ansible init

wtripp180901:
NB: currently doesn't support multiple control nodes (needs support for multiple k3s servers, some stuff to do that [here](https://github.com/k3s-io/k3s-ansible/blob/master/roles/k3s_server/tasks/main.yml)

Install k3s cluster on nodes with control plane on control node. Adds a k3s role to run at bootstrap which installs k3s, k9s and helm as well as an ansible-init playbook to start each node's k3s service as an agent or server depending on metadata specified in terraform.


environments/skeleton/{{cookiecutter.environment}}/terraform/compute/variables.tf
```diff
@@ -67,3 +67,8 @@ variable "root_volume_size" {
 variable "security_group_ids" {
     type = list
 }
+
+variable "k3s_token" {
+    description = "Random cryptographically secure string for K3s token (must be set by ../compute.tf)"
```
sjpb, line 72: I don't think we need descriptions in this file, where the vars are just passed through from the outer module

ansible/roles/passwords/tasks/main.yml
```diff
@@ -7,6 +7,19 @@
   delegate_to: localhost
   run_once: true
 
+- name: Generate k3s token
+  ansible.builtin.set_fact:
+    k3s_token_secret: ""
+
+- name:  Generate k3s token and add to terraform
```
sjpb, line 14: If you see https://developer.hashicorp.com/terraform/language/values/variables#assigning-values-to-root-module-variables, you will see that any vars file ending `.auto.tfvars.json` or `.auto.tfvars` gets automatically loaded.

So rather than regex replacing something in the git-controlled .tfvars file, lets instead *template* a file here like e.g. `ansible-provided.auto.tfvars.json` and template something out using ` ... | to_nice_json` ?

That provides a more general mechanism to set terraform vars from ansible.

Use json so you don't have to template HCL.

ansible/roles/passwords/tasks/main.yml
```diff
@@ -7,6 +7,19 @@
...
+  vars:
+    token: "{{ lookup('ansible.builtin.password', '/dev/null', length=64) }}"
```
sjpb, line 16: Nope. Lets create this in the same place as the other secrets, ansible/roles/passwords/defaults/main.yml

environments/common/layouts/everything
```diff
@@ -81,4 +81,8 @@ openhpc
 
 [ansible_init:children]
 # Hosts to run ansible-init
-cluster
\ No newline at end of file
+cluster
+
+[k3s:children]
+# Hosts to run k3s server/agent
+cluster
```
sjpb, line 88: Lets change this to `openhpc`. There are cases where the "cluster" (i.e. hosts controlled by this ansible) is wider e.g. extra hosts added to do ... stuff. Bit of a subtle argument TBH.

ansible/bootstrap.yml
```diff
@@ -258,3 +258,9 @@
   tasks:
     - include_role:
         name: azimuth_cloud.image_utils.linux_ansible_init
+
+- hosts: k3s
```
sjpb, line 262: This PR needs an image with this included - although I know cross-cloud image handling is hard at present.

ansible/bootstrap.yml
```diff
@@ -258,3 +258,9 @@
...
+  become: yes
```
sjpb, line 263:
```suggestion
  become: yes
  tags: k3s
```

ansible/cleanup.yml
```diff
@@ -38,3 +38,8 @@
 
 - name: Cleanup /tmp
   command : rm -rf /tmp/*
+
+- name: Delete ansible-init sentintel file created during build
```
sjpb, line 42:
```suggestion
- name: Delete ansible-init sentinel file created if ansible-init has run during build
```

ansible/roles/k3s/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+  ansible.builtin.shell:
+    cmd: /usr/bin/k3s-install.sh
+  environment:
+    INSTALL_K3S_VERSION: "v1.31.0+k3s1"
+    INSTALL_K3S_EXEC: "{{ item }}"
+    INSTALL_K3S_SKIP_START: "true"
+    INSTALL_K3S_SKIP_ENABLE: "true"
+    INSTALL_K3S_BIN_DIR: "/usr/bin"
+  changed_when: true
+  with_items:
```
sjpb, line 22: I think generally accepted practice is to use `loop:` now

ansible/roles/k3s/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+    dest: /usr/bin/k3s-install.sh
+    owner: root
+    group: root
+    mode: "0755"
+
+- name: Install k3s
+  ansible.builtin.shell:
+    cmd: /usr/bin/k3s-install.sh
+  environment:
+    INSTALL_K3S_VERSION: "v1.31.0+k3s1"
```
sjpb, line 16: Can we make this version a rolevar please?

ansible/roles/k3s/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+  unarchive:
+    src: https://get.helm.sh/helm-v3.11.0-linux-amd64.tar.gz
+    dest: /usr/bin
+    extra_opts: "--strip-components=1"
+    owner: root
+    group: root
+    mode: 0755
+    remote_src: true
+
+- name: k9s install
```
sjpb, line 36:
```suggestion
- name: Install k9s
```
Just for consistency

ansible/roles/k3s/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+    INSTALL_K3S_EXEC: "{{ item }}"
+    INSTALL_K3S_SKIP_START: "true"
+    INSTALL_K3S_SKIP_ENABLE: "true"
+    INSTALL_K3S_BIN_DIR: "/usr/bin"
+  changed_when: true
+  with_items:
+    - server
+    - agent
+
+- name: Installing helm
```
sjpb, line 26:
```suggestion
- name: Install helm
```

ansible/roles/k3s/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+    owner: root
+    group: root
+    mode: 0755
+    remote_src: true
+
+- name: k9s install
+  ansible.builtin.include_role:
+    name: k9s
+
+- name: Install k3s server ansible-init playbook
```
sjpb, line 40: This isn't just "server", right?

ansible/roles/k3s/tasks/main.yml
```diff
@@ -0,0 +1,43 @@
...
+    dest: /usr/bin
+    extra_opts: "--strip-components=1"
+    owner: root
+    group: root
+    mode: 0755
+    remote_src: true
+
+- name: k9s install
+  ansible.builtin.include_role:
+    name: k9s
```
sjpb, line 38: Hmm. I think this should be separate - so you can add a task to run the k9s role to extras.yml, and in everything groups file default k9s to being control?

ansible/roles/k9s/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
+---
+  - name: Create install directory
+    ansible.builtin.file:
+      path: /root/k9s-temp
```
sjpb, line 4: ideally add owner/group/mode here

ansible/roles/k9s/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+      state: directory
+
+  - name: Download k9s
+    ansible.builtin.get_url:
```
sjpb, line 8: ideally add owner/group/mode here

sjpb: And later in this file...

ansible/roles/k9s/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+      dest: /usr/bin/k9s
+      mode: u+rwx
+      remote_src: yes
+
+  - name: Add k3s kubeconfig as environment variable
+    ansible.builtin.lineinfile:
+      path: /etc/environment
+      line: "KUBECONFIG=/etc/rancher/k3s/k3s.yaml"
+
+  - name: Cleanup k9s install directory
```
sjpb, line 30: nice

sjpb: Actually - this is right behaviour but it' be nice to make it idempotent, i.e. add a while, probaly to a block, so that we only do the download etc etc if its not already installed. Maybe using tmp or a tmpfile plays nicer with that? Not sure

sjpb (changes requested):

environments/skeleton/{{cookiecutter.environment}}/terraform/nodes.tf
```diff
@@ -76,6 +76,9 @@ resource "openstack_compute_instance_v2" "control" {
 
   metadata = {
     environment_root = var.environment_root
+    k3s_token = var.k3s_token
+    k3s_server = "${var.cluster_name}-control"
```
sjpb, line 80: Do we actually need this for the server??

sjpb: actually, we should be able to ditch the agent type key/value and just run an agent if the server is specified?

sjpb:
@bertiethorpe can you review and merge [this](https://github.com/stackhpc/ansible-slurm-appliance/pull/442) into this PR if you're happy pls? Then update the 1st comment above appropriately.

wtripp180901:
build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/10957197822

wtripp180901:
new build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11109726657

wtripp180901:
looks like k3s-agent nodes don't come back up after reboot, don't merge yet

environments/skeleton/{{cookiecutter.environment}}/terraform/variables.tf
```diff
@@ -131,3 +131,7 @@ variable "root_volume_size" {
     type = number
     default = 40
 }
+
+variable "k3s_token" {
```
sjpb, line 135: This does need a description here (doesn't in the compute/variables.tf), including that it is set from the ansible auomatically

ansible/roles/passwords/tasks/main.yml
```diff
@@ -7,6 +7,14 @@
...
 
+- name:  Template k3s token to terraform
+  template:
+    src: k3s-token.auto.tfvars.json
+    dest: "{{ lookup('env', 'APPLIANCES_ENVIRONMENT_ROOT') }}/terraform/k3s-token.auto.tfvars.json"
+  delegate_to: localhost
+  run_once: true
+
+
 # - name: Ensure munge key directory exists
```
sjpb, line 18: Given we're making changes here could you just delete this commented out munge-related code pls?

ansible/roles/k3s/files/start_k3s.yml
```diff
@@ -0,0 +1,42 @@
+- hosts: localhost
+  become: true
+  vars:
+    os_metadata: "{{ lookup('url', 'http://169.254.169.254/openstack/latest/meta_data.json') | from_json }}"
+    k3s_token: "{{ os_metadata.meta.k3s_token }}"
+    k3s_server_name: "{{ os_metadata.meta.k3s_server }}"
+    service_name: "{{ 'k3s-agent' if k3s_server_name is defined else 'k3s' }}"
+  tasks:
+    - name: Add the token for joining the cluster to the environment
+      no_log: false # avoid logging the server token
```
sjpb, line 10: I think `no_log: true` *prevents* logging stuff? can you check please.

sjpb (changes requested):

ansible/roles/k3s/files/start_k3s.yml
```diff
@@ -0,0 +1,42 @@
...
+        path: "/etc/systemd/system/{{ service_name }}.service.env"
+        line: "K3S_TOKEN={{ k3s_token }}"
+
+    - name: Add server url to agents
+      ansible.builtin.lineinfile:
+        path: "/etc/systemd/system/{{ service_name }}.service.env"
+        line: "K3S_URL=https://{{ k3s_server_name }}:6443"
+      when: k3s_server_name is defined
+
+    - name: Add loadbalancer label to server
```
sjpb, line 21:
```suggestion
    - name: Only run loadbalancer on k3s server
    # avoids problems with Ondemand https server
```

ansible/roles/k3s/files/start_k3s.yml
```diff
@@ -0,0 +1,42 @@
...
+      when: k3s_server_name is undefined
+      block:
+        - name: Create override directory
+          ansible.builtin.file:
+            state: directory
+            path: "/etc/systemd/system/{{ service_name }}.service.d"
+        
+        - name: Add label override
```
sjpb, line 29:
```suggestion
        - name: Set loadbalancer label on k3s server
```

sjpb (changes requested):

ansible/roles/k3s/files/start_k3s.yml
wtripp180901: think I forgot to set this back after debugging

ansible/roles/passwords/defaults/main.yml
```diff
@@ -12,4 +12,7 @@ slurm_appliance_secrets:
 secrets_openhpc_mungekey_default:
   content: "{{ lookup('pipe', 'dd if=/dev/urandom bs=1 count=1024 2>/dev/null | base64') }}"
 
+k3s_secrets:
```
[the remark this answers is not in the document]
wtripp180901: I thought maybe it shouldn't be there since the ansible doesn't interact with it after this and it made the templating slightly nice but it works as a key too

wtripp180901:
build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11143143774

ansible/roles/k3s/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+- name: Download k3s install script
+  ansible.builtin.get_url:
+    url: https://get.k3s.io/
+    timeout: 120
+    dest: /usr/bin/k3s-install.sh
+    owner: root
+    group: root
+    mode: "0755"
+
+- name: Install k3s
```
sjpb, line 12: Not idempotent!

sjpb (changes requested):

wtripp180901:
build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11177906960

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11238755140

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11276902293

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11288780863

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11520410025

wtripp180901:
don't merge yet, in an azimuth e2e test grafana (cloudalchemy) has a duplicate uid for some reason

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -397,9 +412,11 @@ resource "openstack_compute_instance_v2" "control" {
   {% else %}
   flavor_id = "{{ control_flavor }}"
   {% endif %}
+  for_each = toset(["control"])
```
sjpb, line 415: Why this change?

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -397,9 +412,11 @@ resource "openstack_compute_instance_v2" "control" {
...
 
   network {
     port = openstack_networking_port_v2.control.id
+    access_network = true
```
sjpb, line 419: OK

ansible/roles/k3s/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+# Warning: changes to these variables won't be reflected in the cluster/image if k3s is already installed
+k3s_version: "v1.31.0+k3s1"
+k3s_selinux_release: v1.6.latest.1
+k3s_selinux_rpm_version: 1.6-1
+rocky_version: "{{ ansible_distribution_major_version }}"
```
sjpb, line 5: Remove this and just use ansible_distribution_major_version in the correct place - this is odd, as it is not a k3s_ prefixed var and I can't see why you would want to override it in its use in ansible/roles/k3s/tasks/main.yml?

ansible/roles/k3s/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
+---
+
+- name: Check for existing k3s installation
+  stat:
+    path: /var/lib/rancher/k3s
+  register: stat_result
+
+- name: Download and air-gapped installation of k3s
```
sjpb, line 8:
```suggestion
- name: Perform air-gapped installation of k3s
```

sjpb, line 8:
```suggestion
- name: Download and air-gapped installation of k3s
  # Using air-gapped install so containers are pre-installed to avoid rate-limiting from registries on cluster startup
```

Is that right?

ansible/roles/k3s/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+      INSTALL_K3S_BIN_DIR: "/usr/bin"
+      INSTALL_K3S_SKIP_DOWNLOAD: "true"
+    changed_when: true
+    loop:
+      - server --disable=traefik
+      - agent
+
+- name: Install helm
+  unarchive:
+    src: https://get.helm.sh/helm-v3.11.0-linux-amd64.tar.gz
```
sjpb, line 61: Can we expose the helm version as a k3s_ role var please?

environments/.caas/hooks/pre.yml
```diff
@@ -1,5 +1,12 @@
 ---
 
+# Generate k3s token
+- name: Generate k3s token
```
sjpb, line 4:
```suggestion
- name: Generate k3s token
  # NB: Although this generates a new token on each run, the actual token set in metadata is retrieved from a set-once tofu resource, hence only the first value ever generated is relevant.
```

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -7,6 +7,19 @@ data "openstack_identity_auth_scope_v3" "scope" {
...
 
+####
+#### Data resources
+####
+
+resource "terraform_data" "k3s_token" {
+    input = "{{ k3s_token }}"
+    lifecycle {
+        ignore_changes = [
+            input,
```
sjpb, line 18:
```suggestion
            input, # makes it a write-once value (set via Ansible)
```

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -386,6 +399,8 @@ resource "openstack_compute_instance_v2" "login" {
         ansible_init_coll_{{ loop.index0 }}_source = "{{ collection.source }}"
       {% endif %}
     {% endfor %} 
+    k3s_server = [for n in openstack_compute_instance_v2.control["control"].network: n.fixed_ip_v4 if n.access_network][0]
```
sjpb, line 402: re. using `openstack_compute_instance_v2.control["control"]`, see comments below about changes to caas resources

sjpb (changes requested):
Nearly there, a few comments though.

sjpb:
@wtripp180901 your initial comment on this PR is I think out of date:
> NB: currently requires working DNS to work without requiring reboots. Can currently be worked around by replacing the values of the k3s_server fields in the terraform metadata with openstack_compute_instance_v2.control["control"].access_ip_v4

docs/k3s.README.md
```diff
@@ -0,0 +1,8 @@
...
+enabled during boot. Nodes with the `k3s_server` metadata field defined will be configured as K3s agents (this field gives them the address of the server). The Slurm control node is currently configured as a server while all other nodes configured as agents. It should be noted that running multiple K3s servers isn't supported. Currently only the root user on the control node has 
+access to the Kubernetes API. The `k3s` role installs Helm for package management. K9s is also installed in the image and can be used by the root user.
+
+# Idempotency
+K3s is intended to only be installed during image build as it is configured by the appliance on first boot with `azimuth_cloud.image_utils.linux_ansible_init`. Therefore, the `k3s` role isn't
+idempotent and changes to variables will not be reflected in the image when running `site.yml`. An additional consequence of this is that for changes to role variables to be correctly applied when extending a base image with a Packer `openhpc-extra` build, the base image must have `ansible-init` installed but not existing K3s instances.
```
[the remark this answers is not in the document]
wtripp180901: i.e a k3s installation which has previously been configured by ansible-init on boot, as you can't install a new k3s over the top of a previous installation that has already ran. Thinking about it this probably wouldn't be an issue anymore after https://github.com/stackhpc/ansible-slurm-appliance/pull/439 so will remove this

ansible/roles/cluster_infra/templates/resources.tf.j2
wtripp180901: I was using the same method to get the control node's IP used as in the changes added in https://github.com/stackhpc/ansible-slurm-appliance/pull/442 to the cookiecutter environment, which require indexed control nodes, otherwise fails with `Because openstack_compute_instance_v2.control does not have "count" or "for_each" set, references to it must not include an index key. Remove the bracketed index to refer to the single instance of this resource.` Could use 
```suggestion
    k3s_server = openstack_compute_instance_v2.control.network[0].fixed_ip_v4
```
instead but probably nice to keep them consistent

ansible/roles/cluster_infra/templates/resources.tf.j2
wtripp180901: see below

ansible/roles/cluster_infra/templates/outputs.tf.j2
```diff
@@ -24,8 +24,8 @@ output "cluster_nodes" {
         }
       },
       {
-        name          = openstack_compute_instance_v2.control.name
-        ip            = openstack_compute_instance_v2.control.network[0].fixed_ip_v4
+        name          = openstack_compute_instance_v2.control["control"].name
```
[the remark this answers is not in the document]
wtripp180901: see below

wtripp180901:
build https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11662901504

ansible/roles/passwords/templates/k3s-token.auto.tfvars.json.j2
```diff
@@ -0,0 +1,3 @@
+{
+    "k3s_token": "{{ slurm_appliance_secrets.k3s_token }}"
```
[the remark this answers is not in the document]
wtripp180901: making it idempotent means that they will still be mismatched on the initial run because ansible hasn't picked up the templated secrets yet because the role is still running, but they will be updated to be the same on subsequent runs. Added a workaround to fix this

wtripp180901:
^ that merge is just updating my local branch with new password role, current images should be fine to merge if no more changes

ansible/roles/cluster_infra/templates/resources.tf.j2
sjpb: Shouldn't make this change

ansible/roles/cluster_infra/templates/resources.tf.j2
sjpb: Lets not make changes to caas if we don't actually need to

ansible/roles/cluster_infra/templates/resources.tf.j2
```diff
@@ -548,6 +566,8 @@ resource "openstack_compute_instance_v2" "{{ partition.name }}" {
...
```
sjpb, line 569: So this will change, b/c only one control node and potentially no access network defined

ansible/roles/cluster_infra/templates/resources.tf.j2
sjpb: - We don't need to keep caas and cookiecutter  consistent if its already different
- caas definitely shouldn't get a set of control nodes (that was a hang-over in cookiecutter from other functionality)
-  I suspect using `openstack_compute_instance_v2.control.network[0].fixed_ip_v4` is fine - it only has 1x network.

ansible/roles/passwords/tasks/main.yml
```diff
@@ -7,14 +7,14 @@
   delegate_to: localhost
   run_once: true
 
-# - name: Ensure munge key directory exists
-#   file:
-#     state: directory
-#     recurse: true
-#     path: "{{ openhpc_passwords_mungekey_output_path | dirname }}"
+- name: Get templated passwords from target environment
```
sjpb, line 10:
```suggestion
- name: Get templated passwords from target environment
# inventory group/host vars created in a play cannot be accessed in the same play, even after meta: refresh_inventory
```

ansible/roles/passwords/tasks/main.yml
```diff
@@ -7,14 +7,14 @@
...
+  ansible.builtin.include_vars:
+    file: "{{ openhpc_passwords_output_path }}"
+    name: templated_secrets
```
sjpb, line 13: Lets remove this so it ends up top-level
```suggestion
```

ansible/roles/passwords/templates/k3s-token.auto.tfvars.json.j2
```diff
@@ -0,0 +1,3 @@
+{
+    "k3s_token": "{{ templated_secrets.vault_k3s_token }}"
```
sjpb, line 2:
```suggestion
    "k3s_token": "{{ vault_k3s_token }}"
```

now top-level as above

sjpb (changes requested):
Lets minimise changes to the caas / cluster_infra TF templates, then call it done.

ansible/roles/k3s/tasks/main.yml
sjpb, line 1: can we change this to install.yml, just makes it clearer

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11799151926

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11859638055

ansible/extras.yml
```diff
@@ -36,3 +36,10 @@
   tasks:
     - import_role:
         name: persist_hostkeys
+
+- name: Install k9s
+  become: yes
+  hosts: k9s
+  tasks:
```
sjpb, line 43: should add a tag here

ansible/roles/k9s/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
+---
+  
+  - name: Check if k9s is installed
+    ansible.builtin.stat:
+      path: "/usr/bin/k9s"
+    register: result
```
sjpb, line 6:
```suggestion
    register: _k9s_stat_result
```

ansible/roles/k9s/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+  - name: Install k9s and clean up temporary files
+    block:
+    - name: Create install directory
+      ansible.builtin.file:
+        path: /tmp/k9s
+        state: directory
+        owner: root
+        group: root
+        mode: "744"
+      when: not result.stat.exists
```
sjpb, line 17:
```suggestion
      when: not _k9s_stat_result.stat.exists
```

ansible/roles/k9s/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+        src: /tmp/k9s/k9s
+        dest: /usr/bin/k9s
+        mode: u+rwx
+        remote_src: yes
+
+    - name: Cleanup k9s install directory
+      ansible.builtin.file:
+        path: /tmp/k9s
+        state: absent
+    when: not result.stat.exists
```
sjpb, line 44:
```suggestion
    when: not _k9s_stat_result.stat.exists
```

ansible/extras.yml
```diff
@@ -36,3 +36,10 @@
   tasks:
     - import_role:
         name: persist_hostkeys
+
+- name: Install k9s
+  become: yes
+  hosts: k9s
```
sjpb, line 42:
```suggestion
  hosts: k9s
  tags: k9s
```

sjpb (changes requested):
LGTM but I have a couple of minor amends you can just accept *if you are happy*, no need to rebuild

sjpb (approved):
LGTM

PR #442, opened 2024-09-19
Remove local DNS as a dependency for k3s

sjpb:
I've tested this and terraform appears to behave properly, i.e. dependency tracking works.

Note using `openstack_compute_instance_v2.control["control"].access_ip_v4` is not, in the general case OK:

> [access_ip_v4](https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/compute_instance_v2#access_ip_v4) - The first detected Fixed IPv4 address.

([source](https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/compute_instance_v2#attributes-reference))

if the terraform is extended with another interface (which it often is) then this might not be right.

### DIFF
--- a/environments/skeleton/{{cookiecutter.environment}}/terraform/nodes.tf
+++ b/environments/skeleton/{{cookiecutter.environment}}/terraform/nodes.tf
@@ -77,7 +77,7 @@ resource "openstack_compute_instance_v2" "control" {
   metadata = {
     environment_root = var.environment_root
     k3s_token = var.k3s_token
-    k3s_server = "${var.cluster_name}-control"
+    k3s_server = "" # think this can go?
     k3s_node_type = "server"
   }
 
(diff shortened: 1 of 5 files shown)


environments/skeleton/{{cookiecutter.environment}}/terraform/nodes.tf
```diff
@@ -77,7 +77,7 @@ resource "openstack_compute_instance_v2" "control" {
   metadata = {
     environment_root = var.environment_root
     k3s_token = var.k3s_token
-    k3s_server = "${var.cluster_name}-control"
+    k3s_server = "" # think this can go?
```
wtripp180901, line 80: yeah you can do checks on if variables are defined so this can go

wtripp180901 (changes requested):

wtripp180901 (approved):
Added some changes to the ansible-init file so it works if the server isn't defined, should be good to merge

PR #445, opened 2024-09-30
Caas updated to use openstack_networking_floatingip_associate_v2

JohnGarbutt:
Before we can move to v3.0.0 we need to use the new floating ip resource, now the compute one has been removed.
To unblock this move, we add a tempory pin to use v2.1.0 for a bit.


sjpb (approved):
LGTM

sjpb:
CI failure is unrelated to change.

PR #449, opened 2024-10-14
Ported monitoring stack to k3s

### LINKED ISSUE #432: SELinux not disabled by default, causes Prometheus install to fail
Produced using Rocky-9-GenericCloud-Base-9.4-20240523.0.x86_64.qcow2 using a custom (non .stackhpc) environment
```
TASK [cloudalchemy.prometheus : Install SELinux dependencies]

failed: [testcluster-control] (item=libselinux-python) => {
    "ansible_loop_var": "item",
    "attempts": 5,
    "changed": false,
    "failures": [
        "No package libselinux-python available."
    ],
    "item": "libselinux-python",
    "rc": 1,
    "results": []
}

MSG:

Failed to install some of the specified packages


failed: [testcluster-control] (item=policycoreutils-python) => {
    "ansible_loop_var": "item",
    "attempts": 5,
    "changed": false,
    "failures": [
        "No package policycoreutils-python available."
    ],
    "item": "policycoreutils-python",
    "rc": 1,
    "results": []
}
```

wtripp180901:
Monitoring stack (prometheus/node exporter/grafana/alertmanager) binary installs removed from site and fatimage, now installs [kube-prometheus-stack Helm chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) into k3s cluster during site run. Containers are pre-pulled by podman and exported into k3s during fatimage build.

As a consequence, the `grafana`, `alertmanager` and `node exporter` groups have been removed and associated roles are now all managed by the `prometheus` role, which is short for kube_prometheus_stack

Also reduced metrics collected by node exporter down to minimal set described in `docs/monitoring-and-logging.README.md`, which was previously unimplemented

Note that because of how OOD's proxying interacts with Grafana's server config and kubernetes, OOD being enabled means that Grafana is only accessible through the OOD proxy. In the caas environment, this means that accessing Grafana requires authenticating with OOD's basic auth. Therefore, accessing Grafana through caas no longer logs you in as the admin user, you instead access the dashboards anonymously

Tests as of [63c3094](https://github.com/stackhpc/ansible-slurm-appliance/pull/449/commits/63c309427a7c750580edb40a05ee4130a7cc3097):

- Prometheus data from cloudalchemy roles successfully migrated to containerised Prometheus, although will likely be under a different job label than the one KPS is hardcoded to use
- Reimage and upgrade from cloudalchemy: done, legacy data needed permission changes
- Final caas test: TODO


wtripp180901:
test image: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11324488678

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11349241640

ansible/fatimage.yml
```diff
@@ -101,12 +101,6 @@
         tasks_from: install.yml
       when: "'filebeat' in group_names"
```
sjpb, line 103: We should pull the containers for kube-prom-stack - maybe add an install.yml in the role which does that.

E.g. see https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/roles/mysql/tasks/install.yml#L18 BUT
note there is a k3s-specific way of using pre-pulled images.

environments/common/inventory/groups
```diff
@@ -36,12 +36,6 @@ mysql
 [prometheus]
 # Single node to host monitoring server.
```
sjpb, line 37:
```suggestion
# Single node to host monitoring stack.
```

sjpb (changes requested):

ansible/roles/kube_prometheus_stack/tasks/main.yml
```diff
@@ -0,0 +1,195 @@
...
+    definition:
+      apiVersion: v1
+      kind: PersistentVolume
+      metadata:
+        name: prometheus-dir
+        labels:
+          app.kubernetes.io/name: prometheus-dir
+      spec:
+        capacity:
+          storage: "{{ prometheus_volume_size }}"
```
sjpb, line 49: This needs fudging somehow, it has no effect and there's really no way to chose it

ansible/roles/kube_prometheus_stack/tasks/main.yml
sjpb, line 1: At some point we need to try to uipgrade a cluster, and see if the exisitng prom state is successfully used.

ansible/roles/kube_prometheus_stack/defaults/main/main.yml
```diff
@@ -0,0 +1,342 @@
...
+kube_prometheus_stack_release_name: kube-prometheus-stack
+
+# The timeout to wait for the release to become ready
+kube_prometheus_stack_wait_timeout: 5m
+
+login_ip: "{{ hostvars[groups['openondemand'][0]]['ansible_host'] }}" # probably needs to be more robust
+control_ip: "{{ ansible_default_ipv4.address| default(ansible_all_ipv4_addresses[0]) }}"
+control_sslip: "{{ control_ip | regex_replace('\\.', '-') }}.sslip.io" 
+
+grafana_volume_size: 10Gi
```
sjpb, line 19: Lets try removing the grafana volume entirely. I'm not convinced this would be ok IF k3s was to start grafana after a rebuild, but at the moment it won't.

sjpb (changes requested):

environments/.stackhpc/terraform/LEAFCLOUD.tfvars
```diff
@@ -1,6 +1,6 @@
 cluster_net = "slurmapp-ci"
 cluster_subnet = "slurmapp-ci"
-control_node_flavor = "ec1.medium" # small ran out of memory, medium gets down to ~100Mi mem free on deployment
+control_node_flavor = "en1.medium" # min 8GB RAM
```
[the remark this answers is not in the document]
wtripp180901: as far as I can tell the SMS and Arcus one's have enough memory already

ansible/roles/kube_prometheus_stack/defaults/main/main.yml
wtripp180901: To be clear, giving grafana a volume is optional and is only used for writing data, not storing grafana's entire state (same case for prometheus). Them actually launching seems to be agnostic of any pre-existing storage volumes

environments/common/inventory/group_vars/all/defaults.yml
```diff
@@ -18,7 +18,7 @@ api_address: "{{ inventory_hostname }}"
 opensearch_address: "127.0.0.1"
 prometheus_address: "{{ hostvars[groups['prometheus'].0].api_address }}"
```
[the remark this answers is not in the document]
wtripp180901: changed this to be more useful to us, there was a bug in the slack receiver I didn't catch so this is used there now

ansible/fatimage.yml
wtripp180901: the issue is that k3s uses an embedded containerd runtime, meaning we can'r pre-pull images without starting k3s, which we don't want to do on build. There is a way to use an external containerd runtime but I can't seem to get it to work. We could potentially pre-pull images with podman and export them as archives, then import into k3s in site.yml, but we probably want to get rid off podman eventually. At the moment all of the image sizes in kube-prometheus-stack sum to less than 1Gb so this might not be worth the effort

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11464464916

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11484750641

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11494840785

ansible/roles/kube_prometheus_stack/tasks/main.yml
wtripp180901: KPS appears to not overwrite existing state, although changes in labels mean that metrics are grouped differently. It does however create a subdirectory of the provided state directory with a hardcoded name, so data will need to be migrated there

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11570541407

bertiethorpe (dismissed):

wtripp180901:
no image changes since last build so last commit should be ready to merge barring review changes

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11782814296

sjpb (changes requested):
Please note in the PR (with better wording!) that
- the "prometheus" group is bascally short for "kube-prometheus-stack" group!
- the monitoring link in CaaS now accesses grafana with anonymous auth (b/c it has to go via OOD), so CaaS users can't change their dashboards
-

ansible/roles/kube_prometheus_stack/defaults/main/install.yml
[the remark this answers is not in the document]
wtripp180901: tested this, looks like they do need to be under the main/ subdir for files other than main.yml to get picked up

sjpb:
@wtripp180901 not a high priority but would be nice to know if this PR reduces the size of the data in the image. And/or whether we can reduce the required root disk size at all - which isn't the same thing, b/c e.g. dnf caches which we throw away require additional size during build

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11909785160

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11932735822

sjpb:
@wtripp180901 when we get back to this we should look at whether https://github.com/stackhpc/ansible-slurm-appliance/blob/main/environments/common/inventory/group_vars/all/systemd.yml needs to be modified.

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/12866628689

wtripp180901:
now prometheus is getting wrong IP for OOD

wtripp180901:
^ might be nice to have some CI to catch targets being down from prometheus' POV

sjpb:
@wtripp180901 also re TASK [systemd : Add dropins for unit files]: - its a bit worrying that this is a change when runnign site.yml:
```
Wednesday 22 January 2025  16:54:02 +0000 (0:00:01.746)       0:00:15.618 ***** 
changed: [RL9-control] => (item={'key': 'opensearch', 'value': {'group': 'opensearch', 'content': '[Unit]\nRequiresMountsFor=/var/lib/state\n'}})
changed: [RL9-control] => (item={'key': 'grafana-server', 'value': {'group': 'grafana', 'content': '[Unit]\nRequiresMountsFor=/var/lib/state\n'}})
changed: [RL9-control] => (item={'key': 'slurmdbd', 'value': {'group': 'openhpc', 'content': '[Unit]\nRequiresMountsFor=/var/lib/state\n'}})
changed: [RL9-control] => (item={'key': 'slurmctld', 'value': {'group': 'openhpc', 'content': '[Unit]\nRequiresMountsFor=/var/lib/state\n'}})
changed: [RL9-control] => (item={'key': 'prometheus', 'value': {'group': 'prometheus', 'content': '[Unit]\nRequiresMountsFor=/var/lib/state\n'}})
```

if we're using /var/lib/state for the kube-prom-stack, we should ensure relevant dropins are installed *in the image*, so that when k3s comes up on boot it waits until the vol is mounted before starting :grimacing:

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/13651603107

sjpb:
> Is this still an issue with the new k3s boot system?

Hmm, no I don't think so. The dropins happen in `boostrap.yml` (systemd group) so before k3s now. This is gonna be easy to miss when we support it in compute-init though ...

sjpb:
I updated this from main, which also bumped the image, so it will need a rebuild. Don't bother for now though, I'm rebuilding on top of this.

PR #451, opened 2024-10-15
Add workflow to cleanup CI clusters

sjpb:
CI clusters are deleted if the OpenTofu provisioning fails or the workflow is cancelled, but are left up if the Ansible fails to allow debugging.

This PR adds a workflow nightly to clean up any remaining CI clusters, unless the control node has been tagged with "keep", e.g. using:

    openstack server set slurmci-RL9-1374-control --tag keep --os-compute-api-version 2.26

It also amends the deployment/reimage workflow so that it is not triggered by changes to workflows, other than itself.


.github/workflows/nightly-cleanup.yml
```diff
@@ -0,0 +1,71 @@
...
+            - LEAFCLOUD
+            - SMS
+            - ARCUS
+  schedule:
+    - cron: '0 20 * * *'  # Run at 8PM - image sync runs at midnight
+
+jobs:
+  ci_cleanup:
+    name: ci-cleanup
+    concurrency: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.cloud }}-${{ matrix.build }}
```
bertiethorpe, line 19:
```suggestion
    concurrency: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.cloud }}
```

.github/workflows/nightly-cleanup.yml
```diff
@@ -0,0 +1,71 @@
...
+          echo "ci_clusters=${CI_CLUSTERS}" >> GITHUB_ENV
+        shell: bash
+      
+      - name: Delete clusters if control node not tagged with keep
+        run: |
+          . venv/bin/activate
+          for cluster_prefix in ${CI_CLUSTERS}
+          do
+            TAGS=$(openstack server show ${cluster_prefix}-control --column tags --format value)
+            if [[ TAGS =~ "keep" ]]; then
```
bertiethorpe, line 65:
```suggestion
            if [[ $TAGS =~ "keep" ]]; then
```

bertiethorpe (changes requested):
Couple things, but it looks good. Definitely needed

bertiethorpe (approved):
LGTM

PR #452, opened 2024-10-15
Allow creating tofu-controlled compute volumes

sjpb:
Adds a "volumes" property to the `compute` terraform variable which defines optional OpenTofu-controlled volumes attached to compute nodes. Useful for simulating additional disks present in baremetal instances using virtualized instances.

Note that:
- Unlike the default `home` and `state` volumes` attached to the control node, compute volumes defined this way are not automatically formatted or mounted.
- Volumes are attached after instance creation and the path under `/dev/{v,s}d*` is likely to change across reboots. Use e.g. `/dev/disk/by-path/` for mounts etc.


sjpb:
> Warning: updating images for compute nodes currently forces them to be destroyed and replaced as the uuid of their block devices changes, even if no volumes are defined for them

This is fixed in final PR.

sjpb:
Closed by https://github.com/stackhpc/ansible-slurm-appliance/pull/528

PR #454, opened 2024-10-15
Ansible playbook to configure sshd for Conch CA certs.

MaxBed4d:


ansible/ca-cert.yml
```diff
@@ -0,0 +1,27 @@
+# An ansible playbook to configure the SSHD configuration to enable CA cert auth for SSH.
+# Remember to export CI_CLOUD if it isn't part of your environment's variables.
```
sd109, line 2: Why is this needed? How are you running this playbook?

ansible/ca-cert.yml
```diff
@@ -0,0 +1,27 @@
...
+
+# NOTE: Change the src for the `ssh_signing_key.pub` to be your corresponding directory.
+
+- hosts: login
+  gather_facts: true
+  become: true
+  tasks:
+    - name: Copy ssh public key
+      ansible.builtin.copy:
+        src: /var/lib/rocky/conch/ssh_signing_key.pub
```
sjpb, line 12: How's this getting configured? You just created this by hand right?

ansible/ca-cert.yml
```diff
@@ -0,0 +1,27 @@
+# An ansible playbook to configure the SSHD configuration to enable CA cert auth for SSH.
+# Remember to export CI_CLOUD if it isn't part of your environment's variables.
+
+# NOTE: Change the src for the `ssh_signing_key.pub` to be your corresponding directory.
+
+- hosts: login
+  gather_facts: true
```
sjpb, line 7: Doesn't look like you need facts, so I'd turn it off or at least leave it out (which will default to on, but at least you're not documenting that you need something you don't).

ansible/ca-cert.yml
sjpb, line 2: CI_CLOUD is only required if you're using app creds. Playbooks aren't the place to document the wriggles of authing to openstack TBH.

sjpb: Yeah we need to talk about where this is included in site.yml, there's a bigger job/discussion here than just adding a playbook I think.

ansible/ca-cert.yml
```diff
@@ -0,0 +1,27 @@
...
+      ansible.builtin.copy:
+        src: /var/lib/rocky/conch/ssh_signing_key.pub
+        dest: /etc/ssh/ca_user_key.pub
+        owner: root
+        group: root
+        mode: '0644'
+        remote_src: true
+
+    - name: Ensure CA Certs are accepted
+      ansible.builtin.lineinfile:
```
sjpb, line 20: Does this need a `regexp:` line to make it idempotent? E.g. if its already there, but commented out?

ansible/ca-cert.yml
```diff
@@ -0,0 +1,27 @@
...
+        line: 'TrustedUserCAKeys /etc/ssh/ca_user_key.pub'
+        dest: /etc/ssh/sshd_config
+
+    - name: Restart SSH service
+      ansible.builtin.systemd:
+        name: sshd
+        state: restarted
```
sjpb, line 27: Depending on how this should actually be integrated, this might need gating somehow

ansible/ca-cert.yml
```diff
@@ -0,0 +1,27 @@
+# An ansible playbook to configure the SSHD configuration to enable CA cert auth for SSH.
+# Remember to export CI_CLOUD if it isn't part of your environment's variables.
+
+# NOTE: Change the src for the `ssh_signing_key.pub` to be your corresponding directory.
+
+- hosts: login
```
sjpb, line 6: Normally we'd have a group for the feature, but I'm not clear on which feature this actually belongs to. I suggested login to @MaxBed4d understanding this was a one-off play running against an already-deployed cluster. But that wouldn't be a PR ...

sd109:
As discussed elsewhere, let's just merge this as-is for now since it's targeting the Waldur proof of concept branch. We can worry about polishing things later if needed.

PR #457, opened 2024-10-18
Bump terraform collection to fix race with waiting for ssh

sjpb:
See https://github.com/azimuth-cloud/ansible-collection-terraform/pull/16

### DIFF
--- a/ansible/roles/cluster_infra/tasks/main.yml
+++ b/ansible/roles/cluster_infra/tasks/main.yml
@@ -10,20 +10,20 @@
 # outputs.cluster_gateway_ip.
 - block:
     - name: Look up floating IP
-      include_role:
-        name: stackhpc.terraform.infra
-        tasks_from: lookup_floating_ip
-      vars:
-        os_floating_ip_id: "{{ cluster_floating_ip }}"
+      azimuth_cloud.terraform.os_floating_ip_info:
+         floating_ip: "{{ cluster_floating_ip }}"
+      register: cluster_floating_ip_info
 
     - name: Set floating IP address fact
       set_fact:
-        cluster_floating_ip_address: "{{ os_floating_ip_info.floating_ip_address }}"
-  when: cluster_floating_ip is defined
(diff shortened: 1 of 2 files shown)


requirements.yml
```diff
@@ -45,7 +45,7 @@ collections:
     version: 2.1.0
   - name: https://github.com/stackhpc/ansible-collection-terraform
```
mkjpryor, line 46: This needs to change to `azimuth-cloud`, as do any uses of the roles from the collection.

mkjpryor: I'm actually surprised it works as it is? Probably the CI doesn't test the CaaS Terraform?

mkjpryor (changes requested):

requirements.yml
sjpb: Yeah CI doesn't test this, I just ran into this while testing manually in a dev azimuth :flushed:

sjpb:
Tested in azimuth using:
```
azimuth_caas_stackhpc_slurm_appliance_git_version: update/tf-collection
azimuth_caas_stackhpc_slurm_appliance_image: 5e9af448-ab2b-4d48-bb63-9ab103a85e80 #openhpc-RL9-241009-1523-354b048a # current main @ c7f3b24
```

worked fine.

ansible/roles/cluster_infra/tasks/main.yml
```diff
@@ -10,20 +10,20 @@
...
+      register: cluster_floating_ip_info
 
     - name: Set floating IP address fact
       set_fact:
-        cluster_floating_ip_address: "{{ os_floating_ip_info.floating_ip_address }}"
-  when: cluster_floating_ip is defined
+        cluster_floating_ip_address: "{{ cluster_floating_ip_info.floating_ip.floating_ip_address }}"
+      when:
+        - cluster_floating_ip is defined
+        - cluster_floating_ip
```
mkjpryor, line 22: This `when` should be on the `block` I think

mkjpryor (changes requested):

mkjpryor (approved):
LGTM

PR #466, opened 2024-11-06
Add blackbox exporter to k3s monitoring stack with probe for OOD

wtripp180901:
Installs blackbox exporter into k3s cluster and adds Prometheus scrape job with probe for OOD. Also adds blackbox alerting rules from https://github.com/azimuth-cloud/capi-helm-charts/blob/2355ba6151289f35bba9a9e8bd7c372930c323a4/charts/cluster-addons/templates/monitoring/blackbox-exporter.yaml#L98

NB: probes are slower than typical probes in capi helm charts due to indirection from accessing OOD through headless service in k3s cluster


wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11702928693

PR #471, opened 2024-11-08
Pinned nvidia driver version to fix nightly cuda builds

wtripp180901:
Temporarily pins nvidia driver module to the 560-open stream to fix fix incompatibility between cuda packages latest (12.6.2) and nvidia-driver latest (565.57.01). Override in packer extra_vars should be deleted when cuda packages compatible with nvidia driver > 561

NB: the cuda_driver_stream rolevar previously seemed to refer to the module profile rather than the stream and was unused in the role anyway


wtripp180901:
nightly build completes but unable to run cuda tests on images because device files for nvidia device files don't seem to get picked up on SMS

sjpb:
Actual problem looked like:

```
Depsolve Error occurred: 
 Problem: package cuda-12-6-12.6.3-1.x86_64 from cuda-rhel9-x86_64 requires cuda-runtime-12-6 >= 12.6.3, but none of the providers can be installed
...
```

sjpb:
Replacing with https://github.com/stackhpc/ansible-slurm-appliance/pull/496

PR #474, opened 2024-11-13
Unset signature_verified property from nightly/latest images

sjpb:
#467 accidently removed the step which unsets the `signature_verified` property on nightly builds. This property existing means volume creation [fails](https://bugs.launchpad.net/cinder/+bug/1823445) for fatimage builds using the nightly images.


bertiethorpe (approved):
LGTM

PR #476, opened 2024-11-13
Ansible-init compute node script

bertiethorpe:


ansible/roles/etc_hosts/tasks/main.yml
```diff
@@ -6,3 +6,26 @@
     group: root
     mode: 0644
   become: yes
+
+- name: Ensure /exports/hosts directory exists and copy /etc/hosts
```
sjpb, line 10: This should be part of compute_init/tasks/main too, I think?

environments/common/inventory/group_vars/all/nfs.yml
```diff
@@ -15,3 +15,16 @@ nfs_configurations:
     nfs_server: "{{ nfs_server_default }}"
     nfs_export: "/exports/home" # assumes skeleton TF is being used
     nfs_client_mnt_point: "/home"
+
+  - comment: Export /etc/hosts copy from Slurm control node
```
sjpb, line 19: Why don't we just have a single directory, /exports/cluster, with the hosts file within that.

ansible/roles/compute_init/files/compute-init.yml
```diff
@@ -0,0 +1,333 @@
...
+          register: _copy_nm_config
+
+        - name: Reload NetworkManager
+          ansible.builtin.systemd:
+            name: NetworkManager
+            state: reloaded
+          when: _copy_nm_config.changed | default(false)
+
+
+    - name: Mount /etc/hosts on compute nodes
```
sjpb, line 93: I think this should be /mnt/cluster, then you just copy stuff out of it

ansible/roles/compute_init/files/compute-init.yml
```diff
@@ -0,0 +1,333 @@
...
+            src: /mnt/hosts/hosts
+            dest: /etc/hosts
+            owner: root
+            group: root
+            mode: 0644
+
+
+    - name: NFS client mount
+      block:
+        - name: ensure mount directory exists
```
sjpb, line 120: This looks like it only supports a single mount point?

ansible/roles/compute_init/files/compute-init.yml
```diff
@@ -0,0 +1,333 @@
...
+            state: directory
+
+        - name: mount the filesystem
+          mount:
+            path: "{{ nfs_client_mnt_point }}"
+            src: "{{ nfs_server }}:{{ nfs_export }}"
+            fstype: nfs
+            state: "{{ nfs_client_mnt_state }}"
+        
+        - name: Ensure the mount directory exists
```
sjpb, line 132: This is part of the linux FHS, so will always exist

ansible/roles/compute_init/files/compute-init.yml
```diff
@@ -0,0 +1,333 @@
...
+          file:
+            path: /mnt/
+            state: directory
+            mode: 0755
+        
+        - name: Mount /mnt/
```
sjpb, line 138: Nah something is confused here

ansible/filesystems.yml
```diff
@@ -24,3 +24,22 @@
...
+            state: directory
+            owner: root
+            group: root
+            mode: 0755
+
+        - name: Copy manila share info to /exports/cluster
+          copy:
+            content: "{{ os_manila_mount_share_info | to_nice_yaml }}"
+            dest: "/exports/cluster/manila_share_info.yml"
+      when: os_manila_mount_share_info is defined
```
sjpb, line 45: This should be sufficent to allow you to dev just running the compute_init role

bertiethorpe:
Tested by running (from deploy host)
`ansible-playbook -v ansible/adhoc/rebuild.yml --limit compute`
`ansible-playbook -v ansible/extras.yml --tags compute_init`

Then in the compute node:
`sudo rm /var/lib/ansible-init.done`
`sudo systemctl restart ansible-init`
`sudo scontrol reconfigure`

Check status of nodes:
`scontrol show nodes`

Run tests (from deploy host):
`ansible-playbook ansible/adhoc/hpctests.yml -v`

ansible/roles/compute_init/files/compute-init.yml
sjpb, line 1: General thoughts on this:
1. Once this is built into the fat image, it will need to be gated via metadata to only run on compute nodes (b/c it will be on all). We can't use `k3s_server` (or whatever you rename that to) b/c that is also on login nodes. Ideally, we'd also do this in a way which allows specialisation *within* the compute nodes.
2. Once this is built into the fact image, it needs to not run on the initial boot of compute nodes, b/c normal ansible won't have configured the control node. So maybe it needs to exit if the nfs server is not available?
3. We need to think about how this interacts with
  a. other ansible-init playbooks (remote collections, or local) - I think this should come after, on the basis that site.yml is gated to run after all ansible-init playbooks and this is basically replacing that. If you explore from e.g. https://github.com/azimuth-cloud/azimuth-images/blob/main/ansible/roles/linux-monitoring/tasks/main.yml#L49, maybe those are using numbers 0-99 and we should use 100+ for this? I don't know. Will need to check how things sort too.
  b. hooks - how can we allow this to be extended?

sjpb (changes requested):

ansible/roles/compute_init/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+
+- name: Ensure directories exist
+  file: 
+    path: "/etc/ansible-init/{{ item.directory }}"
+    state: directory
+    owner: root
+    group: root
+    mode: 0755
+  loop:
+    - { directory: "templates" }
```
sjpb, line 11: Why is this not just:
```
loop:
- templates
- files
...
```

?

ansible/roles/compute_init/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
+---
+
+- name: Ensure directories exist
```
sjpb, line 3: So we also need to inject our group vars. Except actually group vars don't really exist once in a play. So we could create hostvars like `.../inventory/$host.yml`, but then file actually has to be [host-specific](https://docs.ansible.com/ansible/latest/inventory_guide/intro_inventory.html#assigning-a-variable-to-one-machine-host-variables):

```
all:
  {{ inventory_hostname }}:
     {{ hostvars[hostname] | sort_out_indentation }}
```
and we have to somehow make "localhost" *that* host.

Maybe there's a better way ....

ansible/roles/compute_init/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+  copy:
+    src: '{{ item }}'
+    dest: '/etc/ansible-init/files/{{ item | basename }}'
+    owner: root
+    group: root
+    mode: 0644
+  loop:
+    - ../../resolv_conf/files/NetworkManager-dns-none.conf
+
+- name: Inject files
```
sjpb, line 38:
```suggestion
- name: Inject libraries
```

sjpb, line 38: Seeing this does make me wonder if we could just inject files from e.g. `roles/myrole/tasks/runtime.yml` ...

bertiethorpe: Why leave compute init functionality inside the roles though? I think this way is more modular and inline with the purpose of inventory groups. Plus keeping everything centralised aids readability for compute init. (If I'm understanding you)

ansible/roles/compute_init/tasks/main.yml
sjpb: Actually its not that bad. The only host in a compute_init play is localhost, so we can create group_vars/all for each host. E.g we template out `..$host/inventory/group_vars/all.yml` but its is literally just `{{ hostvars[hostname] }}`.

The other alternative is just that that templated out file is read using `include_vars:`.

I'm not sure which we should choose; what we *do* need is that we preserve the "normal ansible" behaviour where group_vars override role defaults; if the compute_init "role default" equivalents are in `vars:` blocks the nfs-exported vars must override those. See the ansible variable precedence docs ...

ansible/roles/compute_init/files/compute-init.yml
```diff
@@ -0,0 +1,271 @@
...
+          service:
+            name: munge
+            enabled: true
+            state: started
+
+        - name: Ensure slurmd state
+          service:
+            name: slurmd
+            enabled: true
+            state: started
```
sjpb, line 271: Need to ensure the node is resumed, because slurmctld will think this has "unexpectedly rebooted"

something like:
```
- name: Ensure node is in cluster
  command: scontrol update state=resume nodename=<hostname>
```

I'm not sure how you get hostname best - normally it'd be `{{ inventory_hostname }}` but that won't work here. It needs to be the short name. Maybe using `={{ lookup('pipe', 'hostname -s') }}` is not mad?

sjpb (changes requested):

ansible/roles/compute_init/tasks/main.yml
```diff
@@ -0,0 +1,150 @@
...
+
+    # - name: Template hostvars
+    #   template:
+    #     src: ../templates/hostvars.j2
+    #     dest: "/exports/cluster/{{ inventory_hostname }}/hostvars.yml"
+    #     owner: root
+    #     group: root
+    #     mode: 0644
+
+  delegate_to: "{{ groups['control'] | first }}"
```
sjpb, line 142: I think the logic here needs breaking up:

1. Stuff which only actually needs control host (e.g. coping /etc/hosts into nfs exports dir): Yes, having the play hosts as `compute_init` and delegating to first control make sense so the `compute_init` group becomes a functionality toggle. But it needs a `run_once: true` else e.g. with 100 hosts you're coping out those files 100 times over each other, and probably at some point that goes wrong.

2. Stuff which is actually host-specific, e.g. templating hostvars into the nfs export, you also want to run on the `compute_init` group but delegated to first control is right, but it needs to run on every host.

ansible/roles/compute_init/tasks/main.yml
```diff
@@ -0,0 +1,150 @@
+---
+
+- name: Ensure directories exist
```
sjpb, line 3: This is running on group `compute_init` which via `everything` layout defaults to `cluster`.

So you're creating the directories etc. on EVERY node. Whereas we only want to do that on the control node.

environments/common/layouts/everything
```diff
@@ -82,6 +82,10 @@ openhpc
 # Hosts to run ansible-init
 cluster
 
+[compute_init:children]
+# Hosts to deploy compute initialisation ansible-init script to.
```
sjpb, line 86: I don't think this description is right.

Deploying the actual compute init script is/will be done in the image build.

This should control which hosts get info templated out (and eventually, metadata set to turn on the feature, I think).

environments/common/layouts/everything
```diff
@@ -82,6 +82,10 @@ openhpc
...
+cluster
```
sjpb, line 87: I don't think this is right, it needs to either be `builder` or `compute`. I need to discuss.

environments/common/inventory/group_vars/all/nfs.yml
```diff
@@ -15,3 +15,9 @@ nfs_configurations:
     nfs_server: "{{ nfs_server_default }}"
     nfs_export: "/exports/home" # assumes skeleton TF is being used
     nfs_client_mnt_point: "/home"
+
+  - comment: Export /exports/cluster from Slurm control node
+    nfs_enable:
+        server: "{{ inventory_hostname in groups['control'] }}"
+        clients: false
+    nfs_export: "/exports/cluster" # control node has to copy in /etc/hosts to here
```
sjpb, line 23:
```suggestion
    nfs_export: "/exports/cluster"
```

environments/common/inventory/groups
```diff
@@ -136,6 +136,9 @@ freeipa_client
 [ansible_init]
 # Hosts to run linux-anisble-init
 
+[compute_init]
```
sjpb, line 139: See comments on everything.

sjpb (changes requested):

PR #483, opened 2024-11-20
Fix adhoc/rebuild wait_for_connection race condition

bertiethorpe:
When rebuilding nodes, a race condition can result in ssh connection being re-established before rebuild has begun.

Delay of 60s fixes this.


wtripp180901 (approved):
LGTM

PR #484, opened 2024-11-20
Pin DNF releasever to RL9.4 to fix RL9.5 failures

bertiethorpe:
Since rockylinux 9.5 got released 19 Nov 2024, we can’t currently can’t build updated mages for slurm at all, due to:

- podman v5 not being compatible with ansible podman module.

This temporary fix pins RL 9.4 so we have working image builds

### DIFF
--- a/.github/workflows/fatimage.yml
+++ b/.github/workflows/fatimage.yml
@@ -33,10 +33,10 @@ jobs:
       SOURCE_IMAGES_MAP: |
         {
           "RL8": {
-            "openstack.openhpc": "rocky-latest-RL8"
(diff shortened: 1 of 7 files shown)


bertiethorpe:
rocky-latest-test builds:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11938849874

bertiethorpe:
fatimage build:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/11941897359

sjpb:
Proven not to work, there is no way to set 9.4 for rocky repos and 9 for epel

PR #485, opened 2024-11-20
Support mapping compute + login instances to Ironic nodes

sjpb:
Add support to non-caas environments for mapping baremetal compute and/or login nodes to ironic hosts with the same names, e.g:

```
# environments/$env/terraform/terraform.tfvars:
...
cluster_name = "mycloud"
login_nodes = {
    login-0 = {
        flavor: "baremetal.highmem"
        match_ironic_node: true
    }
}
compute = {
    hpc = {
        flavor = "baremetal.xlarge"
        match_ironic_node: true
        nodes = ["hpc-0", "hpc-1", "hpc-2"]
    }
}
...
```

will result in an instance named `mycloud-login-0` on ironic node `login-0`, and compute instances `mycloud-hpc-N` on ironic nodes `hpc-N`.

**NB:** To support this the definition of login nodes has changed slightly.

Also fixes a reversion where with the tf variable `volume_backed_instances` set, compute nodes were not actually volume-backed.

Note this PR only affects new environments, unless changes from skeleton terraform are copied over to site-specific environments.

Tests:
- [x] Works for non-mapped nodes on a cloud where user does not have support for listing baremetal nodes
- [ ] Works for mapping nodes on a cloud where user does have support for listing baremetal nodes
- [x] Can create an entire volume-backed cluster


environments/skeleton/{{cookiecutter.environment}}/terraform/compute/nodes.tf
```diff
@@ -1,3 +1,11 @@
+data "external" "nodes" {
+  program = ["bash", "-c", <<-EOT
+    openstack baremetal node list --limit 0 -f json 2>/dev/null | \
+    jq -r 'try map( { (.Name|tostring): .UUID } ) | add catch {}' || echo '{}'
```
jovial, line 4: Is this so it behaves well when Ironic is not enabled?

sjpb: Yes, or e.g. you're not admin/don't have policy amendments and deploying a VM cluster hence can't/don't need to query BM nodes at all

sjpb: Writing that out made me think though - it'd be better if it worked so IF `match_ironic_node` is true, this command has to work, and if you don't this command is actually a no-op.

sjpb: actually we can't really gate the command like that, given `match_ironic_node` could be true on some nodes and not others.

sjpb: Turns out setting a default of `nova` for the AZ doesn't work, b/c if you don't have admin/right policy, you can't set this at all...

sjpb:
Updated in #573

PR #489, opened 2024-12-06
Fix issues when performing OFED/CUDA/Lustre extra builds

wtripp180901:
- Pinned NVIDIA driver stream
- Removed Lustre cleanup task which deletes mellanox packages


sjpb:
Replaced by another PR

PR #496, opened 2024-12-13
Pin nvidia-driver and cuda packages to working packages

sjpb:
- Pins nvidia-driver and cuda packages to avoid a dnf depsolve error on install
- Makes installing cuda packages optional
- Only run nvidia-driver/cuda installs during image build, but can be done from an "extra" build on top of an existing fat image
- Will error if the install changes the kernel, as this is assumed to break things
- Extends CI extra build test to also include cuda role

Depsolve error with `565-open` nvidia-driver stream was:
```
Depsolve Error occurred: 
 Problem: package cuda-12-6-12.6.3-1.x86_64 from cuda-rhel9-x86_64 requires cuda-runtime-12-6 >= 12.6.3, but none of the providers can be installed
...
```

Replaces https://github.com/stackhpc/ansible-slurm-appliance/pull/471/files


ansible/roles/cuda/defaults/main.yml
```diff
@@ -1,7 +1,6 @@
-cuda_distro: "rhel{{ ansible_distribution_major_version }}"
```
wtripp180901, line 1: why was this removed despite its templated value being duplicated elsewhere? Maybe private var instead if we don't want it to be overridden

ansible/roles/cuda/defaults/main.yml
```diff
@@ -1,7 +1,6 @@
...
-cuda_repo: "https://developer.download.nvidia.com/compute/cuda/repos/{{ cuda_distro }}/x86_64/cuda-{{ cuda_distro }}.repo"
-cuda_driver_stream: default
-cuda_package_version: 'latest'
+cuda_repo_url: "https://developer.download.nvidia.com/compute/cuda/repos/rhel{{ ansible_distribution_major_version }}/{{ ansible_architecture }}/cuda-rhel{{ ansible_distribution_major_version }}.repo"
+cuda_driver_stream: '560-open' # 565-open has problems with cuda packages
```
wtripp180901, line 2: nitpick but `cuda_nvidia_driver_stream` might be a better name to make it clearer that people shouldn't be looking for cuda-specific drivers

wtripp180901 (changes requested):
some nitpicks but otherwise LGTM

wtripp180901:
have we been able to test cuda on any GPU nodes yet btw?

sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/12397634280

wtripp180901 (approved):
LGTM assuming CI passes

PR #505, opened 2024-12-18
Added CI to create PRs to bump to latest Ark timestamps

wtripp180901:
Added a nightly workflow which will check the URLs defined in `appliances_repo_timestamp_sources` for newer timestamps, update `appliances_repo_timestamps` and create a PR if any timestamps have been updated


sjpb (changes requested):
See discussion on slack re. simplifying things so we're not defining repo URLs in 3x places.

Also needs to
a) kick off an image build in the PR - and not run the stackhpc.yml workflow
- Could do this by changing that workflow to not run on branch names like e.g. auto/ and setting that branch for the PR
b) make a commit to the PR to change environments/.stackhpc/terraform/cluster* vars to the new image - then run the stackhpc workflow


OR

- do the image build on the branch
- update ../cluster* definition
- THEN open the PR (which runs stackhpc.yml)

wtripp180901:
todo: test fatimage workflow dispatch still works (will affect the auto/bump-timestamps branch too if it doesn't)

wtripp180901:
looks like input.event_name is inherited from caller and is running image build/ci on this branch so needs reworking

wtripp180901:
Todo:
- [ ] Check fatimage dispatch outside of update-timestamps still works as intended
- [ ] Check CI still runs on correct branch outside of update-timestamps
Todo after review but before merging:
- [ ] Change reference for is_callee action to main (will break the workflow here but will work once merged)
- [ ] Remove on.pull_request from update_timestamps workflow

wtripp180901:
successful e2e test https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/12670021951

PR #473, opened 2024-11-11
Change defaults so a cookiecutter environment is fully functional

wtripp180901:
- [x] **NB: Changes default `selinux_state` to `disabled` to allow prometheus to work**
- [x] Enables `etc_hosts` by default to provide working internal name resolution
- [x] Adds working defaults for OpenOndemand (likely to need changing for production):
  - Use basic_auth + PAM for authentication
  - Use `ansible_host` (i.e. IP address) as the ondemand external "servername"
- [x] Creates a default user to access ondemand with
- [x] Default cluster suffix is now "internal" instead of "invalid" now the former is reserved by ICANN for private-use applications.
- [x] Grafana anonymous auth is now enabled by default for new environments

PR #508, opened 2024-12-19
Add more information re. configuring production sites

sjpb:


docs/production.md
```diff
@@ -1,9 +1,146 @@
...
+  generally be moved into the `site` environment. It should be be encrypted
+  using [Ansible vault](https://docs.ansible.com/ansible/latest/user_guide/vault.html)
+  and then committed to the repository.
+
+- Ensure created instances have accurate/synchronised time. For VM instances
+  this is usually provided by the hypervisor, but if not (or for bare metal
+  instances) it may be necessary to configure or proxy `chronyd` via an
+  environment hook.
+
+- The cookiecutter provided tofu configurations define resources for home and
```
sd109, line 79: Nit: We should probably use OpenTofu everywhere instead of just tofu
```suggestion
- The cookiecutter provided OpenTofu configurations define resources for home and
```

docs/production.md
```diff
@@ -1,9 +1,146 @@
...
+  state volumes. The former may not be required if the cluster's `/home` is
+  provided from an external filesystem (or Manila). In any case, in at least
+  the production environment, and probably also in the staging environment,
+  the volumes should be manually created and the resources changed to data
+  resources. This ensures that even if the cluster is deleted via tofu, the
```
sd109, line 84:
```suggestion
  the volumes should be manually created and the resources changed to [data
  resources](https://opentofu.org/docs/language/data-sources/). This ensures that even if the cluster is deleted via tofu, the
```

docs/production.md
```diff
@@ -1,9 +1,146 @@
...
+        fixed_ip {
+            subnet_id = data.openstack_networking_subnet_v2.cluster_subnet.id
+            ip_address = var.control_ip_address
+        }
+    }
+    ```
+    
+  Note the variable `control_ip_address` is new.
+
+  Using fixed IPs will require either using admin credentials or policy changes.
```
sd109, line 116: In Azimuth we handle this by saying you have to pre-allocate the FIP to the project manually (i.e. `openstack floating ip create <network>`) and then put the allocated FIP into the config. Maybe we could do the same here and make `control_ip_address` a data resource instead?

sd109 (dismissed):
LGTM, just a couple of minor suggested improvements.

docs/production.md
sjpb: So there's two different things here; my "Note ... " change is talking about the NON-floating IPs - we don't want control IPs on an external network.

Your comment is valid actually - we don't discuss FIPs at all, and generally yes, for login nodes they should be provisioned outside of TF as you say and attached via TF (which we don't currently have a code for in the skeleton, but we do at every site; I think skeleton TF changes are out of scope here). I will do a docs change at least ...

docs/production.md
```diff
@@ -1,9 +1,146 @@
...
+          same is used for staging and production) should be set as *defaults*
+          in `environments/site/terraform/variables.tf`, and then don't need to
+          be passed in to the module.
+
+- Vault-encrypt secrets. Running the `generate-passwords.yml` playbook creates
+  a secrets file at `environments/$ENV/inventory/group_vars/all/secrets.yml`.
+  To ensure staging environments are a good model for production this should
+  generally be moved into the `site` environment. It should be be encrypted
+  using [Ansible vault](https://docs.ansible.com/ansible/latest/user_guide/vault.html)
+  and then committed to the repository.
```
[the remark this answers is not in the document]
sjpb: Yeah. I'm not sure TBH, prevents some failure modes ...

docs/production.md
sjpb: Actually - login FIPs are too hard to explain in docs, we should actually fix this in https://github.com/stackhpc/ansible-slurm-appliance/pull/473

docs/production.md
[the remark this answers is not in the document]
sjpb: Think we're going to keep same secrets for the moment.

sd109 (approved):

PR #518, opened 2025-01-06
Support and test "re-imageable" compute nodes via compute node metadata

bertiethorpe:
Support and test "re-imageable" compute nodes:
- [x] Add functionality to skeleton TF to allow defining enable_ flags in compute node metadata, as a list of roles/functionalities to enable.
- [x] Set the enable_* metadata in stackhpc environment only to turn on compute, etc_hosts, nfs, basic_users, eessi. This is all currently supported functionality except for resolv_conf (not used in stackhpc), manila (not available at either leafcloud or sms)
- [x] Add steps to stackhpc workflow, after sucessful reimage of control/login nodes, to reimage compute nodes (with same image) via ansible adhoc, and to check that nodes come back successfully.
- [x] Document the supported metadata flags
- [x] Document the above reimaging workflow


.github/workflows/stackhpc.yml
```diff
@@ -170,33 +170,22 @@ jobs:
...
-          ansible all -m wait_for_connection -a 'delay=60 timeout=600' # delay allows node to go down
           ansible-playbook -v ansible/site.yml
           ansible-playbook -v ansible/ci/check_slurm.yml
 
+      - name: Test reimage of compute nodes and compute-init (via rebuild adhoc)
+        run: |
+          . venv/bin/activate
+          . environments/.stackhpc/activate
+          ansible-playbook -v --limit compute ansible/adhoc/rebuild.yml
+          ansible all -m wait_for_connection -a 'delay=60 timeout=600' # delay allows node to go down
```
sjpb, line 186: Why do we need this? The rebuild adhoc already has a wait (as deleted from the other reimage above)

environments/skeleton/{{cookiecutter.environment}}/terraform/compute.tf
```diff
@@ -18,4 +18,6 @@ module "compute" {
   k3s_token = var.k3s_token
   control_address = [for n in openstack_compute_instance_v2.control["control"].network: n.fixed_ip_v4 if n.access_network][0]
   security_group_ids = [for o in data.openstack_networking_secgroup_v2.nonlogin: o.id]
+
+  compute_init_enable = var.compute_init_enable
```
sjpb, line 22: Nope, this means we can't define it per compute node group. It needs to be taken from each.value.

environments/skeleton/{{cookiecutter.environment}}/terraform/variables.tf
```diff
@@ -136,3 +137,9 @@ variable "k3s_token" {
     description = "K3s cluster authentication token, set automatically by Ansible"
     type = string
 }
+
```
sjpb, line 140: Nope, we don't want to define this globally, just needs to be defined in the `compute` variable.

environments/skeleton/{{cookiecutter.environment}}/terraform/variables.tf
```diff
@@ -52,6 +52,7 @@ variable "compute" {
             image_id: Overrides variable cluster_image_id
             vnic_type: Overrides variable vnic_type
             vnic_profile: Overrides variable vnic_profile
+            compute_init_enable: Toggles ansible-init rebuild
```
sjpb, line 55: Need to document here what values are (currently) permissible.

environments/skeleton/{{cookiecutter.environment}}/terraform/compute/nodes.tf
```diff
@@ -45,9 +45,16 @@ resource "openstack_compute_instance_v2" "compute" {
   }
 
   metadata = {
-    environment_root = var.environment_root
-    k3s_token = var.k3s_token
-    control_address = var.control_address
+    environment_root   = var.environment_root
```
sjpb, line 48: This seems pretty redundant. The idea of having `compute_init_enable` as a list matching the metadata was that it would be easy to construct this. Something like:

```
metadata = merge(
    {
        ...
     },
    {for e in var.compute_init_enable: e => true})
)
```

environments/.stackhpc/terraform/main.tf
```diff
@@ -58,6 +58,10 @@ variable "k3s_token" {
     type = string
 }
 
+variable "compute_init_enable" {
```
sjpb, line 61: Nope, not required.

environments/.stackhpc/terraform/main.tf
```diff
@@ -74,6 +78,7 @@ module "cluster" {
     cluster_image_id = data.openstack_images_image_v2.cluster.id
     control_node_flavor = var.control_node_flavor
     k3s_token = var.k3s_token
+    compute_init_enable = var.compute_init_enable
```
sjpb, line 81: Just set this directly here, there's no need for another variable/file. Although following other comments this will want to be defined it the compute= block.

environments/.stackhpc/terraform/compute_init.auto.tfvars
sjpb, line 1: Should be defined directly in main.tf, no need for  an extra var / auto vars file.

environments/skeleton/{{cookiecutter.environment}}/terraform/variables.tf
sjpb: Can just refer to the role docs probably, b/c that is what defines what works.

ansible/roles/compute_init/README.md
```diff
@@ -42,10 +42,35 @@ The following roles/groups are currently fully functional:
   node and all compute nodes.
 - `openhpc`: all functionality
 
-# Development/debugging
+All of the above are defined in the skeleton cookiecutter config, and are
```
sjpb, line 45: I think this is a bit misleading. I'd replace this entire para with something like:

> The above may be enabled by setting the `compute_init_enable` property on the terraform `compute` variable

ansible/roles/compute_init/README.md
```diff
@@ -42,10 +42,35 @@ The following roles/groups are currently fully functional:
...
+toggleable via a terraform compute_init autovar file. In the .stackhpc
+environment, the compute init roles are set by default to:
+- `enable_compute`: This encompasses the openhpc role functionality while being
+  a global toggle for the entire compute-init script.
+- `etc_hosts`
+- `nfs`
+- `basic_users`
+- `eessi`
+
+# CI workflow
```
sjpb, line 55: If useful, I think this should me moved to docs/experiemental/compute-init.md.

ansible/roles/compute_init/README.md
sjpb, line 1: Given this all works, can we delete the bulk of docs/experiemental/compute-init.md which shows results/progress? Its not very interesting now IMO.

sjpb (changes requested):
Need to make it so compute_init_enable can be set differently on different compute groups.

Doing  `git diff HEAD...main` on this branch shows the images are out of date with the current main. So it needs to be updated from main.

.github/workflows/stackhpc.yml
```diff
@@ -170,33 +170,22 @@ jobs:
...
-          ansible all -m wait_for_connection -a 'delay=60 timeout=600' # delay allows node to go down
           ansible-playbook -v ansible/site.yml
           ansible-playbook -v ansible/ci/check_slurm.yml
 
+      - name: Test reimage of compute nodes and compute-init (via rebuild adhoc)
+        run: |
+          . venv/bin/activate
+          . environments/.stackhpc/activate
+          ansible-playbook -v --limit compute ansible/adhoc/rebuild.yml
+          ansible all -m wait_for_connection -a 'delay=60 timeout=600'
```
sjpb, line 186: We don't want to wait here. Instead we want the check_slurm.yml play to retry for some specified period.

edit: And actually looking at it, it already has some retry logic. So you will need to debug why that isn't sufficent.

sjpb (changes requested):

bertiethorpe:
Fat image: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/12691968569

bertiethorpe:
fat image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/12698399333

environments/skeleton/{{cookiecutter.environment}}/terraform/compute.tf
```diff
@@ -25,4 +25,6 @@ module "compute" {
   k3s_token = var.k3s_token
   control_address = [for n in openstack_compute_instance_v2.control["control"].network: n.fixed_ip_v4 if n.access_network][0]
   security_group_ids = [for o in data.openstack_networking_secgroup_v2.nonlogin: o.id]
+
+  compute_init_enable = each.value.compute_init_enable
```
sjpb, line 29: Don't we need to default to the empty list on this? Otherwise we *always* have to supply it?

sjpb (commented):
One query

sjpb (approved):
LGTM

PR #519, opened 2025-01-06
Update ceph to use ark packages and move RL9 to ceph reef

wtripp180901:
- Makes Ark ceph packages from centos stream available for following versions (at current snapshots):
  - RL8 ceph quincy 17.2.0-1 to 17.2.7-2 inclusive
  - RL9 ceph reef 18.2.0-2 to  18.2.4-2 inclusive

- Bumps os-manila-mount role so that default versions are now:
  - RL8: unchanged - quincy 17.2.7. Note reef is not supported
  - RL9: updated - reef 18.2.4


wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/12635773595

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/12653281831

wtripp180901:
^ shouldn't have changed since [a800c9c](https://github.com/stackhpc/ansible-slurm-appliance/pull/519/commits/a800c9c3b2b6682efb907bee94d48360ca55cf07), good to merge

sjpb (commented):
I'm not sure I like this - how do we know that the centos ark repos point to the specific versions of ceph which os-manila-mount v25.1.1 uses by default?

sjpb (commented):
LGTM, I've confirmed that from ark we can get the ceph versions required by os-manila-mount for the snapshots given.

sjpb (approved):
LGTM, I've confirmed that from ark we can get the ceph versions required by os-manila-mount for the snapshots given.

PR #523, opened 2025-01-08
Define cluster images in a single variable

sjpb:
- Update `terraform/` directories to `tofu/` to avoid confusion.
- Change skeleton terraform so that cluster images can be defined in a single variable (which can be supplied via a single auto-loaded variables file). This will support future changes for slurm-controlled rebuild.

Note the changes to the stackhpc terraform are mostly because we need to provide image *names* as the input - because the same image has different IDs across our CI clouds.

- [ ] Implement changes
- [ ] Update docs/production.md once https://github.com/stackhpc/ansible-slurm-appliance/pull/508 merged


sjpb:
Closing, don't think this is the right way to go about it.

PR #525, opened 2025-01-08
Generate and persist hostkeys for ondemand and login nodes

wtripp180901:
A set of hostkeys are now generated idempotently in the state directory and copied to all hosts in the `persist_hostkeys` group. Login and OOD groups are now in `persist_hostkeys` by default


wtripp180901:
tested reimaging login node and then reconfiguring cluster on local azimuth deployment

ansible/roles/persist_hostkeys/tasks/main.yml
```diff
@@ -1,33 +1,46 @@
...
-    dest: /etc/ssh/
-    remote_src: true
+  - name: Check for existing hostkeys
+    find:
+      paths: "{{ persist_hostkeys_state_dir }}/"
+    register: _files_found
+
+  - name: Generate hostkeys
+    when: _files_found.matched == 0
+    shell:
```
sjpb, line 21:
```suggestion
    shell:
    # ssh-keygen -A needs a directory with an /etc/ssh suffix to write hostkeys into
```

ansible/roles/persist_hostkeys/tasks/main.yml
```diff
@@ -1,33 +1,46 @@
...
+      cmd: |
+        mkdir -p {{ persist_hostkeys_state_dir }}/etc/ssh
+        ssh-keygen -A -N \"\" -f {{ persist_hostkeys_state_dir }}
```
sjpb, line 24:
```suggestion
        ssh-keygen -A -N '' -f {{ persist_hostkeys_state_dir }}
```
maybe, escaping shouldn't be required, and "" is hard to understand

sjpb (changes requested):
Looks good, minor suggestion only.

sjpb (approved):
LGTM

PR #526, opened 2025-01-08
Fix epel not using Ark repos for RL8

wtripp180901:
Was previously being overridden back to upstream by OHPC in .stackhpc environment


wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/12675380376

sjpb (approved):
LGTM

PR #527, opened 2025-01-09
Fix volume_backed_instances not working for compute nodes

sjpb:
Relevant variables were not passed in to compute module.

Tested:
- setting volume_backed_instances and root_volume_size just for one compute group
- setting volume_backed_instances at top-level and non-default root_volume_size for one compute group

Note CI does not test this so can be merged without that passing.


wtripp180901 (approved):
LGTM

wtripp180901 (approved):
LGTM

PR #528, opened 2025-01-09
Support additional volumes on compute nodes

sjpb:
Adds an `extra_volumes` property to the tofu `compute` mapping to allow defining volumes which OpenTofu will create and attach to compute nodes. Useful for simulating additional disks present in baremetal instances using virtualized instances.

Note that:

- Unlike the default home and state volumes attached to the control node, compute volumes defined this way are not automatically formatted or mounted.
- Volumes are attached after instance creation and the path under /dev/{v,s}d* may not reflect the ordering in the `extra_volumes` property and may change across reboots. Use e.g. /dev/disk/by-path/ for mounts etc.
- As long as volume-backed compute instances are not used, OpenTofu will update instances in place (i.e. reimage) to change the node image, even with `extra_volumes` specified.


sjpb:
Replaces https://github.com/stackhpc/ansible-slurm-appliance/pull/452 which undesirably deleted the ability to create volume-backed compute instances.

Tested that:
- With 2x volumes of different sizes (to allow distinguishing), volumes are correctly created, attached, and visible via `lsblk`.
- When changing the cluster_image, the opentofu plan showed "update in place" for these compute instances.

wtripp180901 (approved):
LGTM

PR #532, opened 2025-01-13
Fix nightly cleanup to deal with duplicate server names

bertiethorpe:
```openstack server show ${cluster_prefix}-control --column tags --format value```
is returning multiple servers when querying for control node tag. Fix improves logic for duplicate named clusters

### DIFF
--- a/.github/workflows/nightly-cleanup.yml
+++ b/.github/workflows/nightly-cleanup.yml
@@ -63,11 +63,20 @@ jobs:
             echo "No clusters to delete."
             exit 0
           fi
-           
+          
           for cluster_prefix in ${ci_clusters}
           do
             echo "Processing cluster: $cluster_prefix"
-            TAGS=$(openstack server show ${cluster_prefix}-control --column tags --format value)
+            # Get all servers with the matching name for control node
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: .github/workflows/nightly-cleanup.yml [hunk 1]
```diff
@@ -67,7 +67,16 @@ jobs:
...
-            TAGS=$(openstack server show ${cluster_prefix}-control --column tags --format value)
+            # Get all servers with the matching name for control node
+            CONTROL_SERVERS=$(openstack server list --name ${cluster_prefix}-control --format json)
+            SERVER_COUNT=$(echo "$CONTROL_SERVERS" | jq length)
+            
+            if [[ $SERVER_COUNT -gt 1 ]]; then
+              echo "Warning: More than one server found for control node '${cluster_prefix}-control'."
+              continue
+            fi
+            TAGS=$(echo "$CONTROL_SERVERS" | jq -r '.[0].Tags' )
```
sd109, line 78: Does this assume that all servers are tagged the same as the first item in the list?

bertiethorpe: ah yes good spot

sd109 (approved):
LGTM

PR #538, opened 2025-01-15
K3s no longer runs if token isn't provided

wtripp180901:
ansible-init tasks to configure and enable k3s and k3s-agent services are now skipped if no token is provided as Openstack metadata


sd109:
@wtripp180901 Can we close this PR now if we've decided to move away from the k3s via ansible-init approach?

wtripp180901:
> @wtripp180901 Can we close this PR now if we've decided to move away from the k3s via ansible-init approach?

yeah suppose this isn’t needed anymore

wtripp180901:
> @wtripp180901 Can we close this PR now if we've decided to move away from the k3s via ansible-init approach?

yeah suppose this isn’t needed anymore

PR #539, opened 2025-01-16
Support defining default routes (including dummy ones) for compute nodes via cloud-init

sjpb:
All nodes need a default route so k3s can identify the primary IP. This can be hard to do via subet gateways for clusters with some multi-homed nodes. This PR adds tofu compute group parameters `gateway_nmcli_connection` and `gateway_ip` which allow defining default routes for compute nodes via cloud-init to work around this.


sjpb:
Replaced by https://github.com/stackhpc/ansible-slurm-appliance/pull/602

PR #540, opened 2025-01-16
Read k3s_token from secrets.yml file

sjpb:
Allows this to be kept vault-encrypted rather than as a unencrypted terraform variable.

### DIFF
--- a/environments/.stackhpc/terraform/main.tf
+++ b/environments/.stackhpc/terraform/main.tf
@@ -69,8 +69,8 @@ module "cluster" {
     key_pair = "slurm-app-ci"
     cluster_image_id = data.openstack_images_image_v2.cluster.id
     control_node_flavor = var.control_node_flavor
-    # need to override defaults as using terraform/ as a module but secrets templated
-    # out to *this* environment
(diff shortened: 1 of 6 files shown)


environments/.stackhpc/terraform/main.tf
```diff
@@ -73,7 +69,9 @@ module "cluster" {
     key_pair = "slurm-app-ci"
     cluster_image_id = data.openstack_images_image_v2.cluster.id
     control_node_flavor = var.control_node_flavor
-    k3s_token = var.k3s_token
+    # need to override defaults as using terraform/ as a module but secrets templated
+    # out to *this* environment
```
wtripp180901, line 73: in that case, maybe inventory_secrets_path shouldn't have a default and just error if not overridden since this feels like it could bite us at some point

wtripp180901 (changes requested):
one possible suggestion

environments/.stackhpc/terraform/main.tf
sjpb: Tried to clarify

wtripp180901 (approved):
LGTM

PR #542, opened 2025-01-21
EXPERIMENTAL: test lifecycle ignore image changes via variable

sjpb:
When a compute group has  compute_init_enable set/non-empty:
- Using the ansible variable image_id templated out using the compute group tf var of the same name, into the TF-templated ansible inventory for that compute group.

  -  This means that ansible will have image_id for each compute node, and the compute_init:export.yml task will write that out into the /exports/cluster/hostvars.

 - Add a lifecycle block to the openstack_compute_instance_v2 resource so that subsequent changes to the tf cluster_image variable for that compute group don’t actually result in a change via tf plan/apply. This is done with the meta-argument "ignore_changes"
 
   - In practice this isn't possible due to meta-arguments (e.g. lifecycle and its contents) only taking static values, so duplication of the openstack_compute_instance_v2 is necessary with some gated condition on which is used.

### DIFF
--- a/environments/.stackhpc/tofu/main.tf
+++ b/environments/.stackhpc/tofu/main.tf
@@ -81,6 +81,7 @@ module "cluster" {
             nodes: ["compute-0", "compute-1"]
             flavor: var.other_node_flavor
             compute_init_enable: ["compute", "etc_hosts", "nfs", "basic_users", "eessi"]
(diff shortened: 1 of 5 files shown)


bertiethorpe (approved):
LGTM

sjpb:
Only done to demo a workaround.

PR #547, opened 2025-01-22
Define login nodes using an opentofu module

sjpb:
Changes the definition of login nodes in the opentofu variables to match the approach used for compute nodes.

This simplifies the opentofu configurations, allows more flexibility in login node definition, and will allow adding multiple networks in subsequent work.


README.md
```diff
@@ -89,8 +89,11 @@ Create an OpenTofu variables file to define the required infrastructure, e.g.:
     cluster_subnet = "some_subnet" # *
     key_pair = "my_key" # *
     control_node_flavor = "some_flavor_name"
-    login_nodes = {
-        login-0: "login_flavor_name"
+    login = {
+        interactive = {
```
m-bull, line 93: Is `interactive` significant? Would you ever need more than one group here?

If you'd only ever have one group, then perhaps this could be handled down in the module rather than the slightly complicated data structure here. If you could have more than one group, maybe a few lines of docs or a comment to explain the data structure would be helpful?

environments/skeleton/{{cookiecutter.environment}}/tofu/variables.tf
```diff
@@ -29,9 +29,27 @@ variable "control_node_flavor" {
     description = "Flavor name for control name"
 }
 
-variable "login_nodes" {
-  type = map
-  description = "Mapping defining login nodes: key -> (str) nodename suffix, value -> (str) flavor name"
+variable "login" {
+  type = any
+  description = <<-EOF
+    Mapping defining groups of login nodes. Keys are names of groups.
```
m-bull, line 35: Ah, this is where the docs are :facepalm: 

Maybe the readme could describe _why_ you'd want multiple login groups?

m-bull (changes requested):

environments/skeleton/{{cookiecutter.environment}}/tofu/variables.tf
sjpb: You might. One client has real login nodes, plus an ondemand node which is a different size. I agree its a bit nasty though.

README.md
m-bull, line 93:
```suggestion
        # Arbitrary group name for these login nodes
        interactive = {
```

README.md
```diff
@@ -89,8 +89,11 @@ Create an OpenTofu variables file to define the required infrastructure, e.g.:
...
+            nodes: ["login-0"]
+            flavor: "login_flavor_name"
```
m-bull, line 95:
```suggestion
            flavor: "login_flavor_name" # *
```

environments/skeleton/{{cookiecutter.environment}}/tofu/variables.tf
m-bull: That makes complete sense, now you mention it.

Do the group names for the compute and login groups get exposed back up to ansible?

README.md
m-bull: I think `compute_flavor_name` below needs `# *` too?

environments/skeleton/{{cookiecutter.environment}}/tofu/variables.tf
sjpb: Yeah. The compute groups here get turned into "{{ cluster_name }}_{{group_name}}" groups which are what is used for "stackhpc.openhpc" partition definition.

m-bull (approved):
LGTM - I think the docs changes are helpful!

PR #548, opened 2025-01-22
Support multiple networks in OpenTofu configurations

sjpb:
Replaces opentofu variables `cluster_net` and `cluster_subnet` with new `cluster_networks` taking a list of mappings. How vnic properties are defined has changed to suit. Additional networks can be attached to compute and login node groups via an `extra_networks` parameter. Note this means at least the first network in `cluster_networks` will span all nodes - this is a requirement.

Also adds `docs/networks.md` describing supported configurations.

### DIFF
--- a/environments/skeleton/{{cookiecutter.environment}}/tofu/control.tf
+++ b/environments/skeleton/{{cookiecutter.environment}}/tofu/control.tf
@@ -53,7 +53,7 @@ resource "openstack_compute_instance_v2" "control" {
     for_each = {for net in var.cluster_networks: net.network => net}
     content {
       port = openstack_networking_port_v2.control[network.key].id
-      access_network = length(var.cluster_networks) == 1 ? true : lookup(network.value, "access_network", false)
+      access_network = network.key == var.cluster_networks[0].network
     }
   }
 


docs/networks.md
```diff
@@ -0,0 +1,104 @@
...
+  }
+]
+...
+```
+
+## Multiple homogenous networks
+This is similar to the above, except each node has multiple networks. Therefore
+`access_network` must be explicitly set. Note that only one subnet must have
+a gateway defined, else default routes via both subnets will be present causing
+routing problems. It also shows the second network as using direct-type vNICs
```
m-bull, line 46:
```suggestion
routing problems. It also shows the second network (netB) using direct-type vNICs
```

docs/networks.md
```diff
@@ -0,0 +1,104 @@
...
+]
+
+compute = {
+  baremetal = {
+    nodes = ["baremetal-0", "baremetal-1"]
+    networks = [
+      {
+        network = "netA"
+        subnet = "subnetA"
+        access_network = true
```
m-bull, line 90: I think being able to override this inside the compute object might be a bit of a footgun - setting it here could create heterogeneity across compute groups. 

IMHO it should be a network-level property that describes whether a network is connected to all hosts.

docs/networks.md
```diff
@@ -0,0 +1,104 @@
...
+```terraform
+cluster_networks = [
+  {
+    network = "netA"
+    subnet = "subnetA"
+  }
+]
+
+compute = {
+  baremetal = {
```
m-bull, line 84:
```suggestion
  general = {
    nodes = ["general-0", "general-1"]
  }
  baremetal = {
```

docs/networks.md
```diff
@@ -0,0 +1,104 @@
...
+
+This example shows how to override variables for specific node groups. In this
+case a baremetal node group has a second network attached. As above, only a
+single subnet can have a gateway IP.
+
+```terraform
+cluster_networks = [
+  {
+    network = "netA"
+    subnet = "subnetA"
```
m-bull, line 79:
```suggestion
    subnet = "subnetA"
    access_network = true
```

m-bull (changes requested):
All seems pretty good. I've some general docs comments, but they might be due to my misunderstanding of how things are supposed to work...

docs/networks.md
sjpb: As in we have "access_network = $network_name" at the top level? Yeah, that force the correct behaviour wouldn't it. We can still default it to the first network so you don't always have to supply it.

docs/networks.md
sjpb: Nope, not needed, given this is the only network on default nodes.

docs/networks.md
```diff
@@ -0,0 +1,102 @@
+# Networking
+
+The default OpenTofu configurations in the appliance do not provision networks,
+subnets or associated infrastructure such as routers. The requirements are that:
+1. At least one network exists.
+2. The first network defined spans all nodes, referred to as the "access network".
+3. Only one subnet per network is attached to nodes.
+4. At least one network on each node provides outbound internet access (either
```
m-bull, line 8: Could we tighten this further to say that the access network *must* provide internet connectivity (as some nodes could only be attached to this network) directly/via proxy, although "extra" networks may or may not provide access to the internet?

docs/networks.md
m-bull, line 1: This readme is hugely clearer to me now - nice!

m-bull (changes requested):
One final nit to possibly tighten the language around the access network and internet access, but otherwise this looks great now.

docs/networks.md
sjpb: I think the problem is that which network provides internet might vary across the cluster (have a client in mind!). Although making that work properly for other issues probably also requires https://github.com/stackhpc/ansible-slurm-appliance/pull/539. At this level, I don't think the nodes actually care which interface provides that.

m-bull (approved):

Issue #543, opened 2025-01-21
k3s node IPs not necessarily correct for multi-homed hosts


sjpb:
As per the k3s docs, a default route is required and is used to determine the primary IP for a cluster. However in some cases this means the node IPs are not all on the same network, e.g.

```
          NetA   NetB
           |      |
login -----x------x
control----x------x
compute1----------x
```

where NetA has a default route, and NetB doesn't, with a dummy route set on `compute1` via cloud-init using https://github.com/stackhpc/ansible-slurm-appliance/pull/539.

In this case the k3s server on the control node gets an InternalIP on NetA, while compute1's is on NetB.

The IP the _nodes_ should use for the k3s server is set by [templating](https://github.com/stackhpc/ansible-slurm-appliance/blob/d317920bce540690a6922a429b7a1a88f78afed5/ansible/roles/k3s/files/start_k3s.yml#L25) out [K3S_URL](https://docs.k3s.io/cli/agent?_highlight=k3s_url#cluster-options) at boot via ansible-init. However it turns out this is not sufficient in the above case and e.g. shelling into a container running on `compute` from k9s on the control node does not work.

In manual testing, setting `--node-ip` (available for both  [server](https://docs.k3s.io/cli/server) and [agent](https://docs.k3s.io/cli/agent) sub--commands) got this working.

Although those links show it isn't "natively" exposed as an environment variable, [INSTALL_K3S_EXEC](https://docs.k3s.io/installation/configuration#configuration-with-install-script) could be set to something like `--node-ip $K3S_NODE_IP` and then template out "K3S_NODE_IP=$ip" into an environment file via ansible-init (with the environment file reference possibly added by a dropin, configured to not start until that exists).

The [docs suggest](https://docs.k3s.io/installation/airgap#default-network-route) that even if setting `--node-ip` a default route is still required (emphasis added):
> K3s requires a default route in order to auto-detect the node's primary IP, _and for kube-proxy ClusterIP routing to function properly_

sjpb:
Actually for the k3s-server, maybe we should make it so `--node-ip` is *always* set to the same IP as  K3S_URL? The assumption there is that the latter is an IP, not a hostname, but given our dns state that is probably OK.

sjpb:
@wtripp180901 FYI - please comment if the above doesn't make sense (you don't need to do anything yet though, this just capturing thoughts)

sjpb:
[...]

PR #550, opened 2025-01-23
Set K3s node IPs to use access_network

sjpb:
Sets the IP address used for k3s to the address on the "access network" (first cluster network). This ensures that k3s nodes can all reach each other.

Fixes https://github.com/stackhpc/ansible-slurm-appliance/issues/543.


sjpb:
Rebased here: https://github.com/stackhpc/ansible-slurm-appliance/pull/556

PR #553, opened 2025-01-29
Support compute node rebuild/reboot via Slurm RebootProgram

bertiethorpe:
Adds rebuild role to the appliance, which uses https://github.com/stackhpc/slurm-openstack-tools//slurm_openstack_tools/reboot.py, a RebootProgram running on the control node which:

- Reads each of the compute nodes' hostvars.yml file to get the “target” image defined.
- Use openstack API to check the current image of the node.
- If they do not match - reimage the node with the target image via openstack.
- If no target image is defined (i.e. this node does have this functionality enabled) or if it matches the current image, a normal reboot (NB: not rebuild) will be carried out via openstack.

Invoked via slurm by running e.g.

```
# srun --reboot -N 2 uptime
```


ansible/slurm.yml
```diff
@@ -9,6 +9,15 @@
     - include_role:
         name:  mysql
 
+- name: Setup slurm-driven rebuild
+  hosts: rebuild
```
sjpb, line 13:
```suggestion
  hosts: rebuild:!builder
```

sjpb: If this is turned on during build, we don't want the clouds.yml ending up in the image

.github/workflows/stackhpc.yml
```diff
@@ -178,12 +178,13 @@ jobs:
           ansible-playbook -v ansible/site.yml
           ansible-playbook -v ansible/ci/check_slurm.yml
 
-      - name: Test reimage of compute nodes and compute-init (via rebuild adhoc)
+      - name: Test compute node reimage, compute-init, and reboot
```
sjpb, line 181:
```suggestion
      - name: Test compute node reboot and compute-init
```
Currently doesn't test an actual rebuild

ansible/adhoc/reboot_via_slurm.yml
```diff
@@ -0,0 +1,23 @@
+# Reboot compute nodes via slurm. The nodes will be rebuilt if image in hostvars is different to the active one in OpenStack.
```
sjpb, line 1:
```suggestion
# Reboot compute nodes via slurm. Nodes will be rebuilt if `image_id` in inventory is different to the currently-provisioned image.
```

ansible/adhoc/reboot_via_slurm.yml
```diff
@@ -0,0 +1,23 @@
...
+# Example:
+#   ansible-playbook -v ansible/adhoc/reboot_via_slurm.yml
+
+- hosts: login
```
sjpb, line 5: Can we set `run_once: true` on these so we can use on production clouds?

ansible/adhoc/reboot_via_slurm.yml
```diff
@@ -0,0 +1,23 @@
...
+  tasks:
+    - name: Submit a Slurm job to reboot compute nodes
+      ansible.builtin.shell: |
+        set -e
+        srun --reboot -N 2 uptime
+      become_user: root
+      register: slurm_result
+      failed_when: slurm_result.rc != 0
+
+    - name: Fetch Slurm logs if reboot fails
```
sjpb, line 17:
```suggestion
    - name: Fetch Slurm controller logs if reboot fails
```

nice touch

ansible/roles/rebuild/defaults/main.yml
sjpb, line 1: Can you add a README for this role please?

environments/common/inventory/group_vars/all/openhpc.yml
```diff
@@ -35,8 +35,14 @@ openhpc_config_default:
   SlurmctldParameters:
     - enable_configless
   TaskPlugin: task/cgroup,task/affinity
+  ReturnToService: 2
```
sjpb, line 38:
```suggestion
  ReturnToService: 2 # is stackhpc.openhpc default, but templating bug means it is needed here too
```

environments/common/inventory/groups
```diff
@@ -165,3 +162,6 @@ extra_packages
 
 [pulp]
 # Add builder to this group to enable automatically syncing of pulp during image build
+
```
sjpb, line 165: Can you move this up to the same place as before please, avoids merge noise/conflicts

environments/common/layouts/everything
```diff
@@ -111,3 +109,5 @@ control
 [extra_packages:children]
 # Hosts to install specified additional packages on
 builder
+
+# NB: [rebuild] not defined here as likely to need features not currently supported
```
sjpb, line 113: Move up to same place as before please

sjpb (changes requested):
Some minor tweaks. I had a comment re. defaults or something but TBH I can't remember what that was - it was for a file not changed in this PR.

sjpb (approved):
LGTM

PR #555, opened 2025-01-29
avoid tf updates to login/compute on control delete/recreate

sjpb:
Uses the control node port to derive the control node IP instead of the control node's access_ip property. This avoids opentofu recalculating the compute/login metadata when it deletes/recreate the control node. Works because the first cluster_network is always the access network.


m-bull (approved):
LGTM

PR #556, opened 2025-01-30
Set k3s node IP from access network IP

sjpb:
The node IP for k3s server and agent nodes is set to be the hosts's IP on the access network (first cluster network). This ensures nodes can communicate with each other properly.


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/13051327970

m-bull (approved):
LGTM

sjpb:
Tested on a dual-homed cluster. Nodes get correct IP.

PR #558, opened 2025-02-04
Stop CaaS Slurm OOD using TLS

bertiethorpe:


sjpb:
Overriding `openondemand_osc_ood_defaults` is not the way to go, because that also breaks all the other things which are needed, e.g. `clusters`. The osc.ood role variables should simply be set directly.

PR #560, opened 2025-02-05
docs: update README to use new network syntax

priteau:
This follows changes made in https://github.com/stackhpc/ansible-slurm-appliance/pull/548.


sjpb (approved):
Thanks, nice catch

PR #563, opened 2025-02-10
Add lustre_repo variable

jovial:
This allows users to configure the git repository used to build lustre. In some environments external ssh is blocked, so I've been using this https mirror:

https://github.com/lustre/lustre-release.git

### DIFF
--- a/ansible/roles/lustre/defaults/main.yml
+++ b/ansible/roles/lustre/defaults/main.yml
@@ -6,6 +6,7 @@ lustre_mount_state: mounted
 lustre_mount_options: 'defaults,_netdev,noauto,x-systemd.automount,x-systemd.requires=lnet.service'
 
 # below variables are for build and should not generally require changes
+lustre_repo: "git://git.whamcloud.com/fs/lustre-release.git"
 lustre_build_packages:
   - "kernel-devel-{{ ansible_kernel }}"
   - git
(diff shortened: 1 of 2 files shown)


ansible/roles/lustre/tasks/install.yml
```diff
@@ -6,7 +6,7 @@
 - name: Clone lustre git repo
   # https://git.whamcloud.com/?p=fs/lustre-release.git;a=summary
   ansible.builtin.git:
-    repo: git://git.whamcloud.com/fs/lustre-release.git
+    repo: "{{ lustre_git_repo }}"
```
bertiethorpe, line 9:
```suggestion
    repo: "{{ lustre_repo }}"
```
match variable name

bertiethorpe (changes requested):

ansible/roles/lustre/defaults/main.yml
```diff
@@ -6,6 +6,7 @@ lustre_mount_state: mounted
 lustre_mount_options: 'defaults,_netdev,noauto,x-systemd.automount,x-systemd.requires=lnet.service'
 
 # below variables are for build and should not generally require changes
+lustre_repo: "git://git.whamcloud.com/fs/lustre-release.git"
```
m-bull, line 9:
```suggestion
lustre_git_repo: "git://git.whamcloud.com/fs/lustre-release.git"
```

ansible/roles/lustre/tasks/install.yml
m-bull: Nit: this would fix the issue, but I think the slightly more descriptive variable name is helpful. Plus if @jovial has this config overridden somewhere then it won't have to be changed there if we change the name of the default.

jovial: Nice, spot - cheers guys

sjpb:
Lustre isn't used in CI, so no need to wait for that.

bertiethorpe (approved):
LGTM

PR #565, opened 2025-02-10
Support sshd password authentication on Rocky 8

jovial:
Rocky 8 doesn't have an sshd_config.d directory, so we need to adjust the main configuration file.

### DIFF
--- a/ansible/roles/sshd/tasks/configure.yml
+++ b/ansible/roles/sshd/tasks/configure.yml
@@ -1,5 +1,23 @@
-- name: Grab facts to determine distribution
-  setup:
+- name: Ensure drop in directory exists
+  file:
+    path: /etc/ssh/sshd_config.d/*.conf
+    state: directory
+    owner: root
+    group: root
+    mode: 700
+  become: true
+
+- name: Ensure drop in directory is included
+  lineinfile:
+    dest: /etc/ssh/sshd_config
+    regexp: "^Include /etc/ssh/sshd_config.d/*.conf"
+    line: "Include /etc/ssh/sshd_config.d/*.conf"
+    state: present
+    insertafter: EOF
+    validate: sshd -t -f %s
+  notify:
+    - Restart sshd
+  become: true
 
 - name: Template sshd configuration
   # NB: If parameters are defined multiple times the first value wins;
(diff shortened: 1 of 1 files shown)


ansible/roles/sshd/tasks/configure.yml
```diff
@@ -13,3 +16,17 @@
...
+    dest: /etc/ssh/sshd_config
+    regexp: "^PasswordAuthentication"
+    line: "PasswordAuthentication {{ 'yes' if sshd_password_authentication | bool else 'no' }}"
+    state: present
+    validate: sshd -t -f %s
+  notify:
+    - Restart sshd
+  become: true
+  when: ansible_facts.distribution_major_version == '8'
+
```
jovial, line 32:
```suggestion
```

jovial:
Try using Include

jovial:
> Try using Include

Seems to work.

jovial:
This is what we end up with:

```
[root@cclr-dev-login-0 ~]# cat /etc/ssh/sshd_config
#	$OpenBSD: sshd_config,v 1.103 2018/04/09 20:41:22 tj Exp $

# This is the sshd server system-wide configuration file.  See
# sshd_config(5) for more information.

# This sshd was compiled with PATH=/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin

# The strategy used for options in the default sshd_config shipped with
# OpenSSH is to specify options with their default value where
# possible, but leave them commented.  Uncommented options override the
# default value.
# BEGIN ANSIBLE MANAGED BLOCK
# To modify the system-wide sshd configuration, create a  *.conf  file under
#  /etc/ssh/sshd_config.d/  which will be automatically included below
Include /etc/ssh/sshd_config.d/*.conf
# END ANSIBLE MANAGED BLOCK
```
Which isn't ideal, but we don't have [append_newline and prepend_newline ](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/blockinfile_module.html#parameter-append_newline)in our version of ansible.

sjpb:
> I'm guessing it is using the first value it encounters 

From the man page:
> For each keyword, the first obtained value will be used. 
so, yes

sjpb (approved):
LGTM

PR #567, opened 2025-02-10
Fix lustre IP route detection if there is no gateway

jovial:
This handles the following case

```
172.26.0.251 dev ib0 src 172.26.0.196 uid 0
```

in addition to:

```
10.167.128.1 via 10.179.0.2 dev eth0 src 10.179.3.149 uid 1000
```

### DIFF
--- a/ansible/roles/lustre/tasks/configure.yml
+++ b/ansible/roles/lustre/tasks/configure.yml
@@ -43,4 +43,3 @@
     state: "{{ (item.mount_state | default(lustre_mount_state)) }}"
     opts: "{{ item.mount_options | default(lustre_mount_options) }}"
   loop: "{{ lustre_mounts }}"
- 
--- a/ansible/roles/lustre/tasks/configure.yml
+++ b/ansible/roles/lustre/tasks/configure.yml
@@ -12,7 +12,7 @@
     _lustre_interface: "{{ _lustre_ip_r_mgs_info.dev }}"
     _lustre_ip: "{{ _lustre_ip_r_mgs_info.prefsrc }}"
   vars:
-    _lustre_ip_r_mgs_info: "{{ _lustre_ip_r_mgs.stdout | from_json }}"
+    _lustre_ip_r_mgs_info: "{{ _lustre_ip_r_mgs.stdout | from_json | first }}"
 
 - name: Write LNet configuration file
   template:
--- a/ansible/roles/lustre/tasks/configure.yml
+++ b/ansible/roles/lustre/tasks/configure.yml
@@ -9,9 +9,11 @@
 
 - name: Set facts for Lustre interface
   set_fact:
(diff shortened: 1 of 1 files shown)


ansible/roles/lustre/tasks/configure.yml
```diff
@@ -9,9 +9,11 @@
 
 - name: Set facts for Lustre interface
   set_fact:
-    _lustre_interface: "{{ _lustre_ip_r_mgs_info[4] }}"
-    _lustre_ip: "{{ _lustre_ip_r_mgs_info[6] }}"
+    _lustre_interface: "{{ _lustre_ip_r_mgs_info[interface_index | int] }}"
+    _lustre_ip: "{{ _lustre_ip_r_mgs_info[ip_index | int] }}"
   vars:
+    interface_index: "{{ 4 if 'via' in _lustre_ip_r_mgs.stdout_lines.0 else 2 }}"
```
m-bull, line 15: Rather than relying on parsing a pretty nasty string, could we generate `_lustre_ip_r_mgs` as JSON from iproute and parse it that way? Tested quickly on RL8 and RL9 iproute packages...

```
ip --json ro get 192.168.178.1
[{"dst":"192.168.178.1","dev":"wlp165s0","prefsrc":"192.168.178.90","flags":[],"uid":0,"cache":[]}]
```

I know its a slightly bigger change but it feels a bit less brittle!

m-bull (changes requested):

ansible/roles/lustre/tasks/configure.yml
jovial: Nice suggestion :)

ansible/roles/lustre/tasks/configure.yml
```diff
@@ -44,4 +43,4 @@
     state: "{{ (item.mount_state | default(lustre_mount_state)) }}"
     opts: "{{ item.mount_options | default(lustre_mount_options) }}"
   loop: "{{ lustre_mounts }}"
- 
\ No newline at end of file
+
```
jovial, line 46:
```suggestion
```

ansible/roles/lustre/tasks/configure.yml
```diff
@@ -44,4 +43,3 @@
     state: "{{ (item.mount_state | default(lustre_mount_state)) }}"
     opts: "{{ item.mount_options | default(lustre_mount_options) }}"
   loop: "{{ lustre_mounts }}"
-
```
jovial, line 47: Removed a space at the end of the file

ansible/roles/lustre/tasks/configure.yml
m-bull: Lovely!

m-bull (approved):

PR #568, opened 2025-02-11
Support memory limits and pam no-login in compute-init

bertiethorpe:
- Set locked memory limits on user-facing nodes
- Block ssh to compute nodes for non-privileged users without running jobs


sjpb (changes requested):
Can we do this in the same order as in slurm.conf please?

sjpb (approved):
LGTM

PR #571, opened 2025-02-11
Support sssd and sshd in compute-init

bertiethorpe:


ansible/roles/sssd/tasks/configure.yml
```diff
@@ -30,5 +30,5 @@
 - name: "Ensure oddjob is started"
   service:
     name: oddjobd
-    state: "{{ sssd_enable_mkhomedir }}"
+    state: "{{ 'started' if sssd_enable_mkhomedir else 'stopped' }}"
```
jovial, line 33: Was wondering if we should use when so that we never stop it:

```

        - name: "Ensure oddjob is started"
          service:
            name: oddjobd
            state: 'started'
            enabled: "{{ sssd_enable_mkhomedir }}"
      when: sssd_enable_mkhomedir | bool
  ```
Just because it may have some other uses but we definitely want it started if creating home directories

File: ansible/roles/compute_init/files/compute-init.yml [hunk 2]
```diff
@@ -132,6 +142,93 @@
         mode: 0644
       when: enable_etc_hosts
 
+    - name: Configure sssd
```
sjpb, line 145: Looking at the role, it looks like it'd just be enough to run the configure.yml task, overriding `sssd_conf_src`, if we import the role. See comment for sshd below. However this does look ok

sjpb (changes requested):
Think logic/guarding of tasks needs revising

ansible/roles/compute_init/files/compute-init.yml
```diff
@@ -132,6 +142,21 @@
...
+      ansible.builtin.include_role:
+        name: sssd
+        tasks_from: configure.yml
+      vars:
+        sssd_conf_src: "/mnt/cluster/hostconfig/{{ ansible_hostname }}/sssd.conf"
+      when: enable_sshd
```
sjpb, line 151: :warning: Role is `sssd` !

sjpb (changes requested):
Yeah much cleaner, one logic error!

ansible/roles/compute_init/files/compute-init.yml
```diff
@@ -18,6 +20,14 @@
     # TODO: "= role defaults" - could be moved to a vars_file: on play with similar precedence effects
     resolv_conf_nameservers: []
 
+    sssd_enable_mkhomedir: false
```
sjpb, line 23: Do we need these, now the role is injected?

ansible/roles/compute_init/files/compute-init.yml
bertiethorpe: oops, nice catch

environments/.stackhpc/tofu/main.tf
```diff
@@ -80,7 +80,7 @@ module "cluster" {
         standard: { # NB: can't call this default!
             nodes: ["compute-0", "compute-1"]
             flavor: var.other_node_flavor
-            compute_init_enable: ["compute", "etc_hosts", "nfs", "basic_users", "eessi", "tuned"]
+            compute_init_enable: ["compute", "etc_hosts", "nfs", "basic_users", "eessi", "sssd", "sshd", "tuned"]
```
sjpb, line 83: I don't think you can do this here:
- `sssd` isn't turned on at all in the normal groups for stackhpc
- `sshd` is only turned on for `builder` in normal groups for stackhpc
- there's no required `environments/.stackhpcfiles/sssd.conf.j2` file

bertiethorpe: accidental commit, my bad

sjpb (approved):
LGTM

PR #575, opened 2025-02-13
Adds support for configuring chrony

jovial:
Chrony is already bundled in the generic cloud image so we just have to configure it. For this I am using an off the shelf role. I've chosen mrlesmithjr.chrony as it is also used in kayobe (so we can share development effort)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,0 +1,25 @@
+# Configuration
+
+This page provides configuration snippets for various services.
+
+## Chrony
+
+Use variables from the [mrlesmithjr.chrony](https://github.com/mrlesmithjr/ansible-chrony) role.
+
+For example in: `environments/<environment>/inventory/group_vars/all/chrony`:
+
+```
+---
+chrony_ntp_servers:
+  - server: ntp-0.example.org
+    options:
+      - option: iburst
+      - option: minpoll
+        val: 8
+  - server: ntp-1.example.org
+    options:
+      - option: iburst
+      - option: minpoll
+        val: 8
+
+```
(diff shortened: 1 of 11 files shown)


jovial:
Need to add group to layouts and reference docs

docs/configuration.md
sjpb, line 1: I think I'd rather just have a chrony.md page? Follows the other pages a bit better - everything is "configuration" really!

environments/common/inventory/groups
```diff
@@ -168,3 +168,6 @@ extra_packages
 
 [cacerts]
 # Hosts to configure CA certificates and trusts on
+
+[chrony]
+# Hosts where crony configuration is applied. See docs/configuration.md#Chrony for more details.
```
sjpb, line 173: see docs comment above

environments/common/layouts/everything
```diff
@@ -113,3 +113,6 @@ builder
 
 [cacerts]
 # Hosts to configure CA certificates and trusts on
+
+[chrony]
+# Hosts where crony configuration is applied. See docs/configuration.md#Chrony for more details.
```
sjpb, line 118: see docs comment above

sjpb (changes requested):
Minor docs location tweak

environments/common/layouts/everything
jovial: done

environments/common/inventory/groups
jovial: done

docs/configuration.md
jovial: Not a problem. I've moved this now.

sjpb (approved):
LGTM

PR #576, opened 2025-02-13
WIP: Use latest release for initial CI cluster setup

sjpb:


sjpb:
First CI run above - RL8 worked, RL9 didn't. Both compute nodes got rebooted (not rebuilt - as no image change), then they froze up. Could ping but couldn't ssh in. Rescued -1, set a root password, unrescued, then it started working (!) then OOMkilled on HPL.

Just trying the RL9 one again to see if we got unlucky with the cloud.

sjpb:
Ok it passed the 2nd time!

sjpb:
Cancelled tests after rebase, building image: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/13333602095

NB: once bumped, this should trigger a rebuild rather than reimage

bertiethorpe:
Fatimage build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14057769269

bertiethorpe:
new fatimage build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14081400474

.github/workflows/stackhpc.yml
```diff
@@ -103,31 +114,31 @@ jobs:
...
+      - name: Configure cluster at latest release
         run: |
           . venv/bin/activate
           . environments/.stackhpc/activate
           ansible all -m wait_for_connection
           ansible-playbook -v ansible/site.yml
           ansible-playbook -v ansible/ci/check_slurm.yml
 
-      - name: Run MPI-based tests
+      - name: Run MPI-based tests at latest release
```
sjpb, line 141: Do we want to limit this to something quicker, e.g. just pingpong (or actually, pingmatrix may be faster - should test locally). We just need a pre-rebuild job so we can check it still exists in the monitoring, I think.

.github/workflows/stackhpc.yml
sjpb, line 1: Should we move the OOD check to the NEW image bit?

ansible/roles/slurm_tools/defaults/main.yml
sjpb, line 1: Oof we definitely should fix this before release

sjpb (commented):
Few comments.

.github/workflows/stackhpc.yml
sjpb: If you change this - note you'll need to fix up https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/ci/check_grafana.yml#L26 too

bertiethorpe (approved):

PR #577, opened 2025-02-14
Optimize copying of directories

jovial:
The ansible copy module is pretty slow when copying directories especially if there are a lot of files. This is true when coping roles installed from galaxy. Switching to the synchronize module yields much better performance.


bertiethorpe:
Rolled into https://github.com/stackhpc/ansible-slurm-appliance/pull/611

PR #578, opened 2025-02-14
Update terraform provider openstack to v3

sd109:
AFAIK the comment regarding `openstack_compute_floatingip_associate_v2` is no longer applicable so I think we can try to upgrade? I don't think we're using any of the other removed resource types listed here so shouldn't require any other changes in the appliance https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/guides/upgrade-guide-version-3

### DIFF
--- a/ansible/roles/cluster_infra/templates/providers.tf.j2
+++ b/ansible/roles/cluster_infra/templates/providers.tf.j2
@@ -5,10 +5,7 @@ terraform {
   required_providers {
     openstack = {
       source = "terraform-provider-openstack/openstack"
-      # TODO we must upgrade to 3.0.0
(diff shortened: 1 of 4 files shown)


sjpb (changes requested):
Can you update the other hits for this too please:

```
$ grep -r required_providers environments/
environments/.stackhpc/tofu/main.tf:  required_providers {
environments/skeleton/{{cookiecutter.environment}}/tofu/main.tf:  required_providers {
environments/skeleton/{{cookiecutter.environment}}/tofu/node_group/main.tf:  required_providers {
```

sjpb:
@sd109 I'm still interested in getting this in, see comment above.

PR #580, opened 2025-02-17
Reword recommendation about image

priteau:


bertiethorpe (approved):
This makes more sense to me

PR #582, opened 2025-02-17
Fix OpenTofu execution as admin

priteau:
Fetch authentication scope to retrieve the current project ID and use it for security groups. This makes it possible to apply OpenTofu using the admin role, which can be necessary for mapping instances to specific bare metal nodes.

### DIFF
--- a/environments/skeleton/{{cookiecutter.environment}}/tofu/network.tf
+++ b/environments/skeleton/{{cookiecutter.environment}}/tofu/network.tf
@@ -13,14 +13,22 @@ data "openstack_networking_subnet_v2" "cluster_subnet" {
   name = each.value.subnet
 }
 
+data "openstack_identity_auth_scope_v3" "scope" {
+  # This is an arbitrary name which is only used as a unique identifier so an
+  # actual token isn't used as the ID.
+  name = "scope"
+}
+
 data "openstack_networking_secgroup_v2" "login" {
   for_each = toset(var.login_security_groups)
 
   name = each.key
+  tenant_id = data.openstack_identity_auth_scope_v3.scope.project_id
 }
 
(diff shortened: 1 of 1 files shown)


environments/skeleton/{{cookiecutter.environment}}/tofu/network.tf
```diff
@@ -13,14 +13,22 @@ data "openstack_networking_subnet_v2" "cluster_subnet" {
...
+  # This is an arbitrary name which is only used as a unique identifier so an
+  # actual token isn't used as the ID.
+  name = "scope"
+}
+
 data "openstack_networking_secgroup_v2" "login" {
   for_each = toset(var.login_security_groups)
 
   name = each.key
+  tenant_id = data.openstack_identity_auth_scope_v3.scope.project_id
```
JohnGarbutt, line 26: Will this not break with app creds that are project member app creds?

What about instead looking at using the baremetal script with a different OS_CLOUD variable set, that may need to be something different to the creds used for the workload deploy, as in, make a tweak here to have a variable for setting a different OS_CLOUD variable:
https://github.com/stackhpc/ansible-slurm-appliance/blob/879f62775c134f1be11af2a2fd78b9866d1267f3/environments/skeleton/%7B%7Bcookiecutter.environment%7D%7D/tofu/data.tf#L3

(I still think we need to get to the point where we don't need to use an admin role user at all, and we should probably error out when we attempt to use an admin user.).

sjpb: Seems like it doens't from CI, which definitely does not have admin creds.

JohnGarbutt (approved):
I still think we should avoid using role:admin creds for automation like this, but if this gets us out of a problem, lets add this in.

PR #585, opened 2025-02-18
Disable compute-init by default & warn of security issue

sjpb:
- Removes compute_init role from everything template, so it requires a manual opt-in
- Warns of security risk due to nfs export of secrets
- Fixes stackhpc CI (which still tests it)

### DIFF
--- a/ansible/roles/compute_init/README.md
+++ b/ansible/roles/compute_init/README.md
@@ -9,9 +9,7 @@ mount the share on a user-controlled machine by tunnelling through a login
 node. This feature should not be enabled on production clusters at this time.
 
 To enable this:
-1. Add the `compute` group (or a subset) into the `compute_init` group. This is
-   the default when using cookiecutter to create an environment, via the
-   "everything" template.
+1. Add the `compute` group (or a subset) into the `compute_init` group.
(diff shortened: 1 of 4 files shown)


sd109 (approved):

ansible/roles/compute_init/README.md
```diff
@@ -3,6 +3,11 @@
...
 a reboot without running the `ansible/site.yml` playbook.
 
+**CAUTION:** The approach used here of exporting cluster secrets over NFS
+is considered to be a security risk due to the potential for cluster users to
+mount the share on a user-controlled machine by tunnelling through a login
+node. This feature should not be enabled on production clusters at this time.
+
 To enable this:
 1. Add the `compute` group (or a subset) into the `compute_init` group. This is
    the default when using cookiecutter to create an environment, via the
```
sd109, line 13: Is this bit about the cookiecutter environment still true? Should we change it there too?

sjpb: No, the environments/common/layouts/everything change is what cookiecutter uses - fixed README

sd109 (dismissed):

PR #588, opened 2025-02-18
Support setting PYTHON_VERSION

priteau:
This allows us to run the setup-env.sh script on unsupported platforms, for example macOS, with:

    PYTHON_VERSION=/opt/homebrew/bin/python3.9 ./dev/setup-env.sh


sd109 (approved):
LGTM

PR #589, opened 2025-02-19
Use bootstrap tokens provisioned by ansible for K3s instead of persistent tokens in cloud-init metadata

wtripp180901:
Server now generates a short lived bootstrap token which nodes use for initial connection/auth. Performed in site.yml rather than ansible-init. Tokens no longer needs to be generated by ansible, but consistent node passwords are needed (previously we were using the ansible-generated token for this) so added to the secret store.

**NB:** When upgrading previous clusters, the groups`k3s_server` and  `k3s_agent` must now be defined - see `environments/common/layouts/everything`.

**NB:** Given k3s is not currently required, it could be disabled for production clusters (even when using a stackhpc image) by removing any hosts from the `k3s_server` and `k3s_agent` groups.


ansible/roles/k3s/tasks/runtime.yml
```diff
@@ -0,0 +1,64 @@
+---
+- name: Check if k3s agents are already connected
+  service_facts:
+  register: services_state
+
+- name: Initialise and authenticate k3s server and agents
+  vars:
+    k3s_server_name: "{{ hostvars[groups['k3s_server'].0].ansible_host }}"
+    access_ip: "{{ ansible_default_ipv4.address }}"
```
wtripp180901, line 9: @sjpb will this work for sites without default routes?

ansible/roles/k3s/tasks/runtime.yml
```diff
@@ -0,0 +1,64 @@
+---
+- name: Check if k3s agents are already connected
+  service_facts:
+  register: services_state
+
+- name: Initialise and authenticate k3s server and agents
+  vars:
+    k3s_server_name: "{{ hostvars[groups['k3s_server'].0].ansible_host }}"
+    access_ip: "{{ ansible_host }}"
```
wtripp180901, line 9: @sjpb will this work for sites without default routes?

ansible/roles/k3s/tasks/runtime.yml
```diff
@@ -0,0 +1,64 @@
...
+    block:
+    - name: Template k3s env file
+      ansible.builtin.template:
+        dest: /etc/systemd/system/k3s.service.env
+        src: k3s.service.env.j2
+
+    - name: Start k3s server
+      ansible.builtin.systemd:
+        name: k3s
+        daemon_reload: true
```
sjpb, line 27: Maybe conditionalise the daemon reload on whether the env file has changed, to avoid doing that when we don't need it

sjpb: (maybe check the docs to see if this is ignored if state and enable conditions being as-desired mean its ignored anyway

ansible/roles/k3s/tasks/runtime.yml
```diff
@@ -0,0 +1,64 @@
+---
```
sjpb, line 1: So, I would like to split this to a) make support compute-init easier in the future and b) make it a little easier to follow.

I suggest having two task files.
- The server one is basically as-is, except you'll be able to un-nest blocks and have 2x conditions on the when (its a server, and all the agents aren't running). It should set a variable `k3s_bootstrap_token`.
- The agent one. The role has an empty default var `k3s_bootstrap_token`. When the agent playbook is run after the server one, the latter populates this. I think an agent playbook doesn't need the complex `services_states` var - this can simply be a condition on whether, for the current agent node, the service is running or not (i.e. each agent play doesn't need to know state of other agent plays).

Whether you have a "configure.yml" playbook which calls both, and is called from bootstrap.yml, or whether bootstrap.yml has 2x tasks, :shrug:.

ansible/roles/k3s/tasks/runtime.yml
```diff
@@ -0,0 +1,64 @@
...
+        dest: /etc/rancher/node/password
+        content: "{{ vault_k3s_node_password }}"
+        owner: root
+        group: root
+        mode: 640 # normal k3s install is 644 but that doesn't feel right
+
+    - name: Start k3s agent
+      ansible.builtin.systemd:
+        name: k3s-agent
+        daemon_reload: true
```
sjpb, line 61: Same comment as for server

ansible/roles/k3s/templates/k3s-agent.service.env.j2
```diff
@@ -0,0 +1,3 @@
+K3S_NODE_IP={{ access_ip }}
+K3S_TOKEN={{ hostvars[groups['control'] | first]._token_output.stdout }}
```
sjpb, line 2: Use an (empty) role default

ansible/roles/k3s/tasks/runtime.yml
```diff
@@ -0,0 +1,64 @@
+---
+- name: Check if k3s agents are already connected
+  service_facts:
+  register: services_state
```
sjpb, line 4: Can we
- prefix stuff with role name to avoid clashes
- given its only intended to be used internally I'd prefix with `_
- un-confuse it from the very similarly named services_states`

So maybe, `_k3s_services_facts`? I try to include a hint re. the task type (or, the command when using shell/command modules) when registering outputs.

ansible/roles/k3s/tasks/runtime.yml
```diff
@@ -0,0 +1,64 @@
...
+
+- name: Initialise and authenticate k3s server and agents
+  vars:
+    k3s_server_name: "{{ hostvars[groups['k3s_server'].0].ansible_host }}"
+    access_ip: "{{ ansible_host }}"
+    services_states: > # getting list of all unique agent service states
```
sjpb, line 10: Similar comments on namoing conventions again, and make it clear its agent, so maybe `_k3s_agent_states`?

ansible/roles/k3s/defaults/main.yml
```diff
@@ -3,3 +3,4 @@ k3s_version: "v1.31.0+k3s1"
 k3s_selinux_release: v1.6.latest.1
 k3s_selinux_rpm_version: 1.6-1
 k3s_helm_version: v3.11.0
+k3s_bootstrap_token_expiry: 20m
```
sjpb, line 6: Does this even need to be so long: At the point we generate the token we're immediately going to use it, so I'd have thought 10 mins is fine - might depend on how hard recovering a broken cluster is though?

ansible/roles/k3s/tasks/runtime.yml
```diff
@@ -0,0 +1,64 @@
...
+      groups['k3s_agent']
+      | map('extract', hostvars, ['services', 'k3s-agent.service', 'state'])
+      | unique
+  when: not (services_state | length == 1 and services_state[0] == 'running') 
+  block:
+  - name: Initialise server and generate bootstrap tokens
+    when: inventory_hostname in groups['k3s_server']
```
sjpb, line 17: I'm not a massive fan of leaking group names into roles tbh, but it is the simpliest approach here (and acceptable given the role is "in" the appliance)

sjpb: To be clear - no change reqd.

environments/skeleton/{{cookiecutter.environment}}/tofu/variables.tf
sjpb, line 1: Can't `inventory_secrets` in environments/skeleton/{{cookiecutter.environment}}/tofu/data.tf go too? And the python script it calls? If we're not OK pushing secrets thro TF then we shouldn't need it! Although I guess it could be useful for reading non-secrets. Maybe leave the script there but lets delete the inventory_secrets resource.

sjpb (changes requested):

sjpb:
@sd109 
> Do we have a plan yet for how to get nodes to rejoin the k3s cluster after a compute-init driven rebuild? Is it even possible with this bootstrap token approach?

Yeah I think so, we discussed it the other day. The RebootProgram will have to generate a new token, that will be distributed by :magic_wand: (i.e. however we distribute other secrets), then the node picks it up on boot.

A failure to provision a new token which results in RebootProgram failing and node being marked down (unless we do something to handle that) so I think that will work.

ansible/roles/k3s/tasks/runtime.yml
wtripp180901: Now that this is a restart conditional on there being environment changes we probably always want to reload

ansible/roles/k3s/tasks/runtime.yml
```diff
@@ -0,0 +1,64 @@
...
+
+- name: Initialise and authenticate k3s server and agents
+  vars:
+    k3s_server_name: "{{ hostvars[groups['k3s_server'].0].ansible_host }}"
+    access_ip: "{{ ansible_host }}"
+    services_states: > # getting list of all unique agent service states
+      groups['k3s_agent']
+      | map('extract', hostvars, ['services', 'k3s-agent.service', 'state'])
+      | unique
+  when: not (services_state | length == 1 and services_state[0] == 'running')
```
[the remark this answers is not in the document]
wtripp180901: turns out that never worked in the first place, replaced with k3s native connection checking

wtripp180901:
https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/13453455728

ansible/bootstrap.yml
```diff
@@ -317,6 +317,8 @@
   become: yes
   tags: k3s
   tasks:
-    - ansible.builtin.include_role:
+    - name: Install k3s
+      when: "'builder' in group_names"
```
sjpb, line 321: Could do via [host patterns](https://docs.ansible.com/ansible/latest/inventory_guide/intro_patterns.html) at L316

```ansible
- hosts: k3s:&builder
```

but :shrug: this is clear at least. No change please.

ansible/extras.yml
```diff
@@ -1,3 +1,25 @@
...
+        tasks_from: server-runtime.yml
+
+- hosts: k3s_agent
+  become: yes
+  tags: k3s
+  tasks:
+    - name: Start k3s agents
+      when: "'builder' not in group_names"
+      vars: # set outside of role to allow compute init to define own value
+        k3s_bootstrap_token: "{{ hostvars[groups['k3s_server'] | first]._k3s_token_output.stdout | default('') }}"
```
sjpb, line 18: No. Using task vars is bad, b/c a) its yet another precedence level to think about and b) you can't override them in the appliance.

Can we not explicitly set_fact for the k3s_bootstrap_token on these hosts?

ansible/bootstrap.yml
sjpb: Actually changed my mind - I think the agent one should be done via host patterns so I think this should too.

ansible/roles/k3s/tasks/server-runtime.yml
```diff
@@ -0,0 +1,31 @@
...
+
+- name: Start k3s server
+  ansible.builtin.systemd:
+    name: k3s
+    daemon_reload: "{{ _k3s_env_file_status.changed }}"
+    state: started
+    enabled: true
+
+# Possible race here as there is a delay between agents disconnecting and being registered as down, probably won't be hit in general use though
+- name: Check if k3s agents are connected
```
sjpb, line 17: I don't get the logic here: we run a task which never fails, which takes up to 50s, and continues when we find at least one node is ready?? But the next steps generates the bootstrap token which the nodes need to connect? I've obviosuly misunderstood this, maybe you can add a comment or something to explain what's happening!

sjpb (changes requested):

ansible/roles/k3s/tasks/server-runtime.yml
wtripp180901: the ignore_errors got left in there by accident, but the logic was:
1. Get a list of nodes that are up -> this fails if the control node hasn't registered itself yet, hence the retries (sort of noted on L22). The control node doesn't need a token to connect to itself
2. If some but not all nodes are up (this could be just the control node i.e no agents have connected with a token yet), the next task generates a new token

ansible/extras.yml
sjpb: I think what I want to have happen is:

1. The `server-runtime.yml` play gets an extra task:
```
- set_fact:
     k3s_bootstrap_token: ._k3s_token_output.stdout
```

2. That's set as a default via the environment:
```environments/common/group_vars/all/k3s.yml:
k3s_bootstrap_token: "{{ hostvars[groups['k3s_server'] | first].k3s_bootstrap_token | default('') }}"
```

or maybe, default None is safer than accidently providing an empty token somehow??

I guess I'm not clear if tokens can *ever* apply to the k3s-server too?

sjpb: Really, I'd like to move this all back to bootstrap.yml too (late on), i.e. after that we have a "working cluster". To me extras.yml should probably be for optional extras. And monitoring.yml is going to require k3s. So the other choice is to move it there, but I think that's not where we want it long-term.

sjpb (changes requested):
Still want to change token definition.

ansible/roles/k3s/tasks/agent-runtime.yml
```diff
@@ -0,0 +1,32 @@
+---
+
+- name: Template k3s agent env file
```
sjpb, line 3: We should set owner/group/mode restrictively here so normal users cannot access the token

ansible/roles/k3s/defaults/main.yml
```diff
@@ -3,3 +3,6 @@ k3s_version: "v1.31.0+k3s1"
 k3s_selinux_release: v1.6.latest.1
 k3s_selinux_rpm_version: 1.6-1
 k3s_helm_version: v3.11.0
+k3s_bootstrap_token: "{{ None }}" # ansible managed
```
sjpb, line 6:
```suggestion
k3s_bootstrap_token: '' # matches common environment default
```

ansible/roles/k3s/tasks/agent-runtime.yml
```diff
@@ -0,0 +1,32 @@
...
+  when: k3s_bootstrap_token != ""
```
sjpb, line 4:
```suggestion
  when: k3s_bootstrap_token != ''
```

sjpb (commented):
Tiiny comments, one on security

sjpb (approved):
LGTM

PR #590, opened 2025-02-20
Fix basic_users not modifying default nfs-shared home correctly

sjpb:
Fixes an issue where basic_users doesn't always modify /home correctly (i.e. create user home dirs and create/accept ssh keys) for the default NFS configuration, because the control node does not mount /home and might end up being the first host in the play and therefore running these tasks.

This has never been noticed because
a) this aspect of the demo_user isn't tested in CI
b) the caas environment does mount /home on the control node (to provide access to that to allow modification of packages etc across the cluster)

Also changes the default type of the created in-cluster ssh key from rsa to ed25519.

### DIFF
--- a/ansible/roles/basic_users/README.md
+++ b/ansible/roles/basic_users/README.md
@@ -2,16 +2,19 @@
 basic_users
 ===========
 
-Setup users on cluster nodes using `/etc/passwd` and manipulating `$HOME`, i.e. without requiring LDAP etc. Features:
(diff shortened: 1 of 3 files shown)


MaxBed4d (approved):
Looks to me like the changes should have the intended effect. LGTM

JohnGarbutt (approved):
I think this looks good.

I am guessing we were quite dependent on the control host not being the first host in the inventory before?

PR #591, opened 2025-02-20
Lower hpl memory fraction to reduce stress from defaults

sjpb:


JohnGarbutt (approved):
looks good.

jovial (approved):
Seems to implement what we have discussed internally.  Will we still have long runtimes on nodes with a lot of memory? Would we be better off capping the max ram to some low value?  E.g

```
min(30% ram, 32GB)
```

sjpb:
I don't think its worth it TBH.

PR #595, opened 2025-02-28
Prevent nfs being mounted by tunnelling/forwarding through login node

sjpb:
- Disables ssh forwarding and sets `secure` on nfs exports to prevent mounting shares by tunnelling/forwarding through login node.
- Also removes /exports/cluster mount from default NFS configuration for safety.

### DIFF
--- a/ansible/roles/sshd/README.md
+++ b/ansible/roles/sshd/README.md
@@ -5,5 +5,6 @@ Configure sshd.
 ## Role variables
 
 - `sshd_password_authentication`: Optional bool. Whether to enable password login. Default `false`.
+- `sshd_disable_forwarding`: Optional bool. Whether to disable all forwarding features (X11, ssh-agent, TCP and StreamLocal). Default `true`.
 - `sshd_conf_src`: Optional string. Path to sshd configuration template. Default is in-role template.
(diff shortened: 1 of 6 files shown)


environments/common/inventory/group_vars/all/nfs.yml
```diff
@@ -16,8 +16,6 @@ nfs_configurations:
     nfs_export: "/exports/home" # assumes skeleton TF is being used
     nfs_client_mnt_point: "/home"
 
-  - comment: Export /exports/cluster from Slurm control node
-    nfs_enable:
-        server: "{{ inventory_hostname in groups['control'] }}"
-        clients: false
-    nfs_export: "/exports/cluster"
+# Set 'secure' to prevent tunneling nfs mounts
+# Cannot set 'root_squash' due to home directory creation
```
jovial, line 20: Can we create the home directories on the node exporting the nfs share?

sjpb: Yes but it was agreed to do that as a a separate PR, it is quite a complex change c.f. this one.

m-bull (approved):
Looks good to plug a security gap, but we should probably revisit allowing ssh-agent and x11 forwarding in the future as these are pretty handy!

PR #596, opened 2025-02-28
Rename compute-init playbook to allow numerical sort

m-bull:
Change the numerical prefix of the compute-init playbook, so that it can be lexicographically sorted.

### DIFF
--- a/ansible/roles/compute_init/tasks/install.yml
+++ b/ansible/roles/compute_init/tasks/install.yml
@@ -58,7 +58,7 @@
 - name: Add compute initialisation playbook
   copy:
     src: compute-init.yml
-    dest: /etc/ansible-init/playbooks/1-compute-init.yml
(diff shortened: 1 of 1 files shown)


sjpb:
We totally should do this, once https://github.com/stackhpc/ansible-slurm-appliance/pull/581 has been merged

m-bull:
Caution that you might end up with two playbooks if you run `site.yml` on a not-fresh host.

sjpb:
> Caution that you might end up with two playbooks if you run site.yml on a not-fresh host.

Oof good catch!

sjpb:
> Caution that you might end up with two playbooks if you run site.yml on a not-fresh host.

Actually, no. site.yml doesn't run compute_init/tasks/install.yml, only fatimage.yml does. Which is a bit non-standard but avoids this problem.

PR #597, opened 2025-02-28
Fixed bootstrap tokens not being idempotent

wtripp180901:
Rerunning would previously cause an issue where the k3s server runtime would still attempt to set tokens despite them not being generated when all agents were already connected


sjpb (approved):
LGTM

PR #598, opened 2025-03-03
Fix: Support networks not owned by openstack project

bertiethorpe:
Follow on for https://github.com/stackhpc/ansible-slurm-appliance/pull/592.

In projects which don't own their networks, port security cannot be manipulated.

Fix sets `port_security_enabled` default to `null`. When `null`, security groups are applied.

### DIFF
--- a/environments/skeleton/{{cookiecutter.environment}}/tofu/control.tf
+++ b/environments/skeleton/{{cookiecutter.environment}}/tofu/control.tf
@@ -14,8 +14,8 @@ resource "openstack_networking_port_v2" "control" {
     subnet_id = data.openstack_networking_subnet_v2.cluster_subnet[each.key].id
   }
 
-  port_security_enabled = lookup(each.value, "port_security_enabled", true)
-  security_group_ids = lookup(each.value, "port_security_enabled", true) ? [for o in data.openstack_networking_secgroup_v2.nonlogin: o.id] : []
+  port_security_enabled = lookup(each.value, "port_security_enabled", null)
+  security_group_ids = lookup(each.value, "port_security_enabled", null) != false ? [for o in data.openstack_networking_secgroup_v2.nonlogin: o.id] : []
 
   binding {
     vnic_type = lookup(var.vnic_types, each.key, "normal")
--- a/environments/skeleton/{{cookiecutter.environment}}/tofu/node_group/nodes.tf
+++ b/environments/skeleton/{{cookiecutter.environment}}/tofu/node_group/nodes.tf
@@ -45,8 +45,8 @@ resource "openstack_networking_port_v2" "compute" {
     subnet_id = data.openstack_networking_subnet_v2.subnet[each.value.network].id
   }
 
-  port_security_enabled = lookup(each.value, "port_security_enabled", true)
-  security_group_ids = lookup(each.value, "port_security_enabled", true) ? var.security_group_ids : []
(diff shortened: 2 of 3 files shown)


environments/skeleton/{{cookiecutter.environment}}/tofu/control.tf
```diff
@@ -14,8 +14,8 @@ resource "openstack_networking_port_v2" "control" {
     subnet_id = data.openstack_networking_subnet_v2.cluster_subnet[each.key].id
   }
 
-  port_security_enabled = lookup(each.value, "port_security_enabled", true)
-  security_group_ids = lookup(each.value, "port_security_enabled", true) ? [for o in data.openstack_networking_secgroup_v2.nonlogin: o.id] : []
+  port_security_enabled = lookup(each.value, "port_security_enabled", null)
+  security_group_ids = lookup(each.value, "port_security_enabled", null) != false ? [for o in data.openstack_networking_secgroup_v2.nonlogin: o.id] : []
```
JohnGarbutt, line 18: Can we default this to null as well? Is that allowed?

JohnGarbutt: oops, ignore me, we really do want empty list for the null case here, i.e. we need to remove the default security group here.

environments/skeleton/{{cookiecutter.environment}}/tofu/node_group/nodes.tf
```diff
@@ -45,8 +45,8 @@ resource "openstack_networking_port_v2" "compute" {
     subnet_id = data.openstack_networking_subnet_v2.subnet[each.value.network].id
   }
 
-  port_security_enabled = lookup(each.value, "port_security_enabled", true)
-  security_group_ids = lookup(each.value, "port_security_enabled", true) ? var.security_group_ids : []
+  port_security_enabled = lookup(each.value, "port_security_enabled", null)
+  security_group_ids = lookup(each.value, "port_security_enabled", null) != false ? var.security_group_ids : []
```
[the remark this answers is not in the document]
JohnGarbutt: as above, ignore me.

JohnGarbutt (approved):
OK, this looks good, in that we ensure, by default, we don't sent any values for port_security, so we don't trigger the policy violation for networks where RBAC allows you to created ports (i.e. its shared) but you are not the owner of that network.

environments/skeleton/{{cookiecutter.environment}}/tofu/node_group/nodes.tf
sjpb, line 49: I don't understand why this wasn't just:

```
security_group_ids = lookup(each.value, "port_security_enabled", true) ? var.security_group_ids : []
```
what am I missing here?

bertiethorpe: agh you're right I think

priteau: This commit isn't working for me, it still applies the security groups even when port_security_enabled is false. Can you please switch to @sjpb's suggestion?

bertiethorpe: https://github.com/stackhpc/ansible-slurm-appliance/pull/601

PR #599, opened 2025-03-04
Root-squash nfs exports by default

sjpb:
Sets all nfs exports to be root-squashed by default and modifies the `basic_users` role to cope with this. See role README.md for example configurations.


sjpb:
Tested locally, including logging in to ondemand as `demo_user` and checking shell works/looks ok and jobs run

ansible/roles/basic_users/tasks/main.yml
```diff
@@ -21,53 +21,116 @@
...
-  register: basic_users_info
+    label: "{{ item.name }}"
+  vars:
+    _disable_homedir: # ensure this task doesn't touch $HOME
+      create_home: false
+      generate_ssh_key: false
+
+- name: Write sudo rules
+  blockinfile:
+    path: /etc/sudoers.d/80-{{ item.name}}-user
```
sd109, line 36: Nit:
```suggestion
    path: /etc/sudoers.d/80-{{ item.name }}-user
```

ansible/roles/basic_users/tasks/main.yml
```diff
@@ -21,53 +21,116 @@
...
+        manage_dir: false
+        key: "{{ item.public_key }}"
+        path: ~/.ssh/authorized_keys
+      become_user: "{{ item.item.name }}"
+      loop: "{{ _cluster_ssh_keypair.results }}"
+      loop_control:
+        label: "{{ item.item.name }}"
+      when:
+        - item.item.state | default('present') == 'present'
+        - "'public_key' in item"
```
sd109, line 121: Could we use:
```
loop_control:
    loop_var: ...
```
here to get rid of the `item.item` syntax and make it easier to understand what the nested loops are doing?

sd109 (changes requested):
LGTM functionality-wise, just a few minor suggestions

sd109 (approved):

PR #601, opened 2025-03-04
FIX: Tofu attempts to apply security groups when port_security_enabled is false

bertiethorpe:
Previous fix for https://github.com/stackhpc/ansible-slurm-appliance/pull/598 doesn't work.

`lookup(each.value, "port_security_enabled", null) != false ? var.security_group_ids : []`
still applies the security groups even when port_security_enabled is false.

EDIT:

Solution is to circumvent toggling port security entirely with https://registry.terraform.io/providers/vtdc/viettelidcops/latest/docs/resources/networking_port_v2#no_security_groups-1

Then disabling security groups is done by:
```
cluster_networks = [
  {
    network = "some_net"
    subnet = "some_subnet"
    no_security_groups = true
  }
]
```

### DIFF
--- a/environments/skeleton/{{cookiecutter.environment}}/tofu/variables.tf
+++ b/environments/skeleton/{{cookiecutter.environment}}/tofu/variables.tf
@@ -15,7 +15,7 @@ variable "cluster_networks" {
         List of mappings defining networks. Mapping key/values:
             network: Required. Name of existing network
             subnet: Required. Name of existing subnet
-            port_security_enabled: Optional. Bool, default null (for networks not owned by project)
+            no_security_groups: Optional. Bool (default: false). Disable security groups
     EOT
 }
 


sjpb:
So the actual error without this PR, when settting `port_security_enabled: false`  is:

```
│ Error: Error creating openstack_networking_port_v2: Expected HTTP response code [201 202] when accessing [POST https://create.leaf.cloud:9696/v2.0/ports], but got 400 instead: {"NeutronError": {"type": "PortSecurityAndIPRequiredForSecurityGroups", "message": "Port security must be enabled and port must have an IP address in order to use security groups.", "detail": ""}}
│ 
```

even screwier the logic appears to be evaluated properly:
```
> var.cluster_networks[0]
{
  "network" = "slurmapp-ci"
  "port_security_enabled" = false
  "subnet" = "slurmapp-ci"
}
> lookup(var.cluster_networks[0], "port_security_enabled", null) != false
false
```

jovial:
Can we omit the security groups if port_security is false?

jovial:
For me, it used to work when using admin creds. When I switched to non-admin creds, I actually just enabled port security on the network (even though I don't think it is implemented).

JohnGarbutt (approved):
To me, I think this makes sense. In some places we can skip adding security groups on some networks.

I like keeping port security on, unless we find a good use case to turn that off.

JohnGarbutt:
Do we put docs for this extra flag somewhere?

sjpb:
> Do we put docs for this extra flag somewhere?

Yeah this is not ready to merge. It needs documenting in the variables.tf file (in both places it applies)

PR #603, opened 2025-03-06
Disable nightly builds

bertiethorpe:
We don't use nightly builds for fatimages anymore, removing the nightly schedule to not waste resources.

Still allowing workflow dispatch as we might revisit at some point.


sjpb (approved):
LGTM

PR #605, opened 2025-03-06
Fix chrony for nodes w/o network access (yet)

sjpb:
Need chrony early in the process, and it is currently before squid is setup.
The chrony role uses dnf which is a no-op wrt the chrony package (as it is installed in genericcloud images), but because one repo (TurboVNC) is enabled by default this triggers a metadata download, which then fails if the node is proxing through squid.

Fix is simply to never run the install tasks.


jovial (approved):
LGTM

PR #607, opened 2025-03-07
Restrict all nfs shares to nfs group IPs

sjpb:


environments/common/inventory/group_vars/all/nfs.yml
```diff
@@ -5,17 +5,24 @@
 
 nfs_server_default: "{{ groups['control'] | first }}" # avoid using hostvars for compute-init
 
+# only allow the nfs group IPs to mount nfs exports:
+nfs_export_clients_default: "{{ groups['nfs'] | map('extract', hostvars, 'ansible_host') | join(' ') }}"
+nfs_export_clients: "{{ nfs_export_clients_default }}"
```
m-bull, line 10: I think we should use this on L28 or get rid of it

m-bull (changes requested):

environments/common/inventory/group_vars/all/nfs.yml
sjpb: We do? I don't undestand:
```
# NB: this is set as default for all shares above but is repeated here
# to prevevent being accidently overriden when adding shares
nfs_export_clients: "{{ nfs_export_clients_default }}"
```

m-bull: We use the value that it is set to, but not the variable itself, I don't think?

m-bull (approved):
Cheers @sjpb

environments/common/inventory/group_vars/all/nfs.yml
sjpb: Clarified in discussion this is how the role works, it takes both top-level vars and keys in `nfs_configurations`.

PR #613, opened 2025-03-11
Fix documentation of sssd_install_ldap variable

priteau:
The sssd README was documenting the variable as `sssd_ldap_install` when it is actually `sssd_install_ldap`.


sjpb (approved):
LGTM

PR #600, opened 2025-03-04
Add file deletion to cleanup play

sjpb:
Add a cleanup task in image build to delete unnecessary files triggering Trivy vulerability scans. Note this can only be used where removal of the file is sufficent to remove the potential vulerability.

PR #615, opened 2025-03-11
Fix builds not in stackhpc env

sjpb:
Currently builds not using the stackhpc fail because https://github.com/stackhpc/ansible-slurm-appliance/pull/600 means that cleanup.yml needed the ondemand package version which was only defined for the stackhpc environment.


bertiethorpe (approved):
LGTM

priteau (approved):
I just ran into this issue earlier, thanks for fixing it.

PR #617, opened 2025-03-12
Control default routes on boot

sjpb:
Adds a `gateway_ip` OpenTofu variable which can be used to ensure (only) a default route via the specified IP exists.

Tested that:
- it can add a default route if none is provided by the subnet
- it can remove a default route on a dual-homed host where both subnets provide one
- it makes no changes if the specified default route matches the subnet provided one

### DIFF
--- a/ansible/.gitignore
+++ b/ansible/.gitignore
@@ -84,3 +84,5 @@ roles/*
 !roles/pytools/**
 !roles/rebuild/
 !roles/rebuild/**
+!roles/gateway/
+!roles/gateway/**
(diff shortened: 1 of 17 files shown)


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/13809298235

sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/13810727551

jovial (approved):
All looks sensible to me

PR #622, opened 2025-03-14
Removed unused Pulp variables

wtripp180901:
Architecture is hardcoded to x86 in timestamp paths so these aren't actually used anywhere


sjpb:
I'm not merging this, I've got more changes coming as part of a feature branch.

PR #626, opened 2025-03-14
Enable creating users with local homedirs

### LINKED ISSUE #625: Revised basic_users approach broken for client config
Client has 

```yaml
basic_users_users_grid_defaults:
  create_home: true
  generate_ssh_key: false
```

which gets inserted into each basic_user_users directory. These users have local homedirs which are not on the shared filesystem.

With https://github.com/stackhpc/ansible-slurm-appliance/pull/599 there is no way to modify their config to achieive this.

sjpb:
Fixes #625 

Also should allow the role as-is by compute-init as `basic_users_homedir_server` and `basic_users_homedir_client` won't, by default, match the inventory_hostname when running via ansible-init. However if these are overriden to be `inventory_hostname` then both directory creation and ssh key management run on each host, allowing non-shared home directories to be used both for normal ansible and compute-init.

### DIFF
--- a/ansible/roles/basic_users/README.md
+++ b/ansible/roles/basic_users/README.md
@@ -11,23 +11,37 @@ without requiring LDAP etc. Features:
 - Login to the control node is prevented (by default).
 - When deleting users, systemd user sessions are terminated first.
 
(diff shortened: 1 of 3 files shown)


bertiethorpe (approved):
LGTM

PR #630, opened 2025-03-21
Allow use of Ironic node name as instance name

priteau:
This can be used in bare metal deployments where we may want the instance hostname to be exactly the same as the node name.

### DIFF
--- a/environments/skeleton/{{cookiecutter.environment}}/tofu/compute.tf
+++ b/environments/skeleton/{{cookiecutter.environment}}/tofu/compute.tf
@@ -27,6 +27,7 @@ module "compute" {
   ignore_image_changes = lookup(each.value, "ignore_image_changes", false)
   match_ironic_node = lookup(each.value, "match_ironic_node", false)
   availability_zone = lookup(each.value, "availability_zone", "nova")
+  use_ironic_node_name = lookup(each.value, "use_ironic_node_name", false)
 
   # computed
   # not using openstack_compute_instance_v2.control.access_ip_v4 to avoid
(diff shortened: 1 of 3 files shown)


sjpb (commented):
Are you trying to change the hostname, or just the instance name? If you want to change the former too you'll need to change the fqdn in the cloud-init userdata for each instance definition

priteau:
> Are you trying to change the hostname, or just the instance name? If you want to change the former too you'll need to change the fqdn in the cloud-init userdata for each instance definition

In this case, I think both. If the cluster name is `foo`, but bare metal nodes are already called `foo-cpu-01`, `foo-cpu-02`, I want both the instances and the hostnames to be using `foo-cpu-01`, not `foo-foo-cpu-01`, while keeping the automatic naming of the login and control VMs.

sjpb:
Ok, will wait for changes then. I get the intent; I'd note that if you end up moving an ironic node between production/staging to test things, it won't be obvious which cluster is it part of. Although we do set the path to the environment as part of the instance metadata.

priteau:
> Ok, will wait for changes then. I get the intent; I'd note that if you end up moving an ironic node between production/staging to test things, it won't be obvious which cluster is it part of. Although we do set the path to the environment as part of the instance metadata.

I am planning to only use this feature in production. Staging would use the prefix, to make sure nodes can easily be differentiated.

sjpb:
Should replace with https://github.com/stackhpc/ansible-slurm-appliance/pull/651 for more flexibility and to satisfy overlapping requirements.

PR #632, opened 2025-03-25
Attempt leafcloud build on local disk to compare speed

JohnGarbutt:
Leafcloud image builds are volume-backed by default, which causes slow build times

Building images on disk should be quicker for the fatimage build, though volumes are still required for extra-builds


packer/openstack.pkr.hcl
```diff
@@ -178,7 +178,7 @@ source "openstack" "openhpc" {
   ssh_bastion_private_key_file = var.ssh_bastion_private_key_file
   
   # Output image:
-  image_disk_format = "qcow2"
+  # image_disk_format = "qcow2"
```
JohnGarbutt, line 181: This should be set to null when use_blockstorage_volume == false, and otherwise got to "qcow2".

bertiethorpe:
20 min fatimage build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14197320832/job/39775471558

.github/workflows/extra.yml
```diff
@@ -118,10 +119,16 @@ jobs:
           echo $IMAGE_ID > image-id.txt
           echo $IMAGE_NAME > image-name.txt
 
-      - name: Make image usable for further builds
+      - name: Make image usable for further builds (if signature_verified property exists)
```
sjpb, line 122: Why do we need this check here, but not for the same in the fatimage workflow below?

.github/workflows/extra.yml
```diff
@@ -101,6 +101,7 @@ jobs:
           -var "source_image_name=${{ fromJSON(env.FAT_IMAGES)['cluster_image'][matrix.build.source_image_name_key] }}" \
           -var "image_name=${{ matrix.build.image_name }}" \
           -var "inventory_groups=${{ matrix.build.inventory_groups }}" \
+          -var "use_blockstorage_volume=true" \
```
sjpb, line 104: So this will be for all clouds. Is that OK? If not, I would set it in e.g. `environments/.stackhpc/LEAFCLOUD.pkrvars.hcl` instead.

packer/openstack.pkr.hcl
```diff
@@ -109,7 +109,7 @@ variable "manifest_output_path" {
 
 variable "use_blockstorage_volume" {
   type = bool
-  default = true
+  default = false
```
sjpb, line 112: If we're doing this, can you modify docs/image-build.md b/c it currently says vol backed is default. Or, you could just set it to false in CI.

sjpb (changes requested):
Think it needs testing on SMS too TBH.

sjpb:
Replaced by #650

PR #637, opened 2025-04-01
Allow empty items in extra package and user lists

priteau:
This enables defining extra packages and users based on some conditions. For example:

    appliances_extra_packages_other:
      - "{{ 'cuda-toolkit' if 'cuda' in group_names }}"

### DIFF
--- a/environments/common/inventory/group_vars/all/defaults.yml
+++ b/environments/common/inventory/group_vars/all/defaults.yml
@@ -79,7 +79,7 @@ appliances_local_users_default:
 
 # Overide this to add extra users whilst keeping the defaults.
 appliances_local_users_extra: [] # see format of appliances_local_users_default above
-appliances_local_users: "{{ appliances_local_users_default + appliances_local_users_extra }}"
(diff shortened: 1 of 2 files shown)


sjpb:
Normally we'd just do this via some indirection:

```
appliances_extra_packages_cuda:
  - 'cuda-toolkit'
appliances_extra_packages_other: "{{ if appliances_extra_packages_cuda 'cuda' in group_names else [] }}"
```

but seems fine.

sjpb (approved):
LGTM, tested logic locally and the select | list is enough to ditch the empty string you get.

priteau:
> Normally we'd just do this via some indirection:
> 
> ```
> appliances_extra_packages_cuda:
>   - 'cuda-toolkit'
> appliances_extra_packages_other: "{{ if appliances_extra_packages_cuda 'cuda' in group_names else [] }}"
> ```
> 
> but seems fine.

It is feasible with other methods, but with this approach we don't need to define additional variables.

PR #640, opened 2025-04-01
Correct vnic_types var name in skeleton variables

MoteHue:
vnic_types is now a mapping from networks to types.


priteau (approved):

sjpb (approved):
Ouch, good catch, thanks. Reviewed via grep, appears to fix variable docs to match actual code.

PR #641, opened 2025-04-01
Use checksum verification for CernVM-FS GPG key

priteau:
The cvmrepo repository is sometimes down. This avoids fetching the GPG key each time if it was already done, preventing Ansible failures.

It also verifies that the key is the expected one instead of blindly trusting any GPG key.

This should not require much maintenance since the key appears to be the same since it was generated in 2010.


priteau:
I have seen this issue multiple times already.

cvmrepo

sjpb (approved):
LGTM, thanks

PR #643, opened 2025-04-04
Support fixed IP addresses for nodes

### LINKED ISSUE #642: Production variable control_ip_address is not implemented
The variable `control_ip_address` described in the [Production Deployments](https://github.com/stackhpc/ansible-slurm-appliance/blob/main/docs/production.md) documentation is not implemented.

It should also support multiple networks rather than a single fixed IP.

priteau:
Add support for optional fixed IP addresses via opentofu variable `control_ip_addresses` and parameter `ip_addresses` for compute and login node groups.

Example:

```terraform
cluster_networks = [
    {
        network = "foo"
        subnet = "bar"
    }
]
control_ip_addresses = {
   foo = "192.168.0.10"
}
login = {
    interactive = {
        nodes = ["login-0"]
        ...
        ip_addresses = {
            foo = ["192.168.0.20"]
        }
compute = {
    standard = {
        nodes = ["standard-0", "standard-1"]
        ...
        ip_addresses = {
            foo = ["192.168.0.30", "192.168.0.31"]
        }
}
```

Closes #642.

### DIFF
--- a/docs/production.md
+++ b/docs/production.md
@@ -134,22 +134,9 @@ and referenced from the `site` and `production` environments, e.g.:
 
 - Consider whether having (read-only) access to Grafana without login is OK. If not, remove `grafana_auth_anonymous` in `environments/$ENV/inventory/group_vars/all/grafana.yml`
 
-- Modify `environments/site/tofu/nodes.tf` to provide fixed IPs for at least
-  the control node, and (if not using FIPs) the login node(s):
-
-    ```
-    resource "openstack_networking_port_v2" "control" {
-        ...
-        fixed_ip {
-            subnet_id  = data.openstack_networking_subnet_v2.cluster_subnet.id
-            ip_address = (var.control_ip_addresses != {})? var.control_ip_addresses[each.key]: null
-        }
-    }
-    ```
-    
-  Note the variable `control_ip_addresses` is new.
-
-  Using fixed IPs will require either using admin credentials or policy changes.
+- If fixed IP(s) are required for the control node, set the OpenTofu variable
+  `control_ip_addresses`. This will require either using admin credentials or
+  policy changes.
 
 - If floating IPs are required for login nodes, modify the OpenTofu configurations
   appropriately.
--- a/docs/production.md
+++ b/docs/production.md
@@ -134,10 +134,6 @@ and referenced from the `site` and `production` environments, e.g.:
 
 - Consider whether having (read-only) access to Grafana without login is OK. If not, remove `grafana_auth_anonymous` in `environments/$ENV/inventory/group_vars/all/grafana.yml`
 
-- If fixed IP(s) are required for the control node, set the OpenTofu variable
-  `control_ip_addresses`. This will require either using admin credentials or
-  policy changes.
-
 - If floating IPs are required for login nodes, modify the OpenTofu configurations
   appropriately.
 
--- a/docs/production.md
+++ b/docs/production.md
@@ -141,13 +141,13 @@ and referenced from the `site` and `production` environments, e.g.:
     resource "openstack_networking_port_v2" "control" {
         ...
         fixed_ip {
-            subnet_id = data.openstack_networking_subnet_v2.cluster_subnet.id
-            ip_address = var.control_ip_address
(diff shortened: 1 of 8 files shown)


environments/skeleton/{{cookiecutter.environment}}/tofu/control.tf
```diff
@@ -11,7 +11,8 @@ resource "openstack_networking_port_v2" "control" {
   admin_state_up = "true"
 
   fixed_ip {
-    subnet_id = data.openstack_networking_subnet_v2.cluster_subnet[each.key].id
+    subnet_id  = data.openstack_networking_subnet_v2.cluster_subnet[each.key].id
+    ip_address = (var.control_ip_addresses != {})? var.control_ip_addresses[each.key]: null
```
sjpb, line 15: This resource has a for_each parameter. So the particular resource instance (i.e, port) for each item in that for_each has a has a _single_ ip address, not a list. Else you're trying to assign multiple IPs from different subnets to _each_ port.

sjpb (changes requested):

docs/production.md
```diff
@@ -134,22 +134,9 @@ and referenced from the `site` and `production` environments, e.g.:
...
-        }
-    }
-    ```
-    
-  Note the variable `control_ip_address` is new.
-
-  Using fixed IPs will require either using admin credentials or policy changes.
+- If fixed IP(s) are required for the control node, set the OpenTofu variable
+  `control_ip_addresses`. This will require either using admin credentials or
+  policy changes.
```
m-bull, line 139: Describe the format of this variable in the docs here to save going hunting for it in the Tofu

environments/skeleton/{{cookiecutter.environment}}/tofu/variables.tf
```diff
@@ -24,6 +24,20 @@ variable "key_pair" {
     description = "Name of an existing keypair in OpenStack"
 }
 
+variable "control_ip_addresses" {
+    type        = map(string)
```
m-bull, line 28: Something about this feels a bit wrong, but I can see why its been done for ease of using the data structure.

I think I'd prefer to have a list of maps that have consistent keys, like this:

```
control_ip_addresses = [{"network-name" = "net1", "fixed-ip" = "192.168.10.2" }]
```

But I can see why the approach has been taken.

m-bull (changes requested):

environments/skeleton/{{cookiecutter.environment}}/tofu/variables.tf
sjpb: We're going to be expanding this to all nodes too in a subsequent PR. I'm expecting/planning that the `login` and `compute` groups will a also get an `ip_addresses` key which should be the same, except the values here will be a list in same order as `nodes`. So might end up being configured like:


```terraform
compute = {
  standard = {
    nodes: ["compute-0", "compute-1"]
    flavor: var.other_node_flavor
    extra_networks = [
      {
        network = "data"
        subnet = "data"
      }        
    ]
    ip_addresses = [
      {
         network_name = "cluster"
         fixed_ips = [192.168.0.10, 192.168.0.11]
      },
      { 
         network_name = "data"
         fixed_ips = [172.16.2.10, 172.16.2.11]
      }
    ]
  }
}
```

I'm not sure I see that as being better than just having `{network: ips}` as in this PR at present TBH:

```terraform
compute = {
  standard = {
    nodes: ["compute-0", "compute-1"]
    flavor: var.other_node_flavor
    extra_networks = [
      {
        network = "data"
        subnet = "data"
      }        
    ]
    ip_addresses = {
      cluster = [192.168.0.10, 192.168.0.11]
      data = [172.16.2.10, 172.16.2.11]
    }
  }
}
```

docs/production.md
sjpb: I'm going to remove this actually, I don't think this is part of "standard production setup", more a special-case.

sjpb (approved):
LGTM

sjpb:
Tested on smslabs. Works. Correct IPs set in metadata on all nodes. site playbook running ok.

bertiethorpe (approved):
LGTM

PR #647, opened 2025-04-08
bump CI image

sjpb:
testing whether the logic re. waiting for nfs in compute-init is borked.

Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14335855199


sjpb:
turns out this wasn't the problem

PR #651, opened 2025-04-15
Allow modifying nodes fully-qualified name

sjpb:
Adds an opentofu variable `cluster_nodename_template` plus overrides `nodename_template` for compute and login node groups, to allow defining the fully-qualified hostname for all nodes.

Note the appliance (and roles it uses) assume that the short hostname is resolvable (normally ensured by `etc_hosts` role), and that this matches the Ansible `inventory_hostname` and the slurm node name.

### DIFF
--- a/.github/workflows/stackhpc.yml
+++ b/.github/workflows/stackhpc.yml
@@ -91,9 +91,9 @@ jobs:
         run: dev/setup-env.sh
 
       - name: Install OpenTofu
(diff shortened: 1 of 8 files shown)


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14860103783

MoteHue (approved):

wtripp180901 (approved):
LGTM

PR #652, opened 2025-04-16
Fix cuda installs

MoteHue:
Bumps cuda to 12.8.


sjpb (approved):
LGTM, thanks.

No need for new image build.

PR #654, opened 2025-04-23
Add support for Node Health Checks

sjpb:
See ansible/roles/nhc/README.md. Checks are templated out on each compute node during site.yml and then restored after a slurm-controlled rebuild, so that checks after rebuild ensure the node is in the same state as it was before.

Note this also moves the configuration of nfs-exported data for compute-init to the end of the site.yml playbook, to make ordering of roles requiring compute-init support easier.

### DIFF
--- a/environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
+++ b/environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
@@ -1,6 +1,6 @@
 {
     "cluster_image": {
-        "RL8": "openhpc-RL8-250619-1728-4209e400",
-        "RL9": "openhpc-RL9-250619-1728-4209e400"
+        "RL8": "openhpc-RL8-250620-1251-d606a45c",
+        "RL9": "openhpc-RL9-250620-1251-d606a45c"
     }
 }


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14621365864

sjpb:
Failed manual checks when deploying:
```
TASK [nhc : Fetch generated NHC config to control node storage] ********************************************************
Wednesday 18 June 2025  09:37:50 +0000 (0:00:00.134)       0:12:55.714 ******** 
fatal: [RL9-compute-1]: FAILED! => {}

MSG:

Unable to create local directories(/exports/cluster/hostconfigRL9-compute-1): [Errno 13] Permission denied: b'/exports'
fatal: [RL9-compute-0]: FAILED! => {}

MSG:

Unable to create local directories(/exports/cluster/hostconfigRL9-compute-0): [Errno 13] Permission denied: b'/exports'
```

sjpb:
Ok so the problem is:
- The image we're using on this branch now has the `npc-ohpc` package in it.
- In CI, there is a point where the control node and login node are on that image (b/c TF has upgraded them to it) but the compute nodes aren't (b/c they have slurm-controlled rebuild). We run site.yml at that point.
- The slurm.yml playbook skips the openhpc installs for speed. But this means the compute nodes don't get that package. Then the playbook tries to configure nhc, which fails.

sjpb:
Fatimage build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/15762514081

sjpb:
Fat image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/15763342110

sjpb:
Above is failing on the /etc/nhc dir not existing. This is b/c that is created on install of the `nhc-ohpc` package, which is only in the *new* image, which the nodes don't have when running site.yml after upgrading login/control nodes.

wtripp180901 (approved):
Think most of these refactors have been reviewed previously and NHC installs all look sensible so I'm happy to approve

PR #655, opened 2025-04-23
Fix fatimage build without alertmanager secret

sjpb:
`cloudalchemy.prometheus:preflight.yml` checks `prometheus_alertmanager_config`, which means it must be templatable even during build when it is not used. Build in CI does not generate secrets, (deliberately, so we check it is not injecting them at all) so `vault_alertmanager_admin_password` was not defined.


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14622374485

wtripp180901 (approved):
LGTM

PR #657, opened 2025-04-24
Refactor toplevel site.yml and fatimage.yml playbooks

sjpb:
Refactoring to resolve a number of problems and inconsistencies between the `site.yml` and `fatimage.yml` playbooks. The major ones are:
- `iam.yml` and `slurm.yml` playbooks are now run directly from `fatimage.yml` to make making changes to these less error-prone.
- `compute-init` export task moved to a new `final.yml` playbook so it will run after any role which generates configuration which it might require.  `install.yml` task for this role also been moved here for consistency - note this changes commands used for development.
- `gateway` role which installs ansible-init playbook moved adjacent to ansible-init install
- Ordering in `site.yml` and `fatimage.yml` is now consistent in that repos are disabled _after_ any post-hook in both cases.
- Unrunnable build VM cleanup code removed from `site.yml`.
- Correctly define dependency of roles injecting ansible-init playbooks (`compute_init`, `gateway`) in groups hierarchy.

See https://wiki.stackhpc.com/doc/slurm-top-level-playbooks-HW9BlHDMpm for more detail

TODO:
- [ ]: Fix whatever change in mysql means the slurm db isn't getting persisted - probably writing the config during build or something with the appliances_state_directory not set or something.


sjpb:
Build for the above image bump: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14660430262/job/41143461025

sjpb:
Interactive testing @ [3120e49](https://github.com/stackhpc/ansible-slurm-appliance/pull/657/commits/3120e4961ec54605d100a4064c66be3efd859a15):

:heavy_check_mark:  ondemand shell
:heavy_check_mark:   ondemand remote desktop
:heavy_check_mark: ondemand files: 
:heavy_check_mark: ondemand jupyter
:heavy_check_mark: ondemand monitoring

ansible/iam.yml
```diff
@@ -47,5 +55,6 @@
   tags: sssd
   tasks:
     - name: Configure sssd
-      import_role:
+      ansible.builtin.include_role:
         name: sssd
+        tasks_from: "{{ 'install.yml' if appliances_mode == 'build' else 'main.yml' }}"
```
wtripp180901, line 60:
```suggestion
        tasks_from: "{{ {'build':'install.yml', 'configure':'configure.yml'}[appliances_mode] | default('main.yml') }}"
```

ansible/slurm.yml
```diff
@@ -3,11 +3,11 @@
 - name: Setup DB
   hosts: mysql
   become: true
-  tags:
-    - mysql
+  tags: mysql
   tasks:
     - include_role:
         name:  mysql
+        tasks_from: "{{ 'install.yml' if appliances_mode == 'build' else 'main.yml' }}"
```
wtripp180901, line 10:
```suggestion
        tasks_from: "{{ {'build':'install.yml', 'configure':'configure.yml'}[appliances_mode] | default('main.yml') }}"
```

environments/common/layouts/everything
```diff
@@ -123,5 +124,5 @@ builder
 # Hosts where crony configuration is applied. See docs/chrony.md for more details.
 
 [gateway:children]
-# Add builder to this group to install gateway ansible-init playbook into image
-builder
+# Hosts to install compute_init playbook on during image build to configure ip gateway on boot
```
wtripp180901, line 127: maybe make it clearer that the configuration on boot is performed separately from this group via the ansible_init group which has this as a dependency

environments/common/inventory/groups
```diff
@@ -181,4 +187,4 @@ extra_packages
 # Hosts where crony configuration is applied. See docs/chrony.md for more details.
 
 [gateway]
-# Add builder to this group to install gateway ansible-init playbook into image
+# Hosts to install compute_init playbook on during image build to configure ip gateway on boot
```
wtripp180901, line 190: See above

wtripp180901 (changes requested):
Looks good, a few suggestions for keeping the pattern of how appliances_mode is used consistent. As you said it may also worth documenting how appliances_mode is intended to be used vs builder groups

environments/common/layouts/everything
sjpb: Agh, yeah you're right. I think the new pattern is correct - and the old was wrong - problem with using `builder` as the switch is that it always runs on build, even during an "extra build". But that is a bit hard to explain.

sjpb:
Testing image build - should rebuild once fix/zenith merged to main anyway: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14778751137

sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14878191724

PR #659, opened 2025-04-28
Fix creation of hpctests directory

priteau:
The hpctests role creates a directory that is user-owned and group-owned by hpctests_user, which defaults to ansible_user. However, this group is not guaranteed to exist, especially in LDAP environments where groups dedicated to a single user are often not used.

Support customising the value via the hpctests_group variable, while still defaulting to hpctests_user.


sjpb (approved):
LGTM

PR #664, opened 2025-05-02
Update grafana to v10 using Ark rpms

sjpb:
Bumps grafana to v10 and uses Ark for stackhpc build, as upstream grafana repos appear to have removed 9.5.21 on or after 2025-04-25. As ark was in on-demand mode the packages are not there either.

Also makes it possible to change whether Packer cleans up resources from failed builds or not on a per-run basis.

Notes:
1.  Ark is now in immediate mode for grafana, see https://github.com/stackhpc/stackhpc-release-train/pull/399.
2. All Granfana dashboards are showing warnings for [Angular depreciation](https://grafana.com/docs/grafana/latest/developers/angular_deprecation/), needs to be fixed before moving to v11.
3. The (forked) cloudalchemy.grafana role creates a repofile, and does so disabled to work around previous issues with grafana repos being unavaiable. In order to make minimal changes, this adds a custom repofile with the same content which is generated by the dnf_repos role.
4. Although a single repo is used for both RockyLinux 8 & 9, separate repo structures and snapshots have been added in case these need to diverge in future, and for consistency with other repo definitions.

### DIFF
--- a/environments/common/files/grafana/grafana.repo.j2
+++ b/environments/common/files/grafana/grafana.repo.j2
@@ -1,0 +1,8 @@
+{{ ansible_managed | comment }}
+[grafana]
+baseurl = {{ appliances_pulp_url }}/pulp/content/{{ appliances_pulp_repos.grafana[ansible_distribution_major_version] | appliances_repo_to_subpath }}
+name = grafana
+async = 1
+gpgcheck = 0
+password = {{ dnf_repos_password }}
+username = {{ dnf_repos_username }}
--- a/environments/common/files/grafana/grafana.repo.j2
+++ b/environments/common/files/grafana/grafana.repo.j2
@@ -1,6 +1,7 @@
 {{ ansible_managed | comment }}
 [grafana]
 baseurl = {{ appliances_pulp_url }}/pulp/content/{{ appliances_pulp_repos.grafana[ansible_distribution_major_version] | appliances_repo_to_subpath }}
+enabled = 0
 name = grafana
 async = 1
 gpgcheck = 0
--- a/environments/common/files/grafana/grafana.repo.j2
+++ b/environments/common/files/grafana/grafana.repo.j2
@@ -5,7 +5,7 @@ enabled = 0
 name = grafana
 async = 1
 gpgcheck = 0
-{% if dnf_repos_password is defined %}
+{% if 'dnf_repos' in group_names and dnf_repos_password is defined %}
 password = {{ dnf_repos_password }}
 username = {{ dnf_repos_username }}
 {% endif %}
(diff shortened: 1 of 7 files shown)


sjpb:
Fatimage build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14796564290

sjpb:
Testing the above build manually failed, repo logic is wrong.

Fatimage build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14797550985

sjpb:
Tested all dashboards at this commit. All worked, all warned re. angular.

environments/common/files/grafana/grafana.repo.j2
```diff
@@ -0,0 +1,9 @@
+{{ ansible_managed | comment }}
+[grafana]
+baseurl = {{ appliances_pulp_url }}/pulp/content/{{ appliances_pulp_repos.grafana[ansible_distribution_major_version] | appliances_repo_to_subpath }}
+enabled = 0
+name = grafana
+async = 1
+gpgcheck = 0
+password = {{ dnf_repos_password }}
```
wtripp180901, line 8: this is going to leak creds, so some mechanism is needed to re-template this without username and password before the build finishes. We also need to account for the case of using a local Pulp server where the repo is unauthenticated. Might be worth just biting the bullet and trying to get rid of this altogether. If not we'll probably want a cred leak check similar to https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/bootstrap.yml#L135

wtripp180901 (changes requested):
If we're duplicating the repofile in the Grafana role, we'll need some way of protecting Ark creds

sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14857401425

environments/common/files/grafana/grafana.repo.j2
sjpb: Yeah, also breaks site.yml where dnf_repos_password isn't defined. Will still use dnf_repos to do the cleanup.

wtripp180901 (dismissed):
LGTM

[edit added from Slack chat:] there is still one potential case where creds can get leaked, if you have dnf_repos_password defined but dnf_repos not in groups [then the check in bootstrap doesn't get run]. If you're using creds to do a local build from Ark directly, you could then run site without removing the creds, where dnf_repos wouldn't be in groups since iirc its just set to builder

sjpb:
Image build https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/14860103783

wtripp180901 (approved):
LGTM

PR #665, opened 2025-05-08
Adds support for disabling the generation of partitions.yml

jovial:
This allows you to manually create your own variant. You can use the auto-generated one as a template.


sjpb:
I'm wondering if there's a better approach where we can make the environments/common/group_vars/openhpc.yml "do the default thing" then you can just dump stuff in your site group_vars, as for normal.

PR #667, opened 2025-05-09
Allow enabling package installs for caas clusters via extravars

sjpb:
Allows caas to be configured from Azimuth extravars to enable package installs. Requires the following to be set:

```yaml
dnf_repos_enabled: true
dnf_repos_username: $CLIENT_USERNAME
dnf_repos_password: $CLIENT_PASSWORD
dnf_repos_allow_insecure_creds: true
```

> [!WARNING]
> This will expose Ark credentials to the `azimuth` user.

### DIFF
--- a/ansible/disable-repos.yml
+++ b/ansible/disable-repos.yml
@@ -5,3 +5,4 @@
       ansible.builtin.include_role:
         name: dnf_repos
         tasks_from: disable_repos.yml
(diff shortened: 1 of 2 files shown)


sjpb:
Tested by provisioning on Azimuth and installing emacs on login, compute & control nodes. All OK.

sd109 (approved):
LGTM

PR #673, opened 2025-05-16
Make home volume creation optional

sjpb:
- Adds a new OpenTofu variable `home_volume_provisioning` to support production cluster configurations. See docs/production.md.
- Automatically disables the NFS home share when the home volume is disabled (e.g. when home directories are provided via manila).
- Provides new inventory variable `nfs_configurations_extra` to make it easier to add site-specific NFS shares.


docs/production.md
```diff
@@ -76,18 +76,39 @@ and referenced from the `site` and `production` environments, e.g.:
...
-  resources](https://opentofu.org/docs/language/data-sources/). This ensures that even if the cluster is deleted via tofu, the
-  volumes will persist.
-
-  For a development environment, having volumes under tofu control via volume
-  resources is usually appropriate as there may be many instantiations
-  of this environment.
-
+- By default, the cookiecutter provided OpenTofu configurations provision
+  a volume named "$cluster_name-state" and attach it to the control node. This
+  is used to persist monitoring and Slurm data when the control node is re-built.
```
wtripp180901: Looks like fragments of old docs, none of the changes in this PR affect the state volume

environments/skeleton/{{cookiecutter.environment}}/tofu/control.tf
```diff
@@ -1,5 +1,9 @@
 locals {
-  control_volumes = concat([openstack_blockstorage_volume_v3.state], var.home_volume_size > 0 ? [openstack_blockstorage_volume_v3.home][0] : [])
+  control_volumes = concat(
+    [openstack_blockstorage_volume_v3.state],
+    # convert map to list:
+    [for v in data.openstack_blockstorage_volume_v3.home: v]
```
wtripp180901, line 5: may be worth reiterating that this is only a loop to check for the existence of the volume and so will only produce a 0 or 1 length list

wtripp180901 (changes requested):
Looks good, docs just need updating

sjpb:
Tested state volume is correctly attached both with manage and with attach. Also tested migration route described in production docs for existing clusters.

wtripp180901 (approved):
LGTM

PR #678, opened 2025-05-22
Production end to end deployment docs

MoteHue:


docs/production.md
```diff
@@ -142,12 +330,31 @@ and referenced from the `site` and `production` environments, e.g.:
...
+- Enable alertmanager if Slack is available - see
+  [docs/alerting.md](./alerting.md).
+
+- By default, the appliance uses a built-in NFS share backed by an OpenStack
+  volume for the cluster home directories. You may find that you want to change
+  this. The following alternatives are supported:
+
+  - External NFS <!--- External NFS docs TODO ---> 
+  - CephFS via OpenStack Manila <!--- filesystems docs TODO ---> 
+  - [Lustre](../roles/lustre/README.md)
```
MoteHue, line 352: Open question: are the any others I've missed?

docs/production.md
```diff
@@ -1,88 +1,267 @@
...
+  ansible-playbook ansible/site.yml
+  ```
+
+Once it completes you can log in to the cluster using:
+
+  ```bash
+  ./dev/ansible-ssh login
+  ```
+
+## Production further configuration
```
MoteHue, line 243: Open question: does anyone have opinions on ordering here? I'm trying to write the points that roughly relate next to each other, wondering if it could be worth the effort to have a more structured order...

sjpb (changes requested):
Few suggestions

docs/production.md
sjpb: I think you can leave out external NFS docs for the moment really, we can add those into filesystems at some point. Generally I think using NFS is a bad idea!

docs/production.md
```diff
@@ -1,51 +1,247 @@
...
 
 - Vault-encrypt secrets. Running the `generate-passwords.yml` playbook creates
   a secrets file at `environments/$ENV/inventory/group_vars/all/secrets.yml`.
   To ensure staging environments are a good model for production this should
```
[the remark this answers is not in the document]
MoteHue: That was the agreement, yes. I'll update this to suggest it.

sjpb (approved):
LGTM, thanks

PR #679, opened 2025-05-23
Support cluster deploy with only Ansible tasks proxied

sjpb:
This PR provides a capacity to create clusters in an environment where the only internet access is via a proxy, where the proxy should not be available to cluster users.

It does this by:

* Where possible, skipping dnf install tasks for packages already installed in StackHPC images, to avoid triggering dnf cache updates
* For tasks actually requiring internet access, proxy variables are set via ansible’s [[environment keyword](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_environment.html)](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_environment.html). This means the proxy variables are only set during plays and are not present on the host. The proxy should use basic authentication to prevent un-authorised use from inside the cluster.
* TEMPORARY: Removes any grafana dashboards sourced from downloads. These should be replaced with in-appliance copies (the curl command used for other reasons doens’t pick up the proxy vars, and anyway we want these in the appliance to perform upgrades to grafana v11 at some point)

Currently this supports:
- Those features enabled in the everything template LINK TODO (as copied into a new environment)
- Features enabled in the stackhpc environment.
- FreeIPA client (not tested)

To configure this:
1. In your environment:
   1. Enable proxying for the cluster by adding `cluster` to the `proxy` group, see e.g. `environments/.stackhpc/inventory/extra_groups`
   2. Set `proxy_http_proxy` as for `environments/.stackhpc/inventory/group_vars/all/proxy.yml`. Note that this should be in the format `http://user:password@address:port`. In general address should be an IP address, not a hostname (TODO: explain). The corresponding proxy_https and no_proxy variables are derived automatically, see ansible/roles/proxy/defaults.md
   3. Set proxy_plays_only to true as shown in 2). This means the proxy is only available to Ansible.

Note: because the `proxy` role must be run to set the host facts for proxying, this precludes running indivudual playbooks during development. To work around this for development it would be possible to set the relevant variables directly, so it might look like:

```yaml
# environments/$ENV/inventory/group_vars/all/proxy.yml

# your actual proxy configuration as per 1.b above:
proxy_http_proxy: http://user:password@address:port
proxy_plays_only: true

# from proxy role defaults:
proxy_https_proxy: "{{ proxy_http_proxy }}"
proxy_no_proxy_defaults: "{{ ['localhost', '127.0.0.1'] + groups['all'] + hostvars.values() | map(attribute='ansible_host') }}"
proxy_no_proxy_extras: []
proxy_no_proxy: "{{ (proxy_no_proxy_defaults + proxy_no_proxy_extras) | unique | sort | join(',') }}"
proxy_dnf: "{{ not proxy_plays_only }}"
proxy_systemd: "{{ not proxy_plays_only }}"

# then do what the role actually does:
appliances_environment_vars:
  http_proxy: "{{ proxy_http_proxy }}"
  https_proxy: "{{ proxy_https_proxy }}"
  no_proxy: "{{ proxy_no_proxy }}"
```

(TODO: consider if we should just set these in common inventory to facilitate these?

Why the appliance can’t run without internet access:

* While most repos point to Ark and are disabled, some are not disabled (cernvm, ondemand-web, TurboVNC, epel-cisco-openh264, epel.rpmnew), hence *any* dnf tasks require internet access as it will try to update the cache for those repos.
* For in-appliance roles, the tasks running dnf can generally be skipped using an appliances_mode flag, where the packages are present in the image. But for “external” roles this is not always possible if the dnf tasks are not broken out into.
* Some tasks do actually require internet access to install things not present in the image, generally either dnf packages or pypi packages.

Plays with

```yaml
environment: "{{ appliances_environment_vars }}"
```

are the ones currently requiring internet access.


sjpb:
Confirmed at above revision that:
- hpctests run
- ondemand dashboard works

PR #681, opened 2025-05-27
Update openhpc role to allow removing slurm.conf parameters

sjpb:
Enables setting 'omit' as a value in `openhpc_config` to remove a default parameter from slurm.conf entirely.


sjpb:
Replaced by #690

PR #689, opened 2025-05-28
Remove description of un-implemented dummy interface/default route

### LINKED ISSUE #645: network dummy gateway description is not right
docs/networks.md includes this:
> If there is no default route at all (either from a subnet gateway or from gateway_ip) then a dummy route is created via the access network interface to ensure [correct](https://docs.k3s.io/installation/airgap#default-network-route) k3s operation.

but that didn't make it into the finalised PR.

sjpb:
Fixes #645

### DIFF
--- a/docs/networks.md
+++ b/docs/networks.md
@@ -157,11 +157,6 @@ compute = {
 }
 ```
 
-If there is no default route at all (either from a subnet gateway or from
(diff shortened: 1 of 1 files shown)


m-bull (approved):
LGTM

PR #691, opened 2025-05-28
Pin nvidia driver version

sjpb:
TODO

### DIFF
--- a/ansible/roles/cuda/tasks/install.yml
+++ b/ansible/roles/cuda/tasks/install.yml
@@ -32,26 +32,22 @@
 - name: Read module info for list of packages
   # Even a single version of the module stream contains multiple package versions
   # so need to find the ones we want.
-  # The dnf .. command returns a list of lines like
-  # '                 : libnvidia-cfg-3:570.133.20-1.el9.x86_64'
-  # Some are duplicated for arch x86_64/.i684
-  # Some don't have arch or distro, or have .noarch
-  # so want to get full package name without any suffixes so we can use
-  # {{ | unique }} later to get the actual packages to dnf install.
-  # The literal dot after the vesion is important so we don't get
-  # e.g. -10. when we want -1.
-  # So from the example line above we extract
-  # ' libnvidia-cfg-3:570.133.20-1.'
-  ansible.builtin.shell:
-    cmd: >-
-      dnf module info nvidia-driver:{{ cuda_nvidia_driver_stream }} |
-      grep -o '[^:]*:{{ cuda_nvidia_driver_version }}\.'
-  changed_when: false
-  register: _cuda_driver_module_packages
-  
-- name: Install nvidia driver packages
+  dnf_module_info:
+    name: nvidia-driver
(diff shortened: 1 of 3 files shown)


ansible/roles/cuda/tasks/install.yml
```diff
@@ -29,17 +29,22 @@
   when: "'No matching Modules to list' in _cuda_driver_module_enabled.stderr"
   changed_when: "'Nothing to do' not in _cuda_driver_module_enable.stdout"
 
-- name: Check if nvidia driver module is installed
-  ansible.builtin.command: dnf module list --installed nvidia-driver
+- name: Read module info for list of packages
+  ansible.builtin.shell:
+    cmd: >-
+      dnf module info nvidia-driver:{{ cuda_nvidia_driver_stream }} |
+      grep -F {{ cuda_nvidia_driver_version }}.el{{ ansible_distribution_major_version }}.{{ ansible_architecture }}
```
jovial, line 36: I think some packages don't have el9 suffix (distro independent) so does that mean this isn't a complete list?

sjpb: oh yeah like `nvidia-imex-570-0:570.124.06-1.x86_64`

So `dnf module info nvidia-driver:570-open` only appears to show packages ending .el9.x86_64, `el9.noarch` and `.x86_64`. So maybe its enough to just suffix the version with a `.` when grepping - I was trying to avoid the case where you are after e.g. `570-0:570.124.06-1` and just grepping for that also gets you `570-0:570.124.06-10`.

What do you think?

sjpb: Oh good spot, I'd missed those :facepalm:

jovial: So might need to filter for x86_64 and noarch on top of that you suggested? :thinking:

sjpb: I guess we can just `| unique` the list?

sjpb: Doh no grep just version+ . obvs returns the entire package name

jovial: or the other option is to just install the i686 packages too (not sure how much bloat that adds)

sjpb: Hmm, even with this we get:
```
Depsolve Error occurred: 
 Problem 1: conflicting requests
  - nothing provides cuda-drivers-570 = 570.133.20 needed by cuda-drivers-fabricmanager-570-570.133.20-1.x86_64 from cuda-rhel9-x86_64
 Problem 2: package cuda-drivers-fabricmanager-570.133.20-1.x86_64 from cuda-rhel9-x86_64 requires cuda-drivers-fabricmanager-570 = 570.133.20, but none of the providers can be installed
  - conflicting requests
  - nothing provides cuda-drivers-570 = 570.133.20 needed by cuda-drivers-fabricmanager-570-570.133.20-1.x86_64 from cuda-rhel9-x86_64
```

PR #693, opened 2025-05-29
Allow specifying instance root volume type

sjpb:
Adds new OpenTofu variable `root_volume_type` and overrides for compute and login node groups.

Useful on Leafcloud for testing large "extra build" images, where default volume type is encrypted.

### DIFF
--- a/environments/skeleton/{{cookiecutter.environment}}/tofu/variables.tf
+++ b/environments/skeleton/{{cookiecutter.environment}}/tofu/variables.tf
@@ -228,6 +228,12 @@ variable "root_volume_size" {
     default = 40
 }
 
+variable "root_volume_type" {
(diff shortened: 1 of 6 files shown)


environments/skeleton/{{cookiecutter.environment}}/tofu/variables.tf
```diff
@@ -228,6 +228,12 @@ variable "root_volume_size" {
     default = 40
 }
 
+variable "root_volume_type" {
+    description = "Type of root volume, if using volume backed instances. Default gives default volume type"
```
m-bull, line 232:
```suggestion
    description = "Type of root volume, if using volume backed instances. If unset, the target cloud default volume type is used."
```

m-bull (changes requested):
Should this be set in Leafcloud Packer vars as part of this PR?

PR #698, opened 2025-06-05
Change host definition for cacert play to allow builder

technowhizz:
Allows the cacerts role to be run on the builder group.


sjpb (approved):
LGTM.

PR #700, opened 2025-06-05
Ensure Ansible Galaxy installs are up to date

sjpb:
Feedback from users and developers shows it is very easy to forget to run `dev/setup-env.sh` to update Ansible Galaxy roles/collections after e.g. changing branches or merging in updates which change `requirements.yml`. The error messages occuring in this situation do not usually make the root cause clear.

While this cannot be actually fixed from `ansible/site.yml` as roles/collections are already loaded once Ansible is running, this PR raises a clear error if the installed roles/collections do not match the current `requirements.yml`

Adds a new tag `galaxy` which can be used to skip this check e.g. during role development.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ venv
 *.pyc
 packer/openhpc2
 .vscode
+requirements.yml.last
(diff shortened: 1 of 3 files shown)


MoteHue (approved):
LGTM, thanks :)

PR #712, opened 2025-06-23
Bump OFED version to latest 24.10 release

priteau:
The 23.10 series does not officially support Rocky Linux 9.5.


sjpb (changes requested):
~This needs to go thro the extra build CI. The fact that isn't triggered by changes to the doca role (and the lustre and cuda roles) is TBH a bug.~


All wrong, this is fine :facepalm:

PR #713, opened 2025-06-23
Add s-nail package

priteau:
This includes the /usr/bin/mail command, which is required for Slurm to send email notifications [1].

[1] https://slurm.schedmd.com/slurm.conf.html#OPT_MailProg

### DIFF
--- a/environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
+++ b/environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
@@ -1,6 +1,6 @@
 {
     "cluster_image": {
-        "RL8": "openhpc-RL8-250704-1445-ff88ca4e",
-        "RL9": "openhpc-RL9-250704-1445-ff88ca4e"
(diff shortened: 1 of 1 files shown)


priteau:
> Is the package alone enough to be useful? I was expecting this was e.g. to send emails on job status and therefore to see e.g. [Mail*](https://slurm.schedmd.com/slurm.conf.html#OPT_MailDomain) parameters configured in slurm.conf configured?

There is actually no other configuration to perform in Slurm itself. Of course, you need a functional Postfix SMTP relay, but its configuration is mostly site-specific. I was planning to push some documentation separately, as I still want to check if we should import an existing Ansible role for Postfix configuration.

priteau:
To be more precise, this is because the default value of `MailProg` is already `/bin/mail` (or `/usr/bin/mail` if `/bin/mail` does not exist but `/usr/bin/mail` does exist), as documented in https://slurm.schedmd.com/slurm.conf.html.

priteau:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/16146871938

PR #727, opened 2025-07-03
Define login subgroups in Ansible inventory

priteau:


environments/skeleton/{{cookiecutter.environment}}/tofu/inventory.tpl
```diff
@@ -35,6 +35,12 @@ login:
         ${cluster_name}_${group_name}:
 %{ endfor ~}
 
+%{ for group_name in keys(login_groups) ~}
```
sjpb, line 38: Can we not make this consistent with way this is done for compute, and hoist the group defn inside the loop from line 19?

Or else make compute consistent with this, I don't care, but I think them being different is mentally hard to work out if they are doing something different. I don't think they are...

sjpb (changes requested):

PR #733, opened 2025-07-11
Output baremetal instances from terraform

wtripp180901:
Added templating to hosts file to include a list of compute instances which are mapped to Ironic nodes. The motivation for this was to be able to distinguish between VMs and baremetal instances for inter-host topology aware scheduling (`host_id` refers unique hypervisors in the VM case, but can just refer to Ironic service in baremetal case) but we've decided to exclude baremetal instances in the initial iteration of this


sjpb:
@wtripp180901 do we want to close this, for the moment??

PR #717, opened 2025-06-24
Support clusters with no outbound internet

sjpb:
Allows clusters to be deployed without outbound internet access. Key configuration changes:
- Some proxy  variables are now defined for convenience in `environments/common/inventory/group_vars/all/proxy.yml`, making it easier to define a proxy using basic authentication.
- New variable `proxy_remove` allows removing proxy configuration at the end of Ansible runs. This defaults to `false` for clusters but `true` for build VMs. NB: This is a change in behaviour; before if the `proxy` group was enabled for the build then the proxy configuration would have been left in the image.
- New variable `openondemand_install_apps_extra` to simplify installing additional OpenOndemand Apps.

See docs/experimental/isolated-clusters.md for more information.

--- 

Design notes:

- All repos are now disabled at the end of both fatimage.yml site.yml playbooks via `dnf_repos:disable_repos.yml`. This means `ansible.builtin.dnf` tasks during site.yml do not trigger dnf cache updates and hence complete successfully (assuming relevant package(s) are actually installed).
- For supported features, installation tasks have been moved into `fatimage.yml`, with additional guards for `appliances_mode` added to avoid running them during site.yml.

PR #735, opened 2025-07-11
Fix convenience variables for proxy

sjpb:
PR #717 added some convenience proxy vars, but the logic wasn't quite correct. This:
- Makes it so simply defining `proxy_http_address` is enough to produce a correct `proxy_http_proxy` variable (which in turn is all that is required to produce a working proxy config if the other defaults are sufficent).
- Fixes a whitespace issue.


bertiethorpe (approved):
LGTM

PR #737, opened 2025-07-14
Add support for topology-aware scheduling

wtripp180901:
Enables Slurm's topology aware scheduling plugin for virtualised clusters. A topology tree is automatically created with four levels as follows (from the bottom up):
- Compute VMs (leafs)
- Hypervisors (switches)
- Availability zones - used to model server racks (switches)
- Top-level inter-rack switch

Note that the OpenTofu node group parameter `availability_zone` can now also be used for VMs, not just node groups with `match_ironic_node` set.

### DIFF
--- a/ansible/roles/topology/README.md
+++ b/ansible/roles/topology/README.md
@@ -1,16 +1,21 @@
 topology
 ========
 
-Templates out /etc/slurm/topology.conf file based on an Openstack project for use by
+Templates out /etc/slurm/topology.conf file based on an OpenStack project for use by
 Slurm's [topology/tree plugin.](https://slurm.schedmd.com/topology.html) Models
-cluster as tree with a heirarchy of:
+cluster as tree with a hierarchy of:
 
 Top-level inter-rack Switch -> Availability Zones -> Hypervisors -> VMs
+
+Warning: This role doesn't currently trigger a restart of Slurm so will therefore not
+reconfigure an already running cluster after a `ansible/site.yml` run. You will therefore need
+to run the `ansible/adhoc/restart-slurm.yml` playbook for changes to topology.conf to be
+recognised.
 
 Role Variables
 --------------
 
-- `topology_topology_nodes:`: Required list of strs. List of inventory hostnames of nodes to include in topology tree. Must be set to include all compute nodes in Slurm cluster. Default `[]`.
+- `topology_nodes:`: Required list of strs. List of inventory hostnames of nodes to include in topology tree. Must be set to include all compute nodes in Slurm cluster. Default `[]`.
 - `topology_conf_template`: Optional str. Path to Jinja2 template of topology.conf file. Default
   `templates/topology.conf.j2`
 - `topology_above_rack_topology`: Optionally multiline str. Used to define topology above racks/AZs if 
--- a/ansible/roles/topology/README.md
+++ b/ansible/roles/topology/README.md
@@ -13,3 +13,16 @@ Role Variables
 - `topology_topology_nodes:`: Required list of strs. List of inventory hostnames of nodes to include in topology tree. Must be set to include all compute nodes in Slurm cluster. Default `[]`.
 - `topology_conf_template`: Optional str. Path to Jinja2 template of topology.conf file. Default
   `templates/topology.conf.j2`
+- `topology_above_rack_topology`: Optionally multiline str. Used to define topology above racks/AZs if 
+   you wish to partition racks further under different logical switches. New switches above should be
+   defined as [SwitchName lines](https://slurm.schedmd.com/topology.html#hierarchical) referencing 
+   rack Availability Zones under that switch in their `Switches fields`. These switches must themselves
+   be under a top level switch. e.g
+   ```
+   topology_above_rack_topology: |
+     SwitchName=rack-group-1 Switches=rack-az-1,rack-az-2
+     SwitchName=rack-group-2 Switches=rack-az-3,rack-az-4
+     SwitchName=top-level Switches=rack-group-1,rack-group-2
+   ```
+   Defaults to an empty string, which causes all AZs to be put under a
+   single top level switch.
--- a/ansible/roles/topology/README.md
+++ b/ansible/roles/topology/README.md
@@ -3,12 +3,13 @@ topology
 
 Templates out /etc/slurm/topology.conf file based on an Openstack project for use by
 Slurm's [topology/tree plugin.](https://slurm.schedmd.com/topology.html) Models
-project as tree with a heirarchy of:
+cluster as tree with a heirarchy of:
 
-Project -> Availability Zones -> Hypervisors -> VMs
+Top-level inter-rack Switch -> Availability Zones -> Hypervisors -> VMs
 
 Role Variables
 --------------
 
-- `topology_topology_nodes: []`: Required list[str]. List of nodes to include in topology tree. Must be set to include all compute nodes in Slurm cluster. Default `[]`.
-- `topology_topology_override:`: Optional str. If set, will override templating and be provided as custom topology.conf content. Undefined by default.
+- `topology_topology_nodes:`: Required list of strs. List of inventory hostnames of nodes to include in topology tree. Must be set to include all compute nodes in Slurm cluster. Default `[]`.
+- `topology_conf_template`: Optional str. Path to Jinja2 template of topology.conf file. Default
+  `templates/topology.conf.j2`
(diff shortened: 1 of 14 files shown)


environments/common/layouts/everything
```diff
@@ -135,3 +135,6 @@ builder
 [nhc:children]
 # Hosts to configure for node health checks
 compute
+
+[topology:children]
```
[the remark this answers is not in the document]
sjpb: Oh sorry, basically as in common groups, but noting the above!

environments/common/inventory/group_vars/all/openhpc.yml
```diff
@@ -57,6 +57,7 @@ openhpc_config_default:
     - enable_configless
   TaskPlugin: task/cgroup,task/affinity
   ReturnToService: 2 # workaround for templating bug TODO: Remove once on stackhpc.openhpc v1.2.0
+  TopologyPlugin: topology/tree
```
sjpb, line 60: What happens if you have nothing in `topology` group and this is enabled? Does it need to be guarded/nulled out somehow if that is the case, or is it OK?

environments/common/layouts/everything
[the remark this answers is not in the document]
sjpb: In general I think common and everything should be identical in terms of comments.

ansible/slurm.yml
```diff
@@ -61,6 +61,11 @@
   tags:
     - openhpc
   tasks:
+    - include_role:
+        name: topology
+      # Gated on topology group having compute nodes but role also
+      # needs to run on control and login nodes
```
sjpb, line 67: Thanks, nice to have that clarified here.

ansible/slurm.yml
```diff
@@ -61,6 +61,11 @@
...
+      when: appliances_mode == 'configure' and (groups['topology'] | length) > 0
```
sjpb, line 68: Just a style thing, can we do:
```suggestion
      when:
        - appliances_mode == 'configure'
        - groups['topology'] | length > 0
```

ansible/roles/topology/README.md
```diff
@@ -0,0 +1,14 @@
+topology
+========
+
+Templates out /etc/slurm/topology.conf file based on an Openstack project for use by
+Slurm's [topology/tree plugin.](https://slurm.schedmd.com/topology.html) Models
+project as tree with a heirarchy of:
+
+Project -> Availability Zones -> Hypervisors -> VMs
```
sjpb, line 8: Can we modify this iaw the discussion re the PR description please?

ansible/roles/topology/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+# Nodes to be included in topology tree, must include all Slurm compute nodes
+topology_topology_nodes: []
+
+# If set, will override topology.conf file auto-detected from OpenStack project
+# topology_topology_override:
```
sjpb, line 5: Nah looking at the templating this is the wrong approach. Instead of having a massive if/else in the actual template just let users voerride the template entirely, i.e. replace this var with `topology_conf_template: topology.conf.j2`.

(I think actually in general we should ALWAYS allow users to override templates, gets you out of a lot of places you otherwise need to fork roles etc.)

ansible/roles/topology/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+Role Variables
+--------------
+
+- `topology_topology_nodes: []`: Required list[str]. List of nodes to include in topology tree. Must be set to include all compute nodes in Slurm cluster. Default `[]`.
```
sjpb, line 13:
```suggestion
- `topology_topology_nodes: []`: Required list of strs. List of nodes to include in topology tree. Must be set to include all compute nodes in Slurm cluster. Default `[]`.
```

We don't really use that formal typing anywhere else (maybe we should!)

sjpb: And, this is actually a list of "inventory hostnames" right? I'm fine with the role var name but if I'm right we shold definitlely clarify that.

ansible/roles/topology/templates/topology.conf.j2
```diff
@@ -0,0 +1,13 @@
+# topology.conf
+# Switch Configuration
+{% if topology_topology_override is defined %}
```
sjpb, line 3: See README comment

ansible/roles/topology/templates/topology.conf.j2
```diff
@@ -0,0 +1,13 @@
...
+{{ topology_topology_override }}
+{% else %}
+{% for az in _topology.topology.keys() %}
+{% for instance_host in _topology.topology[az].keys() %}
+SwitchName={{ instance_host }} Nodes={{ _topology.topology[az][instance_host] | join(",") }}
+{% endfor %}
+SwitchName={{ az }} Switches={{ _topology.topology[az].keys() | join(",") }}
+{% endfor %}
+SwitchName=master Switches={{ _topology.topology.keys() | join(",") }}
```
sjpb, line 12: I'd really like to have something so you can override the top-level definition entirely, e.g. if you know you actually have network structure ABOVE the AZs, rather than just a single top-level switch, you can define this yourself. I think it is as simple as having something like:

```markdown
# README.md
- topology_topology_top_level: Optional. A possibly-multiline string giving [SwitchName](https://slurm.schedmd.com/topology.html#hierarchical) lines defining the network topology above the racks/Availability Zones. Each Availbility Zone should be referned as a Switch. The default of an empty string defines a single Switch with all Availability Zones as children
```

except the wording is probaly not quite right/consistent with what you've written elsewhere...

ansible/roles/topology/library/map_hosts.py
```diff
@@ -0,0 +1,98 @@
...
+from ansible.module_utils.basic import AnsibleModule
+import openstack
+
+DOCUMENTATION = """
+---
+module: map_hosts
+short_description: Creates map of OpenStack VM network topology
+description:
+    - Creates map representing the network topology tree of an OpenStack project with a heirarchy
+      of: Availability Zone -> Hypervisors/Baremetal nodes -> VMs/Baremetal instances
```
sjpb, line 15: Does it actually handle BM nodes??

ansible/roles/topology/library/map_hosts.py
```diff
@@ -0,0 +1,98 @@
...
+    conn = openstack.connection.from_config()
+
+    servers = [s for s in conn.compute.servers() if s["name"] in module.params["compute_vms"]]
+
+    topo = {}
+    all_host_ids = []
+    for s in servers:
+        az = s['availability_zone']
+        host_id = s['host_id']
+        if host_id != '':
```
sjpb, line 72: Can you add a comment on which circumstances the host_id *is* the empty string please?

ansible/roles/topology/library/map_hosts.py
```diff
@@ -0,0 +1,98 @@
...
+
+    for az in topo:
+        topo[az] = dict((k[:uuid_len], v) for (k, v) in topo[az].items())
+
+    result = {
+        "changed": True, 
+        "topology": topo,
+    }
+    
+    module.exit_json(**result)
```
sjpb, line 90: I don't really care if we support check mode or not (I think!) b/c lots of other things in the appliance don't, but we should set the flag to say if we do or not.

wtripp180901: This is set at module creation https://github.com/stackhpc/ansible-slurm-appliance/pull/737/files#diff-3d29407cf3e33d58390bf5d895a8084b88b8e7bd4b648f8abb7703ce7c3e5ff0R61 I think this supports check mode given that it makes no changes to any hosts, but will always show up as changed given that we don't store the previous state

ansible/slurm.yml
```diff
@@ -61,6 +61,11 @@
   tags:
     - openhpc
   tasks:
+    - include_role:
```
sjpb, line 64: Without any integration with the openhpc role, this can't actually trigger a slurm reconfigure/restart when the topology.conf file changes, is that correct? I'm kind of OK with that at least for now, but can we at least document that, given that maintenance might well move these around ...

sjpb (changes requested):

ansible/roles/topology/library/map_hosts.py
sjpb:
```suggestion
        if host_id != '': # empty string if e.g. server is shelved
```

sjpb (changes requested):

wtripp180901:
^ added suggestion manually so I didn't have to cancel the CI again!

PR #745, opened 2025-07-31
Add metadata service to no_proxy defaults

sjpb:
When using a proxy, normally requests to the metadata service should not go through the proxy.


JohnGarbutt (approved):
oh, good catch.

PR #704, opened 2025-06-11
Support additional nodegroups

sjpb:
Adds opentofu variable `additional_nodegroups` to support defining non-Slurm nodes in the cluster. E.g.:

```terraform
additional_nodegroups = {
    squid = {
        nodes = ["squid-0"]
        flavor = var.other_node_flavor
    }
}
```

Nodes are automatically added to an inventory group of the same name as the node group.

PR #749, opened 2025-08-05
Fixup additional_nodes inventory group

sjpb:
#704 added support for additional node groups. This PR fixes these missing from the `cluster` nodegroup. Note they are not part of the `openhpc` group as they do not run slurm daemons.


bertiethorpe (approved):
LGTM

PR #753, opened 2025-08-07
Support creating EESSI CernVM-FS Stratum 1 server

sjpb:
Intended for use in some isolated environments. See `ansible/roles/cvmfs_server`.


sjpb:
Note there's some info on required stratum 1 size and growth here:
- https://github.com/EESSI/cvmfs-status-page-rust/issues/7

PR #763, opened 2025-08-22
Enable persistent journaling

jovial:
In the cloud images, /var/log/journal doesn't exist. When this directory does not exist, journald will not write any files to disk, and as such, logging data is lost accross reboots.

Persistent logging is useful for debugging issues that caused the nodes to reboot (provided that the logs are flushed to disk).

### DIFF
--- a/environments/.stackhpc/tofu/main.tf
+++ b/environments/.stackhpc/tofu/main.tf
@@ -85,7 +85,7 @@ module "cluster" {
     standard = { # NB: can't call this default!
       nodes                = ["compute-0", "compute-1"]
       flavor               = var.other_node_flavor
-      compute_init_enable  = ["compute", "chrony", "etc_hosts", "nfs", "basic_users", "eessi", "tuned", "cacerts", "nhc", "mounts"]
+      compute_init_enable  = ["compute", "chrony", "etc_hosts", "nfs", "basic_users", "eessi", "tuned", "cacerts", "nhc", "mounts", "journald"]
       ignore_image_changes = true
     }
     # Normally-empty partition for testing:


environments/common/inventory/groups
```diff
@@ -26,6 +26,10 @@ login
 openhpc
 additional
 
+[journald:childen]
```
sjpb, line 29: If its non-empty here, rather than environments/site/inventory/groups, it can NEVER be turned off. Is that OK?

ansible/bootstrap.yml
sjpb, line 1: Does it need compute-init support?? Or, do we get "compute-init" support "for free" when using stackhpc images (b/c the path/drop-ins etc have been created correctly during build, as that runs bootstrap.yml). Either way, ansible/roles/compute_init/README.md should be updated.

sjpb (changes requested):
- pending compute-init comment
- also needs referencing in docs/experimental/isolated-networking.md

environments/common/inventory/groups
jovial: I wouldn't expect anyone to want to turn this off. You can still tune journald_storage to volatile, but I'm not sure why you'd want to do that.

ansible/bootstrap.yml
jovial: > Does it need compute-init support?? Or, do we get "compute-init" support "for free" when using stackhpc images (b/c the path/drop-ins etc have been created correctly during build, as that runs bootstrap.yml). Either way, ansible/roles/compute_init/README.md should be updated.

Yeah, I believe it comes from "free", but I'll update `ansible/roles/compute_init/README.md`

environments/common/inventory/groups
jovial: I've moved it to environments/site as that sounded preferable

environments/site/inventory/groups
```diff
@@ -81,6 +81,10 @@ cluster
 # Hosts to recompile Slurm for - allows supporting Slurm autodetection method 'nvml'
 cuda
 
+[journald:childen]
```
jovial, line 84:
```suggestion
[journald:children]
```

ansible/roles/compute_init/README.md
```diff
@@ -41,6 +41,7 @@ it also requires an image build with the role name added to the
 | hooks/pre.yml            | ?                             | None at present                 | n/a                 |
 | validate.yml             | n/a                           | Not relevant during boot        | n/a                 |
 | bootstrap.yml            | (wait for ansible-init)       | Not relevant during boot        | n/a                 |
+| bootstrap.yml            | journald                      | Fully supported                 | No                  |
```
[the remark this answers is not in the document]
jovial: Easiest thing to do is to make it run on boot too. It looks like I'm missing the handlers!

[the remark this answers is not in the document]
sjpb: Well ... for that you'd need the whole compute-init setup (I've tried to document the pattern you arrived at [here](https://wiki.stackhpc.com/doc/slurm-development-ZXjBRByl6K#h-patterns-for-compute-init). I don't really see why people are going to want to change this, so TBH I would just add the note to this docs, saying the default is supported, changing it requires building a new image with the default changed. I think that's fine.

sjpb (changes requested):
Looks good, other than unnecessary changes to builder.yml defaults?

sjpb (approved):

Issue #722, opened 2025-06-26
confusing caas ssh args


sjpb:
https://github.com/stackhpc/ansible-slurm-appliance/blob/main/environments/.caas/ansible.cfg#L14 is missing a `-o` in front of ControlPath. However this file isn't actually used at all - it is the one in the repo root which is used for caas.

failure when using that one:

```
==> openstack.openhpc: Data could not be sent to remote host "128.232.227.159". Make sure this host can be reached over ssh: ssh: Could not resolve hostname %h-%p: Name or service not known
```

PR #766, opened 2025-08-28
Remove ansible.cfg from CaaS environment

bertiethorpe:
Removes ansible.cfg from .caas environment. It is not used.

Addresses https://github.com/stackhpc/ansible-slurm-appliance/issues/722


sjpb (approved):
LGTM

PR #767, opened 2025-08-29
CaaS pre-hook fix for galaxy requirements validation

bertiethorpe:
CaaS doesn't use setup-env.sh which, since https://github.com/stackhpc/ansible-slurm-appliance/pull/700, is used to validate galaxy requirements.yml install.

Workaround uses pre-hook to copy requirements.yml to requirements.yml.last so this galaxy validate step doesn't error.


bertiethorpe:
CI cancelled because not tested there

sjpb (approved):
LGTM, thanks

PR #769, opened 2025-09-04
Update openhpc role to support custom slurm binaries

sjpb:
See https://github.com/stackhpc/ansible-role-openhpc/pull/163


sjpb:
Replaced by #811

PR #773, opened 2025-09-05
Add support for setting server group in scheduler hints

sjpb:
Adds new OpenTofu variable `control_server_group_id` and node group parameter `server_group_id` to allow specifying a server group to set for scheduler hints. This allows controlling affinity/anti-affinity placement of cluster nodes on hypervisors.

If wishing to set these IDs from a data.openstack_compute_servergroup_v2 resource, note this was only added in v3.2 of the Terraform OpenStack provider, which is more recent than the current minimum requirement.


MoteHue (approved):
Looks good, thanks @sjpb

jovial (approved):

PR #774, opened 2025-09-08
Make CaaS specific role: `persist_openhpc_secrets` idempotent

bertiethorpe:
CaaS env now inherits from both common and site envs

As a result of this we hit a problem where alertmanager is now turned on by default but there’s no secret for it. Also secret is added to `persist_openhpc_secrets`.

Aligning CaaS to site also re-raises the issue of CaaS secrets not being idempotent. This is resolved by:

- Fixing upgrade by changing the persist_openhpc_secrets role to read the facts file if it exists
- template it out again, with the template being changed to be idempotent for values a la the approach in https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/roles/passwords/defaults/main.yml
- then read it again (getting any new passwords)

### DIFF
--- a/ansible/roles/persist_openhpc_secrets/tasks/main.yml
+++ b/ansible/roles/persist_openhpc_secrets/tasks/main.yml
@@ -14,14 +14,18 @@
   loop:
     - "{{ appliances_state_dir }}/ansible.facts.d"
     - "/etc/ansible/facts.d"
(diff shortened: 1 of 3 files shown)


bertiethorpe:
Tested fully now so going to go ahead and merge

PR #666, opened 2025-05-08
Update appliance for stackhpc.openhpc nodegroup/partition changes

sjpb:
Update appliance to cope with https://github.com/stackhpc/ansible-role-openhpc/pull/183 which changed `openhpc_slurm_partitions` to `openhpc_nodegroups` and `openhpc_partitions`.

As an example,  a configuration like:
```yaml
#environments/site/inventory/group_vars/all/openhpc.yml:
openhpc_slurm_partitions:
  - name: standard
    default: 'YES'
    groups:
       - name: largemem
       - name: gpu
         gres:
          - conf: gpu:A40:2
            file: /dev/nvidia[0-1]
```
would now be represented as:
```yaml
#environments/site/inventory/group_vars/all/openhpc.yml:
openhpc_nodegroups:
  - name: gpu
      gres:
        - conf: gpu:A40:2
          file: /dev/nvidia[0-1]
openhpc_partitions:
  - name: standard
    default: 'YES'
    nodegroups:
      - largemem
      - gpu
```

Note that:
- Arbitrary [slurm.conf Node-level parameters](https://slurm.schedmd.com/slurm.conf.html#lbAE) can now be included via the optional `node_params` entry on `openhpc_nodegroup` elements.
- The skeleton OpenTofu configuration no longer templates out partition configuration variables. Instead OpenTofu templates an Ansible variable `cluster_compute_groups` variable into the `inventory/hosts.yml` file. **NB**: This file should be removed once OpenTofu configurations have been updated.
- A default `openhpc_nodegroups` is calculated in `environments/common/inventory/group_vars/all/openhpc.yml`. This uses the above to creates a node group per key in the OpenTofu `compute ` variable. This is usually what is required, but this variable will need overriding if custom Node-level Slurm parameters/configuration is required (e.g. for GRES).
- A default `openhpc_partitions` is calculated. This includes one partition per node group, plus a rebuild partition if the `rebuild` group is active. This should not generally be overriden. Instead normal (non-rebuild) partitions should be modified using `openhpc_user_partitions`, which is the same format.

This affects:
- [x] common environment openhpc config
- [x] skeleton templating
- [x] caas environment infra creation and ansible variables
- [x] stackhpc environment openhpc overides
- [x] configuration required to support rebuild - now automatic
- [x] can now call stackhpc.openhpc:validate.yml from ansible/validate.yml

#665 is replaced by this.

PR #775, opened 2025-09-09
Add validation for tofu-templated vars

sjpb:
Will help catch errors if an Ansible update introduces a new variable provided by OpenTofu (e.g. when #666 added `compute_groups`) but the OpenTofu configurations haven't been correctly upgraded. Currently, this just leads to a templating error at some point during site.yml which is hard to debug.

This also "documents" the expected interface from OpenTofu to Ansible.

### DIFF
--- a/ansible/validate.yml
+++ b/ansible/validate.yml
@@ -57,7 +57,10 @@
         that:
           - item in groups
           - groups[item] | length > 0
-        fail_msg: "Expected inventory group '{{ item }}' is missing or empty: is OpenTofu inventory template up to date?"
+        fail_msg: >
+          Expected inventory group '{{ item }}' is missing or empty:
+          - Check OpenTofu inventory template is up to date
+          - Check OpenTofu configuration defines 'login' and 'compute' variables properly
       loop:
         - control
         - compute


ansible/validate.yml
```diff
@@ -43,6 +43,34 @@
...
+  tags:
+    - validate
+    - opentofu
+  tasks:
+    - name: Check templated groups
+      assert:
+        that:
+          - item in groups
+          - groups[item] | length > 0
+        fail_msg: "Expected inventory group '{{ item }}' is missing or empty: is OpenTofu inventory template up to date?"
```
wtripp180901, line 60: Thinking about it, couldn't this also indicate that you've misconfigured the compute or login groups so that they're empty? We would also want to fail on that, but might be worth updating message

ansible/validate.yml
```diff
@@ -43,6 +43,34 @@
...
+        - control
+        - compute
+        - login
+    - name: Check templated 'all' vars
+      assert:
+        that:
+          - openhpc_cluster_name is defined
+          - cluster_domain_suffix is defined
+          - cluster_home_volume is defined
+          - cluster_compute_groups is defined
```
wtripp180901, line 71: do we also want to check for
```suggestion
          - cluster_compute_groups is defined
          - hostvars[groups['control'] | first].appliances_state_dir is defined
```

wtripp180901 (changes requested):
looks good, just double checking for cases we might miss

ansible/validate.yml
sjpb: Well, theoretically its OK to run without that, in that the ansible copes ... but I agree it isn't sensible. I think I want to say that is out of scope for the moment!

PR #778, opened 2025-09-10
Ignore changes to port binding and dhcp options

sjpb:
Ignores changes to the following for port lifecycle to fix issues with re-running OpenTofu:
- `binding`: When running with OpenStack admin privileges, binding information gets returned
- `extra_dhcp_option`: The Neutron networking-mlnx plugin adds client-id to the dhcp options

The lifecycle meta-argument cannot be parameterised ([issue](https://github.com/opentofu/opentofu/issues/2525)) but these are considered safe to always set.

### DIFF
--- a/environments/site/tofu/control.tf
+++ b/environments/site/tofu/control.tf
@@ -37,7 +37,7 @@ resource "openstack_networking_port_v2" "control" {
 
   lifecycle {
     ignore_changes = [
-      binding, # fixes running as admin
+      binding,          # fixes running as admin
       extra_dhcp_option # required for networking-mlnx neutron plugin
     ]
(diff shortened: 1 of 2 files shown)


environments/site/tofu/control.tf
```diff
@@ -34,6 +34,10 @@ resource "openstack_networking_port_v2" "control" {
   binding {
     vnic_type = lookup(var.vnic_types, each.key, "normal")
   }
+
+  lifecycle {
+    ignore_changes = [ binding ]
```
priteau, line 39: If you have to update it, please change to `[binding]` as per `tofu fmt`.

PR #783, opened 2025-09-12
Use source repos from Ark for openhpc package rebuild

bertiethorpe:


sjpb (changes requested):
I think the source timestamps need to match the non-source ones. Else we're getting source packages which don't match the dnf ones.

Also, it at least needs an extrabuild to test this ...

sjpb:
I checked the slurm_recompile role and it doesn't add repos or anything, so changing `dnf_repo_timestamps.yml` should be sufficent.

PR #785, opened 2025-09-16
Support software raid root disks in stackhpc images

sjpb:
See ansible/roles/raid/README.md


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/18012807049

sjpb:
Carried out same checks on "Enable autoassembly of mdraid devices" task and handler as above, OK

PR #786, opened 2025-09-17
Expose vgpu group in site inventory

priteau:
Also fix various comments in Ansible group files.


sjpb (changes requested):
Good catch. Can the `vgpu` group also be added to environments/common/groups please? Generally that file should be the "source of truth" for what groups exist.

Yes, there would be an argument for unduplicating common and site groups, but it is what it is ...

sjpb (approved):
LGTM

PR #792, opened 2025-09-19
Update dnf repo snapshots

sjpb:
Also uses source repos from Ark (used for openhpc package rebuilds).

Replaces #783


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/17858384753

PR #782, opened 2025-09-12
Bump Open OnDemand to v4 & install apps in fatimage

bertiethorpe:
- Upgrade Open OnDemand to v4 (from v3)
- Fixes fatimage not installing ondemand app packages
- Increases fatimage size to 20GB (from 15GB)

NB: The password for "code server app" sessions can be found using My Interactive Sessions >  Session ID > connection.yml

Open Ondemand v4 adds a cluster status page:
image

image

PR #794, opened 2025-09-19
Fix RStudio and Code Server apps

priteau:
They were both missing from the fat image build. Also fix a syntax error with the RStudio version.


sjpb:
Updated to OnDemand v4 in #782 instead.

PR #798, opened 2025-09-29
Fix .caas secrets not persisting post-reimage and skip tofu vars validation for .caas

wtripp180901:
Fixes bug where, after reimaging the cluster, persisted .caas secrets aren't symlinked to the ansible facts directory until after facts are loaded for writing secrets, causing secrets to be regenerated.

Also skips validation task for checking tofu templated vars exist. The vars it checks for aren't currently used in the .caas case:
- `cluster_domain_suffix` is used in `freeIPA` and `resolv_conf` roles, neither of which are active in .caas
- `cluster_home_volume` is set to true in the normal appliance if the `home_volume_provisioning` tofu var isn't set to `"none"`. The `home_volume_provisioning` var doesn't exist in the .caas tofu as the home volume is always tofu managed. NFS options controlled by `cluster_home_volume` default to true if it is unset, so this still works
- `cluster_compute_groups` gets templated into `openhpc_nodegroups` and `openhpc_rebuild_partition.nodegroups` in the normal appliance, but .caas templates `openhpc_nodegroups` with Azimuth values and `openhpc_rebuild_partition.nodegroups` isn't (currently) relevant to .caas
so it was opted to skip the validation task rather than adding misleading dummy values for those variables

### DIFF
--- a/ansible/roles/persist_openhpc_secrets/tasks/main.yml
+++ b/ansible/roles/persist_openhpc_secrets/tasks/main.yml
@@ -15,6 +15,14 @@
     - "{{ appliances_state_dir }}/ansible.facts.d"
     - "/etc/ansible/facts.d"
 
+- name: Symlink to persisted facts if present
+  ansible.builtin.file:
+    state: link
+    src: "{{ appliances_state_dir }}/ansible.facts.d/openhpc_secrets.fact"
+    dest: /etc/ansible/facts.d/openhpc_secrets.fact
+    owner: root
+  when: openhpc_secrets_stat.stat.exists
+
 - name: Load existing OpenHPC secrets if present
   ansible.builtin.setup:
     filter: ansible_local
--- a/ansible/validate.yml
+++ b/ansible/validate.yml
@@ -77,6 +77,7 @@
           - cluster_home_volume is defined
           - cluster_compute_groups is defined
         fail_msg: "One or more expected variables are missing: is OpenTofu inventory template up to date?"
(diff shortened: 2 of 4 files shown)


wtripp180901:
still needs caas testing

ansible/roles/persist_openhpc_secrets/tasks/main.yml
sjpb, line 38: @wtripp180901 do we need THIS symlink task as well as the one above? My reading of the https://github.com/stackhpc/ansible-slurm-appliance/pull/774/files diff is just that we added the "Load existing ..." task in the wrong place - but I might be missing something?

ansible/validate.yml
```diff
@@ -77,6 +77,7 @@
           - cluster_home_volume is defined
           - cluster_compute_groups is defined
         fail_msg: "One or more expected variables are missing: is OpenTofu inventory template up to date?"
+      when: not appliances_caas_skip_validate_vars
```
sjpb, line 80: I'd rather spell this like this (and default it to `true` in common), but don't bother if it means more work now.
```suggestion
      when: appliances_validate_tofu_vars | bool
```
(using bool for ease of CLI extravars overriding).

ansible/roles/persist_openhpc_secrets/tasks/main.yml
wtripp180901: it's still needed since the persistent file might not exist at the first symlink's creation

PR #805, opened 2025-10-01
Fix image sync workflow for new larger fat images

sjpb:
Since https://github.com/stackhpc/ansible-slurm-appliance/pull/782 merged increasing the image size to 20GB, the GitHub hosted runner has been out of space when converting the .raw to .qcow2.

This uses the `/mnt` directory which appears to be on a mostly-empty 74GB disk as the working directory to avoid that issue.


bertiethorpe (approved):
LGTM

PR #811, opened 2025-10-07
Bump OpenHPC role to v1.4.0

bertiethorpe:
Adds:
- Enable use of custom Slurm builds by @sjpb in https://github.com/stackhpc/ansible-role-openhpc/pull/163
- CI: Switch to latest rockylinux/rockylinux images by @priteau in https://github.com/stackhpc/ansible-role-openhpc/pull/198
- Add support for mpi.conf templating by @bertiethorpe in https://github.com/stackhpc/ansible-role-openhpc/pull/201

### DIFF
--- a/environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
+++ b/environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
@@ -1,6 +1,6 @@
 {
   "cluster_image": {
-    "RL8": "openhpc-RL8-251002-1537-1d21952c",
-    "RL9": "openhpc-RL9-251002-1456-1d21952c"
+    "RL8": "openhpc-RL8-251008-0814-cda7084d",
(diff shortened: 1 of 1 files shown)


ansible/fatimage.yml
```diff
@@ -117,7 +117,7 @@
     - name: Install OpenHPC
       ansible.builtin.import_role:
         name: stackhpc.openhpc
-        tasks_from: install.yml
+        tasks_from: install-ohpc.yml
```
sjpb, line 120: So we need a full image build here then, so we are dragging in the right file.

sjpb (changes requested):

bertiethorpe:
Fatimage CI build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/18338215318/job/52227047294

PR #827, opened 2025-10-15
Add docs for eessi

claudia-lola:


docs/eessi.md
```diff
@@ -0,0 +1,63 @@
+# EESSI Guide
+
+## How to load EESSI
+Loading an EESSI environment module:
```
sjpb: I think that:
```suggestion
Loading an EESSI environment module:
```
a) It needs to make clear these are alternatives, so something like "EESSI can be initialised using either ..."
b) the eessi module approach is marked as preferred in the docs, BUT there's the expandable section  "what if I already have Lmod installed ..." - which we do. So I think we need the bits from 1. in there too.

docs/eessi.md
```diff
@@ -0,0 +1,63 @@
...
+source /cvmfs/software.eessi.io/versions/2023.06/init/lmod/bash
+```
+Activating EESSI environment:
+```[bash]
+source /cvmfs/software.eessi.io/versions/2023.06/init/bash
+```
+
+## GPU Support with EESI
+To enable GPU support you need a site-specific build that has `cuda` enabled. For guide to do this please refer to `docs/image-build`. This is
+because CUDA-drivers are host specific and EESSI can not ship NVIDIA drivers due to licensing + kernel specific constraints. This means that the
```
sjpb: typo "can not" -> "cannot" but also; they could ship the open drivers, but they don't. They cannot ship all the CUDA packages. So the situation is complex and probably more accurately explained on their page TBH.

So actually - I'd just remove the last two sentences in this para.

docs/eessi.md
```diff
@@ -0,0 +1,63 @@
...
+-> /cvmfs/software.eessi.io/host_injections/2023.06/software/linux/x86_64/amd/zen3/software/CUDA/12.1.1/bin/nvcc
+```
+To expose the Nvidia GPU drivers.
+```[bash]
+/cvmfs/software.eessi.io/versions/2023.06/scripts/gpu_support/nvidia/link_nvidia_host_libraries.sh
+```
+
+### Buidling with GPUs
+
+Checking `nvcc --version` and `which nvcc` to see if `CUDA` compiler is found.<br>
```
sjpb:
```suggestion
Check `nvcc --version` and `which nvcc` to see if `CUDA` compiler is found.<br>
```

why's there  `<br>` tags in markdown? Just add an extra newline if you need a para.

docs/eessi.md
```diff
@@ -0,0 +1,63 @@
...
+<br>
+If `nvcc` not found run:<br>
+```[bash]
+export PATH=/usr/local/cuda-13.0/bin:$PATH
+```
+(with your specific cuda version)<br>
+`which nvcc` should now show path to compiler.<br>
```
sjpb:
```suggestion
`which nvcc` should now show the path to the compiler.
```

but fold it into previous line please!

docs/eessi.md
```diff
@@ -0,0 +1,92 @@
...
+## How to Load EESSI
+EESSI can be initialised using either of the following methods:
+
+1. Load the EESSI Environment Module.
+
+    The following command configures Lmod for your system and automatically loads the EESSI module:
+    ```[bash]
+    source /cvmfs/software.eessi.io/versions/2023.06/init/lmod/bash
+    ```
+    If Lmod is already installed, the following steps can be used to set up EESSI.
```
sjpb: `lmod` is always installed in slurm appliances, so we don't need the "if" and "check" bits.

docs/eessi.md
```diff
@@ -0,0 +1,92 @@
...
+If `nvcc` is not found, add the CUDA path to your environment. Replace `13.0` with your specific CUDA version:
+```[bash]
+export PATH=/usr/local/cuda-13.0/bin:$PATH
+```
+
+`which nvcc` should now show the path to the CUDA compiler.
+
+#### Loading EESSI module for the GCC compiler
+
+Running `which gcc` should initially show a path `.../2023.06/compat...` which points to the compatibility compiler.
```
sjpb:
```suggestion
Running `which gcc` with EESSI initialised should initially show a path `.../2023.06/compat...` which points to the compatibility compiler.
```

is that suggestion right/better?

docs/eessi.md
```diff
@@ -0,0 +1,92 @@
...
+```[bash]
+/cvmfs/software.eessi.io/versions/2023.06/scripts/gpu_support/nvidia/link_nvidia_host_libraries.sh
+```
+Rerun this script when your NVIDIA GPU drivers are updated. It is also safe to rerun at any time as the script will detect if the driver versions have already been symlinked.
+
+### Building with GPUs
+
+Check `which nvcc` to confirm that the CUDA compiler is found.
+
+If `nvcc` is not found, add the CUDA path to your environment. Replace `13.0` with your specific CUDA version:
```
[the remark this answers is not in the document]
claudia-lola: I think we can just use `export PATH=/usr/local/cuda/bin:$PATH`  I don't think it needs to be version specific. I have checked this works

docs/eessi.md
claudia-lola: yes I agree this is better

docs/eessi.md
```diff
@@ -0,0 +1,72 @@
+# EESSI Guide
+
+## How to Load EESSI
+EESSI can be initialised using the following method.
```
sjpb:
```suggestion
```

docs/eessi.md
```diff
@@ -0,0 +1,72 @@
...
+The EESSI environment is sourced (in a non-reversible way) by running:
+```[bash]
+source /cvmfs/software.eessi.io/versions/2023.06/init/bash
+```
+
+This is non-reversible because it:
+- Changes your `$PATH`, `$MODULEPATH`, `$LD_LIBRARY_PATH`, and other critical environment variables.
+- Sets EESSI-specific variables such as `EESSI_ROOT`.
+
+This is the recommended method because it:
```
sjpb:
```suggestion
However it is the recommended method because it:
```

docs/eessi.md
```diff
@@ -0,0 +1,72 @@
...
+- Sets up Lmod (environment module system).
+
+The [EESSI docs](https://www.eessi.io/docs/using_eessi/setting_up_environment/) offer another method to load EESSI, in addition to one above. The alternative method only initialises the Lmod module system and does not load a platform-specific setup. For these reasons, it is recommened to use the method detailed above.
+
+Successful environment setup will show `{EESSI 2023.06}` at the start of your CLI.
+
+To deactivate your EESSI environment you can either restart your shell using `exec bash` or exit the shell by `exit`.
+
+## GPU Support with EESSI
+To enable GPU support, you need a site-specific build that has CUDA enabled. For a guide on how to do this, please refer to [docs/image-build.md](../image-build.md).
```
sjpb:
```suggestion
To enable GPU support, the cluster must be running a site-specific image build that has CUDA enabled. For a guide on how to do this, please refer to [docs/image-build.md](../image-build.md).
```

docs/eessi.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Using GPUs
+All CUDA-enabled software in EESSI expects CUDA drivers in a specific `host_injections` directory.
+
+#### To expose the NVIDIA GPU drivers:
+Use the `link_nvidia_host_libraries.sh` script, provided by EESSI, to symlink your GPU drivers into `host_injections`.
+```[bash]
+/cvmfs/software.eessi.io/versions/2023.06/scripts/gpu_support/nvidia/link_nvidia_host_libraries.sh
+```
+Rerun this script when your NVIDIA GPU drivers are updated. It is also safe to rerun at any time as the script will detect if the driver versions have already been symlinked.
```
sjpb: So, this is an actual meaty question: does EACH user need to run this, i.e. is it a shell-specific thing? Or is it actually a "run-once" thing? Does it survive reboots (not rebuilds!)?

If its a run-once thing which survives reboots, we should probably do it ourslves, like we do for CVMFS config in:
- ansible/roles/eessi/tasks/configure.yml
- ansible/roles/compute_init/files/compute-init.yml
(search for EESSI)

sjpb: Actually if the answer is "we should do it ourselves" I don't think that blocks merging this PR actually. But I think we should understand the answer under the current ticket, and create a follow-on ticket if necessary.

sjpb (changes requested):
Looks OK to me now but it has linter errors which need fixing.

PR #828, opened 2025-10-17
Don't install hpl source during extra builds

sjpb:
Currently the hpl download runs unconditionally during extra builds, which is not necessary and fails these if the HPL website is down.


bertiethorpe (approved):
LGTM

PR #831, opened 2025-10-21
Use Ark repos for Open Ondemand installs (with in-repo GPG key)

sjpb:
- Fixes logic for `openondemand` role defaults overriding `osc.ood` role defaults (including now during build). Ark mirror of Open Ondemand repos now used for StackHPC image builds.
- Adds GPG keys for Open Ondemand repositories into the appliance - removes dependency on upstream server during image builds.
- Skips entire task files for Open Ondemand apps during "extra" image builds (rather than individually skipping tasks) to make build output less confusing.

TODO: needs manual testing.

### DIFF
--- a/environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
+++ b/environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
@@ -1,6 +1,6 @@
 {
   "cluster_image": {
-    "RL8": "openhpc-RL8-260106-1648-6e3cec46",
-    "RL9": "openhpc-RL9-260107-1036-6e3cec46"
+    "RL8": "openhpc-RL8-260107-1747-7e14a51d",
+    "RL9": "openhpc-RL9-260107-1747-7e14a51d"
   }
 }


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/18718573463

sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/20749668979

ansible/roles/openondemand/tasks/pam_auth.yml
```diff
@@ -3,29 +3,30 @@
...
   ansible.builtin.dnf:
     name: mod_authnz_pam
+  when: appliances_mode != 'configure'
 
-- name: Enable Apache PAM module
-  ansible.builtin.lineinfile:
-    path: /etc/httpd/conf.modules.d/55-authnz_pam.conf
-    line: LoadModule authnz_pam_module modules/mod_authnz_pam.so
-    regexp: ^LoadModule authnz_pam_module modules/mod_authnz_pam.so
+- when: appliances_mode != 'build'
```
elelaysh:
```suggestion
- when: appliances_mode == 'configure'
```

ansible/roles/openondemand/tasks/main.yml
```diff
@@ -24,139 +24,11 @@
...
-    path: /etc/ssh/ssh_known_hosts
-    create: true
-    block: "{{ _openondemand_login_key.stdout_lines | sort | join('\n') }}"
-    marker: "# {mark} ANSIBLE MANAGED BLOCK: openondemand login host" # allows other tasks to use blockinfile on this file
-    owner: root
-    group: root
-    mode: o=rw,go=r
+- ansible.builtin.include_tasks:
+    file: configure.yml
+  when: appliances_mode != 'build'
```
elelaysh:
```suggestion
  when: appliances_mode == 'configure'
```

elelaysh (changes requested):
Looks good to me.

Minor changes to keep conditions ==/!= 'configure as much as possible.

Untested

ansible/roles/openondemand/tasks/main.yml
sjpb: See above

sjpb:
> rpm_repo_url

Its not. We set `ood_use_existing_repo_file` instead which means the osd.ood task to install the repofiles doesn't run:
```shell
$ grep -rF ood_use_existing_repo_file ansible
ansible/roles/openondemand/defaults/main.yml:  ood_use_existing_repo_file: true
ansible/roles/osc.ood/tasks/install-package.yml:  when: ansible_os_family == "RedHat" and not ood_use_existing_repo_file
ansible/roles/osc.ood/tasks/install-package.yml:  when: ansible_os_family == "Debian" and not ood_use_existing_repo_file
ansible/roles/osc.ood/tasks/install-package.yml:  when: ansible_os_family == "RedHat" and not ood_use_existing_repo_file
ansible/roles/osc.ood/defaults/main/install.yml:ood_use_existing_repo_file: false
```

ansible/roles/openondemand/tasks/pam_auth.yml
elelaysh: There was no occurrence of `!= 'build'` and plenty of `== 'configure'` before this PR.

sjpb:
> is now actually taken into account. 

Yes. I should have called that out, but its complicated. Previously we didn't actually do the "override defaults" bit during build.

It also meant, I think, that `ondemand_package` didn't get used (it was malformed, so it couldn't have been) but we just got lucky that the upstream version was the one we specified (which is no longer the case). Which I only spotted b/c [this](https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/cleanup.yml#L43) does use the version, and wasn't deleting the installed files which mean trivy failed the vulnerability check :scream: This is why I do not like our image build logic ... [edit] But basically the problem is that 3rd party roles really are not written to be used in separate install/configure phases, which we need.

PR #833, opened 2025-10-21
Add retries to CI tofu apply

bertiethorpe:
Adds retries to CI tofu apply, due to frequent ephemeral failed runs


sjpb (commented):
LGTM, note tested added so final failure exits immediately without sleeping.

PR #836, opened 2025-10-23
Add option to install nvidia-fabricmanager

claudia-lola:
Adds `cuda_install_nvidiafabricmanger` to allow installing NVIDIA Fabric Manager during an image build with `cuda` enabled. Required for functioning GPUs on e.g. DGX H200 nodes.


sjpb (changes requested):
Once changes are done, could you [manually trigger an extra build](https://github.com/stackhpc/ansible-slurm-appliance/actions/workflows/extra.yml) on this branch please, AND add a comment linking to the workflow run?

Then we'll merge once that build passes/comments are addressed.

PR #841, opened 2025-10-28
Make eessi configure gpu node automatically

claudia-lola:
Add tasks to `eessi/configure.yml` and `compute-init.yml` to run the EESSI `link_nvidia_host_libraries.sh` script on gpu nodes with nvidia drivers installed. The tasks will be run when either `site.yml` is run or a rebuild via slurm is completed.

### DIFF
--- a/environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
+++ b/environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
@@ -1,6 +1,6 @@
 {
   "cluster_image": {
-    "RL8": "openhpc-RL8-251112-1307-e34d64c4",
-    "RL9": "openhpc-RL9-251112-1307-e34d64c4"
+    "RL8": "openhpc-RL8-251119-1202-332ac921",
+    "RL9": "openhpc-RL9-251119-1202-332ac921"
   }
 }


ansible/roles/compute_init/files/compute-init.yml
sjpb: I think we should replace this whole EESSI block with running the configure task directly. But it is not obvious how to do this TBF ...

ansible/roles/eessi/tasks/configure.yml
```diff
@@ -15,3 +15,20 @@
 - name: Ensure CVMFS config is setup # noqa: no-changed-when
   ansible.builtin.command:
     cmd: "cvmfs_config setup"
+
+# configure gpus
+- name: Check for NVIDIA driver
```
sjpb: I'm not sure whether there is always a /dev/nvidia0? Could you check with @jovial please for e.g MIG and vGPU configs? Else we'd have to do something like https://github.com/stackhpc/ansible-role-openhpc/blob/be6196540ca8007a0e45f2c3b2596ed0ff77fc13/library/gpu_info.py#L42 but TBH this approach here is much simpler!

ansible/roles/eessi/tasks/configure.yml
```diff
@@ -15,3 +15,20 @@
...
+  ansible.builtin.stat:
+    path: /dev/nvidia0
+  register: nvidia_driver
+
+- name: Set fact if NVIDIA driver is present
```
sjpb:
```suggestion
- name: Set fact if NVIDIA GPU is present
```

ansible/roles/eessi/tasks/configure.yml
jovial: Not in a slurm appliance, but on an openstack compute host with mig (it's hard for me to login into the slurm deployment with mig, but @priteau could check there):
```
[stack@gpu3 ~]$ ls -lia /dev/nvidia*
4226 crw-rw-rw-. 1 root root 503,   3 Feb 19  2025 /dev/nvidia-vgpu3
3610 crw-rw-rw-. 1 root root 503,   4 Feb 11  2025 /dev/nvidia-vgpu4
3665 crw-rw-rw-. 1 root root 503,   5 Feb 11  2025 /dev/nvidia-vgpu5
2205 crw-rw-rw-. 1 root root 503,   8 Jan 23  2025 /dev/nvidia-vgpu8
2091 crw-rw-rw-. 1 root root 503,   0 Jan 23  2025 /dev/nvidia-vgpuctl
2082 crw-rw-rw-. 1 root root 195,   0 Jan 23  2025 /dev/nvidia0
2086 crw-rw-rw-. 1 root root 195,   1 Jan 23  2025 /dev/nvidia1
2080 crw-rw-rw-. 1 root root 195, 255 Jan 23  2025 /dev/nvidiactl
```
So checking for /dev/nvidia0 looks good to me. You will likely not have the nvidia-vgpu* device nodes in a slurm deployment as they only appear if you create a vgpu instance.

sjpb (changes requested):
Looks fine, except the built images need adding  back into the CI config.

ansible/roles/compute_init/tasks/install.yml
```diff
@@ -54,6 +54,8 @@
       dest: roles/
     - src: ../../nhc
       dest: roles/
+    - src: ../../eessi/tasks/configure.yml
```
sjpb: Lets just copy in entire role, as we do for nhc

sjpb (changes requested):
My suggested refactoring was wrong, sorry! Lets try again - will need a new image build

Issue #680, opened 2025-05-26
Image build fails because of undefined cluster_compute_groups


priteau:
Packer image build fails because it refers to variable `cluster_compute_groups` defined by OpenTofu.

```
    openstack.openhpc: TASK [osc.ood : Install OOD apps] **********************************************                                                                                                                                                                                                                                                                     
    openstack.openhpc: task path: /home/rocky/image-build/ansible/roles/osc.ood/tasks/install-apps.yml:1                                                                                                                                                                                                                                                                    
    openstack.openhpc: The conditional check 'ood_install_apps is defined' failed. The error was: error while evaluating conditional (ood_install_apps is defined): {{ ood_install_apps_defaults if openondemand_jupyter_partition | default(none) else {} }}: {{ openhpc_partitions[0]['name'] }}: {{ openhpc_user_partitions + ([openhpc_rebuild_partition] if groups['reb
uild'] | length > 0 else []) }}: {{ openhpc_nodegroups }}: {{ cluster_compute_groups | map('community.general.dict_kv', 'name') }}: 'cluster_compute_groups' is undefined                                                                                                                                                                                                   
    openstack.openhpc:                                                                                                                                                                
    openstack.openhpc: The error appears to be in '/home/rocky/image-build/ansible/roles/osc.ood/tasks/install-apps.yml': line 1, column 3, but may
    openstack.openhpc: be elsewhere in the file depending on the exact syntax problem.                                                                                                
    openstack.openhpc:                                                                                                                                                                                                                                                                                                                                                      
    openstack.openhpc: The offending line appears to be:                                                                                                                              
    openstack.openhpc:                                                                                                                                                                                                                                                                                                                                                      
    openstack.openhpc:                                                                                                                                                                
    openstack.openhpc: - name: Install OOD apps                                                                                                                                                                                                                                                                                                                             
    openstack.openhpc:   ^ here                                                       
    openstack.openhpc: fatal: [default]: FAILED! => {}                                                                                                                                
    openstack.openhpc:                                                                                                                                                                                                                                                                                                                                                      
    openstack.openhpc: MSG:                                                                                                                                                           
    openstack.openhpc:                                                                                                                                                                                                                                                                                                                                                      
[...]

PR #843, opened 2025-11-03
Ensure OnDemand app installs during image build do not need a cluster to be deployed

bertiethorpe:
Fixes https://github.com/stackhpc/ansible-slurm-appliance/issues/680

This PR breaks the existing dependency for openondemand app installs to require a cluster to be deployed in the environment used for image builds.

- `openondemand_*_partition` is overridden for builder vars so that it is always set during image builds
- In this case the named partitions don't actually exist and so vars are simply truthy
- In the common environment, logic handles cases where openhpc_partitions is defined and populated, defined and empty and undefined

### DIFF
--- a/.github/workflows/workflow-cleanup.yml
+++ b/.github/workflows/workflow-cleanup.yml
@@ -1,0 +1,38 @@
+name: Workflow Cleanup
+on:
+  workflow_dispatch:
+      # checkov:skip=CKV_GHA_7: "The build output cannot be affected by user parameters other than the build entry point and the top-level source location. GitHub Actions workflow_dispatch inputs MUST be empty. "
+      inputs:
+        ci_cloud:
+          description: 'Select the CI_CLOUD'
+          required: true
+          type: choice
+          options:
+            - LEAFCLOUD
+            - SMS
+            - ARCUS
+        cluster_name:
+          description: 'Cluster to delete'
+          type: string
+          required: true
+
+
+permissions:
+  contents: read
+  packages: write
+  # To report GitHub Actions status checks
+  statuses: write
+
+jobs:
+  dummy:
+    name: dummy-job1
+    runs-on: ubuntu-22.04
+    env:
+      CI_CLOUD: ${{ github.event.inputs.ci_cloud }}
+      CLUSTER_NAME: ${{ github.event.inputs.cluster_name }}
+
+    steps:
+      - name: print input vars
+        run: |
+          echo CI_CLOUD: ${{ env.CI_CLOUD }}
+          echo CLUSTER_NAME: ${{ env.CLUSTER_NAME }}
--- a/environments/site/tofu/control.tf
+++ b/environments/site/tofu/control.tf
@@ -105,9 +105,8 @@ resource "openstack_compute_instance_v2" "control" {
       - [LABEL=state, ${var.state_dir}]
       %{if var.home_volume_provisioning != "none"}
       - [LABEL=home, /exports/home]
-      %{endif}
+      %{endif}%{if var.additional_cloud_config != ""}
 
-    %{if var.additional_cloud_config != ""}
     ${templatestring(var.additional_cloud_config, var.additional_cloud_config_vars)}
     %{endif}
   EOF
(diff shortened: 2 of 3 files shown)


environments/common/inventory/group_vars/all/openondemand.yml
```diff
@@ -12,11 +12,12 @@ openondemand_servername: "{{ hostvars[groups['openondemand'].0].ansible_host if
 
 openondemand_auth: basic_pam
 
-openondemand_jupyter_partition: "{{ openhpc_partitions[0]['name'] }}"
-openondemand_desktop_partition: "{{ openhpc_partitions[0]['name'] }}"
-openondemand_rstudio_partition: "{{ openhpc_partitions[0]['name'] }}"
+# Manages case where openhpc_partitions are not defined e.g. for site image builds
+openondemand_jupyter_partition: "{{ openhpc_partitions[0].name if (openhpc_partitions | default([])) else '' }}"
```
sjpb: So what this has missed is that in https://github.com/stackhpc/ansible-slurm-appliance/issues/680#issuecomment-3460798861 we said:
> - define cluster_compute_groups in common envs, empty, so that without any nodes deployed/inventory file we get empty openhpc_nodegroups and empty openhpc_partitions.


So with that done, I think this is/can just be the more obvious/ideomatic:

```suggestion
openondemand_jupyter_partition: "{{ openhpc_partitions[0].name | default('') }}"
```

docs/openondemand.md
```diff
@@ -63,6 +63,10 @@ The appliance automatically configures Open OnDemand to proxy Grafana and adds a
...
 
+## Image Build
+
+For local site image builds, the preferred method of installing ood apps in the image is by toggling the `openondemand_<app_name>_partition` variables in `environments/common/inventory/group_vars/all/builder/defaults.yml`. In this case the variables are not strings and are instead simply truthy i.e. they do not describe cluster partition groups but just whether those apps will be installed in the image or not.
```
sjpb: Hmm, I think we want to explicitly tell people how to override, we don't want people changing `common/`.

And b/c of how precedence works, changing `site/group_vars/all` won't work - it'll need changing `site/group_vars/builder/`

So maybe something _like_ :

```markdown
## Image Build
By default, most ondemand apps are installed in image builds when the build includes the inventory group `openondemand` (which is the default for "fatimage" builds). The apps installed are defined by the `openondemand_<app>_partition` variables in `environments/common/inventory/group_vars/all/builder/defaults.yml`.  Note that in this case the values are not strings and are instead simply truthy, i.e. they do not describe cluster partition groups but just whether those apps will be installed in the image or not.

For e.g. site-specific image builds where different app installs are required, due to precedence rules these must overriden in a `builder`-groupvars file e.g. `environments/site/inventory/group_vars/all/builder/defaults.yml`.
```

environments/common/inventory/group_vars/all/openondemand.yml
sjpb: And, given that "truthy" is not always obvious, is it better to make the default `false`?

environments/common/inventory/group_vars/builder/defaults.yml
```diff
@@ -27,3 +27,9 @@ sssd_enabled: false
 slurm_exporter_state: stopped
 appliances_mode: build
 proxy_remove: true
+# for image build ood partition var is just truthy as no cluster_groups defined
+openondemand_jupyter_partition: true
```
JohnGarbutt: I would be tempted by keeping the types consistent, with "fake_builder_partition" or something like that? With empty string for the ones bellow. With an updated comment.

... but this is also fine, if we get it working.

sjpb (changes requested):
There also appears to be a later image build run at e306e16 which failed: [601](https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/19229543159) 

Is there a reason we don't need that?

bertiethorpe:
> There also appears to be a later image build run at [e306e16](https://github.com/stackhpc/ansible-slurm-appliance/commit/e306e163591922dd69ac648185b7d60042e71640) which failed: [601](https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/19229543159)
> 
> Is there a reason we don't need that?

That failed build was for SMS, testing purposes

PR #844, opened 2025-11-04
Lock instances to prevent accidental changes

bertiethorpe:
The `ansible/site.yml` playbook now locks instances to prevent inadvertent changes/deletion to cluster instances. Instances are automatically unlocked by the `ansible/adhoc/rebuild-via-slurm.yml`. Instances must be manually unlocked using the `ansible/adhoc/unlock.yml` playbook before running `tofu apply`, `tofu destroy` or the development `ansible/adhoc/rebuild.yml` playbook.

---

An image build is not required to be commited in this PR, but should be carried out to test the revised code.

This only protects instances. Volumes for production environments should not be tofu-controlled as per the production docs.

Some alternative approaches were considered and rejected:
- Unlocking and relocking compute nodes from the RebootProgram: This is not feasible as the [rebuild call](https://github.com/stackhpc/slurm-openstack-tools/blob/96a7d826751d7d0cdee1da96ecc5da6134bc4e4e/slurm_openstack_tools/reboot.py#L111) returns immediately, and we do not want to block the script checking for rebuild completion.
- Preventing OpenTofu operations (apply, destroy etc) in a protected environment without some manual acceptance (in addition to the confirmation OpenTofu always prompts for). This can be implemented e.g. using variable conditions, but has to be done in the "root" module (i.e. environments/*/tofu/), not in the cluster module, as there is no way to directly set variables for the latter. The best approach at present would seem to be to define an extra cookiecutter file for tofu:

```hcl
variable "allow_tofu" {
  type = bool
  description = "Set to true to acknowledge the current environment is protected (e.g. production)"
  default = false
  validation {
    condition = anytrue(
      [
        ! contains(var.protected_environments, basename(var.environment_root)),
        var.allow_tofu
      ]
    )
    error_message = "Running tofu in a protected environment requires var.allow_tofu = true"
  }
}

variable "protected_environments" {
  type = list(string)
  default = ["production"]
}
```
This results in an error message like this if in a protected env:
```shell
tofu apply
...
Planning failed. OpenTofu encountered an error while generating this plan.

╷
│ Error: Invalid value for variable
│ 
│   on protected_environments.tf line 1:
│    1: variable "allow_tofu" {
│     ├────────────────
│     │ var.allow_tofu is false
│     │ var.environment_root is "/home/steveb/Documents/smslabs/ansible-slurm-appliance/environments/.stackhpc"
│     │ var.protected_environments is list of string with 1 element
│ 
│ Running tofu in a protected environment requires var.allow_tofu = true
│ 
│ This was checked by the validation rule at protected_environments.tf:5,3-13.
```

and actually running a plan/apply/delete requires using `tofu apply -var=allow_tofu=true`.

this would only apply to new environments, and would have to be manually-copied in to existing environments to get the benefit. Given this would not (on its own) protect from e.g. CLI/Horizon changes, it does not currently seem worth implementing.


sjpb:
Should probably do a fatimage build for this to check it doens't somehow break that. We don't need to bump the image though.

ansible/adhoc/lock_unlock_instances.yml
```diff
@@ -0,0 +1,11 @@
+---
+
+- hosts: "{{ target_hosts | default('all') }}"
```
sjpb: I don't think target_hosts is necessary, this can just be `hosts: cluster`. If you want to only run on some _as an adhoc_ you can just use `ansible-playbook --limit ...` which is what we do for e.g. the rebuild adhoc.

I know you suggested being able to "tweak" this for rebuild groups (presumably when running this from site, rather than as an ad-hoc, but TBH with the way that's passed via `vars:` at the moment you can't override it from inventory anyway, so I wouldn't bother.

sjpb (changes requested):
Initial review

sjpb (changes requested):

ansible/adhoc/lock_unlock_instances.yml
```diff
@@ -0,0 +1,11 @@
+---
+
+- hosts: "{{ target_hosts | default('cluster') }}"
```
sjpb: Now I understand why this is required, I don't really like the name `target_hosts`, I kinda feel it should be "namespaced" via a prefix to this lock/unlock business.

Maybe:
`lock_unlock_hosts` (or `_group`) and `lock_unlock_action` would be better?

But I could live with it 🤷 .

environments/common/inventory/group_vars/all/defaults.yml
```diff
@@ -4,6 +4,8 @@ ansible_user: rocky
 appliances_repository_root: "{{ lookup('env', 'APPLIANCES_REPO_ROOT') }}"
 appliances_environment_root: "{{ lookup('env', 'APPLIANCES_ENVIRONMENT_ROOT') }}"
 appliances_environment_name: "{{ appliances_environment_root | basename | regex_replace('\\W+', '') }}" # [a-zA-Z0-9_] only
+appliances_protected_environments:
+  - prd
```
sjpb: As per [production cluster docs](https://github.com/stackhpc/ansible-slurm-appliance/blob/main/docs/production.md#environments-structure) we recommend environments `staging` and `production`, so this should be `production`.

docs/experimental/slurm-controlled-rebuild.md
sjpb: I can't see any where in `docs/operations.md` or the main `README.md` which actually *directly* needs the lock/unlock adhoc referencing. But this feels a bit odd - what people have been doing won't work any more, so we do need to document this outside of slurm controlled rebuild.

sjpb (changes requested):

docs/experimental/slurm-controlled-rebuild.md
```diff
@@ -12,34 +12,36 @@ In summary, the way this functionality works is as follows:
...
    TODO: Describe the logic if they DO match
-6. After a rebuild, the compute node runs various Ansible tasks during boot,
+7. After a rebuild, the compute node runs various Ansible tasks during boot,
    controlled by the [compute_init](../../ansible/roles/compute_init/README.md)
    role, to fully configure the node again. It retrieves the required cluster
    configuration information from the control node via an NFS mount.
-7. Once the `slurmd` daemon starts on a compute node, the slurm controller
+8. Once the `slurmd` daemon starts on a compute node, the slurm controller
    registers the node as having finished rebooting. It then launches the actual
    job, which does not do anything.
```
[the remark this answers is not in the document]
bertiethorpe: Especially not with CI, as we wouldn't be able to clean up instances. I think we have to settle with manually relocking instances or re-running site.yml after rebuild / changes to tofu

wtripp180901 (changes requested):
Looks good, just a few docs tweaks

PR #845, opened 2025-11-06
Describe buildenv in EESSI docs

claudia-lola:
Adding section on `buildenv` module and removing Loading EESSI module for the GCC compiler section


sjpb (approved):
Thanks, this is very useful

PR #846, opened 2025-11-12
[draft] bootstrap: restore removing cockpit via dnf module

elelaysh:
This reverts a change made in #440 to debug errors in rocky update workflow

### DIFF
--- a/ansible/bootstrap.yml
+++ b/ansible/bootstrap.yml
@@ -207,10 +207,10 @@
   become: true
   tags: cockpit
   tasks:
-    - name: Remove RHEL cockpit # noqa: no-changed-when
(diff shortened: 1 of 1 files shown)


elelaysh:
Thank you @sjpb for the review.

What I really want to fix is the task showing as changed when it is not.

We could add a `dnf check-upgrade` command with `changed_when: false` to perform the initial metadata sync.
And then the dnf module to remove cockpit which handles changed correctly.

Or we could inspect the output of `dnf remove` for `"No package to remove"` or `"Nothing to do"` to compute the `changed_when` value.

How would you suggest reproducing the issues with the dnf module?

PR #857, opened 2025-12-02
Fix image-set-properties.sh: hw_disk_bus=virtio

elelaysh:
a wrong property was set, preventing launch from qcow2 images


sjpb (approved):
LGTM c.f. internally agreed docs.

PR #866, opened 2025-12-10
Fix sssd & sshd for slurm-controlled rebuild

sjpb:
Fixes an issue where sssd and sshd roles fail for compute-init, as their `export.yml` tasks write files to the cluster share as only readable by root. As the share is root-squashed these files cannot be retrieved by ansible-init on compute node boot. Other files were written by the `slurm` user but this is not considered appropriate for sensitive files.

The fix is to create a new "ansible-init" user and to use this for all writes and reads to/from the cluster share.

### DIFF
--- a/ansible/roles/hpctests/tasks/pingpong.yml
+++ b/ansible/roles/hpctests/tasks/pingpong.yml
@@ -64,9 +64,10 @@
   become: false
   when: hpctests_pingpong_plot | bool
 
-- ansible.builtin.debug:
-    # yamllint disable rule:line-length
-    msg: |
+- name: Set fact for summary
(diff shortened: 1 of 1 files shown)


sjpb:
Looks like its got stuck again at the "Add system users" task. Although logging into a CI VM that user at least has hte right gid now:
```
[rocky@slurmci-RL8-450-control ~]$ id ansible-init
uid=301(ansible-init) gid=301(ansible-init) groups=301(ansible-init)
```

so maybe somehow there were two problems??

sjpb:
Ok above CI did get past the point it had hung, so going to push new image ..

PR #867, opened 2025-12-11
Attempt to fix permissions for trivy scan on main

JohnGarbutt:
While the scan works on PRs, we have permissions related failures for the run on the main branch, this attempts to fix that.

More details on what is needed are here:
https://github.com/github/codeql-action/blob/c43362b91a940600cde2ebae39ec7a35ad66bdc0/upload-sarif/action.yml#L23C253-L24C20

Similarly, this moves to always scan the image on main, as a way to double check there are no new blocking CVEs more regularly than just when the image is bumped.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,6 +19,8 @@ permissions:
   # To report GitHub Actions status checks
   statuses: write
   id-token: write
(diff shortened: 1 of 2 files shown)


sjpb (approved):
LGTM

PR #869, opened 2025-12-11
Open OnDemand: Add dropdown to select select partition

elelaysh:
- use `auto_queues` instead of `bc_queue`, resulting in a `select` form item
- fix missing `--partition` argument in some app launch options


sjpb:
Needs a comment in PR that its been tested against multiple partitions with app default partition not being the slurm default partition.

PR #870, opened 2025-12-12
Write summary from hpctests pingpong to repo

sjpb:
Writes the summary description for the hpctests pingpong test (currently printed to stdout only) to the repo. This allows committing it alongside the existing plot and raw results to make reviewing performance drift easier.


elelaysh (dismissed):
LGTM
confirmed that it works in a small instance: the files appear in my checkout.

PR #875, opened 2025-12-18
CI: Cleanup OpenStack resources from previous attempts

sjpb:
CI clusters which fail during ansible (rather than opentofu) are deliberately not cleaned up at the end of the run, to allow debugging. However currently re-running the job fails because of duplicate volumes (as well as leaving resources around which are only cleaned up at night).

This PR cleans up resources from attempts before deploying the cluster at the current release.


elelaysh (approved):
LGTM

PR #877, opened 2025-12-19
alertmanager: support tuning systemd sandboxing

priteau:
We had to disable some systemd sandboxing options in an environment with NFS shares mounted.


sjpb (approved):
LGTM

PR #913, opened 2026-02-24
Fix node health checks failure for auto mounted filesystems

jovial:
Automounted filesystems need to be configured with the -F flag so that automounting is triggered[1]. Otherwise, node health checks fail with the following error:

```
rc:1 output:ERROR:  nhc:  Health check failed:  check_fs_mount:  /lustre mounted from systemd-1
```

[1] https://github.com/mej/nhc?tab=readme-ov-file#check_fs_mount

### DIFF
--- a/ansible/library/mount_facts.py
+++ b/ansible/library/mount_facts.py
@@ -1,0 +1,779 @@
+# -*- coding: utf-8 -*-
+# Copyright: Contributors to the Ansible project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+# NOTE: Adapted for Ansible 2.6 compatibility from the upstream code.
+# https://raw.githubusercontent.com/ansible/ansible/fbe9d7f32a99852cd952d7f4efb740069cbaf4bc/lib/ansible/modules/mount_facts.py
+
+DOCUMENTATION = """
+---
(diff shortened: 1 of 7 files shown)


ansible/library/mount_facts.py
```diff
@@ -0,0 +1,779 @@
+# -*- coding: utf-8 -*-
+# Copyright: Contributors to the Ansible project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+# NOTE: Adapted for Ansible 2.6 compatability from the upstream code.
```
jovial:
```suggestion
# NOTE: Adapted for Ansible 2.6 compatability from the upstream code.
# https://raw.githubusercontent.com/ansible/ansible/fbe9d7f32a99852cd952d7f4efb740069cbaf4bc/lib/ansible/modules/mount_facts.py
```

sjpb (changes requested):
Templating/facts look fine, boot logic needs fixing (and linting, as per slack)

jovial:
> Building a new fat image: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/24185772323/job/70590014076#logs

Actually - don't need this as no longer change boot.yml

sjpb (approved):

PR #917, opened 2026-03-03
WIP: make workflows restartable

sjpb:
TODO:
- also provides a job which can be required as a check


sjpb:
Don't like this approach either.

PR #931, opened 2026-03-18
Add psi=1 to kernel command-line

elelaysh:
via /etc/default/grub, using the grubcmdline role.

Pressure Stall Information refines the system load indicator into more up to date (<1 minute) and detailed (cpu, ram, io) metrics.
Node Exporter will grab them if available.
The grafana Node Exporter dashboard will have to be updated to a newer version to display the information.

### DIFF
--- a/ansible/fatimage.yml
+++ b/ansible/fatimage.yml
@@ -45,6 +45,22 @@
         pulp_site_target_distribution_version: "{{ hostvars[groups['pulp_site'][0]]['ansible_facts']['distribution_version'] }}"
       when: appliances_mode != 'configure'
 
+- hosts: builder
+  become: true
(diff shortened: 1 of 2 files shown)


elelaysh:
Fixed the GRUB_CMDLINE_LINUX_DEFAULT parameter override: I needed to specify `grubcmdline_kernel_cmdline_remove`, to avoid a catchall regex removing default items.

This is not a bug in the stackhpc.linux collection, as one shall always specify `grubcmdline_kernel_cmdline_remove`.

sjpb:
Should be possible to move on this now https://github.com/stackhpc/ansible-slurm-appliance/pull/987 has merged.

PR #940, opened 2026-04-01
Draft: Fix security issues flagged by grype

elelaysh:
Still a WIP, going through critical CVEs one at a time


elelaysh:
Part of #943

PR #942, opened 2026-04-07
Add support for setting minimum UID for users defined via basic_users

sjpb:
Adds a variable `basic_users_uid_min` to the `basic_users` role, and checks that the UIDs specified for normal (i.e. non-system) users are above this. This makes it harder to accidently mix interactive users defined via the role in with the autogenerated UIDs for users installed for services etc.

Note that testing showed setting `UID_MIN` in `/etc/login.defs` did not prevent users being created with incorrect UIDs - it simply prints a warning. Hence this is implemented as an assert.

Requires image build as role copied into basic_users.


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/25721747044

sjpb:
Above image build failing repeatedly on RL8 for openhpc.


edit: ok launched builder manually, hits

> [root@slurm-rl8-test-260512-1329-397bd574 rocky]# dnf list gcc
OpenHPC                                                                                                                                                    52  B/s |  50  B     00:00    
Errors during downloading metadata for repository 'OpenHPC':
  - Status code: 403 for https://ark.stackhpc.com/pulp/content/OpenHPC/2/EL_8/20241218T154614/repodata/repomd.xml (IP: 45.135.56.128)
Error: Failed to download metadata for repo 'OpenHPC': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried

But:
```
$ curl https://$ARKUSER:$ARKPW@ark.stackhpc.com/pulp/content/OpenHPC/2/EL_8/20241218T154614/repodata/repomd.xml
302: Found
```

So have we lost auth?? Yes we have:
```
[root@slurm-rl8-test-260512-1329-397bd574 rocky]# grep password /etc/yum.repos.d/OpenHPC.repo 
password = 
password = 
```

edit: not a fair test, local environment not set up to inject ark creds - will have to launch a CI build with no cleanup

ansible/roles/basic_users/defaults/main.yml
```diff
@@ -8,6 +8,7 @@ basic_users_userdefaults:
   ssh_key_comment: "{{ item.name }}"
   ssh_key_type: ed25519
   shell: "{{'/sbin/nologin' if 'control' in group_names else omit }}"
+basic_users_uid_min: 2000
```
sjpb: I don't think we can actually change this from the default of 1000 TBH, there will be clients relying on this

sjpb:
image builds https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/25739548617 - should not get cleaned up on failure

Turns out it was an infra problem, new Ark certs didn't have an intermediary which RL8 required. Fixed.

sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/25793029656

wtripp180901 (approved):
LGTM

PR #946, opened 2026-04-09
Run tofu from anywhere

jovial:
It's a pain the arse to switch to the tofu directory everytime you want to run tofu. This adds an alias which allows you to run tofu from anywhere, respecting the currently activated environment.

This can prevent mistakes such as when a user changes to the tofu directory of a different environment.

### DIFF
--- a/cookiecutter/{{cookiecutter.environment}}/activate
+++ b/cookiecutter/{{cookiecutter.environment}}/activate
@@ -18,3 +18,5 @@ echo "Setting PKR_VAR_repo_root to $PKR_VAR_repo_root"
 if [ -f "$APPLIANCES_ENVIRONMENT_ROOT/ansible.cfg" ]; then
    export ANSIBLE_CONFIG=$APPLIANCES_ENVIRONMENT_ROOT/ansible.cfg
 fi
+
+alias tofu="`which tofu` -chdir=$APPLIANCES_ENVIRONMENT_ROOT/tofu"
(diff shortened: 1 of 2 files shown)


sjpb (changes requested):
See comment

sjpb (changes requested):
Also, it'd be useful to change the .stackhpc activate script at the same time?

jovial:
> I don't think I like this; I think it would be better to use something like:
> 
> ```
> TF_CLI_ARGS="-chdir=${APPLIANCES_ENVIRONMENT_ROOT}/tofu"
> ```
> 
> There was a reason we didn't do this, I cannot totally recall details but it was something to do with processing variables or config, but I think it went away now we're using modules from site/tofu anyway.



> I don't think I like this; I think it would be better to use something like:
> 
> ```
> TF_CLI_ARGS="-chdir=${APPLIANCES_ENVIRONMENT_ROOT}/tofu"
> ```
> 
> There was a reason we didn't do this, I cannot totally recall details but it was something to do with processing variables or config, but I think it went away now we're using modules from site/tofu anyway.

Agree that would be cleaner, but.. I think -chdir behaves slightly differently to other options in that it must come before the subcommand[1]. Unfortunately, I think TF_CLI_ARGS only works for arguments after the subcommand [2].

I do find it quite a nice usability enhancement.

[1] https://github.com/opentofu/opentofu/issues/3976
[2] https://github.com/hashicorp/terraform/issues/28037

sjpb (approved):
LGTM

PR #947, opened 2026-04-09
Add pins from github-actions-list @ e6fd9c5

sjpb:


elelaysh (approved):
That's needed indeed for CI

PR #948, opened 2026-04-14
Support using a trunk port for node networking

MoteHue:
Locked behind the use_trunk flag. Unless this is set to true, the default networking is used.

I did want to do away with the use_trunk flag and just have it determined from the other vars being set, but it wasn't happy with that.
```
The "for_each" set includes values derived from resource attributes
```

There's still room for improvements here, these were just outside the scope of my initial support:
- Supporting multiple subports in the trunk
- Supporting the control node
- Docs

Would also be great to support one shared definition that can be used by mutliple node groups, no idea if this is doable though.

### DIFF
--- a/environments/site/tofu/node_group/network.tf
+++ b/environments/site/tofu/node_group/network.tf
@@ -12,3 +12,34 @@ data "openstack_networking_subnet_v2" "subnet" {
 
   name = each.value.subnet
 }
+
+resource "openstack_networking_port_v2" "trunk-parent" {
+  for_each = var.use_trunk ? toset(var.nodes) : []
+  name = "${split(".", local.fqdns[each.key])[0]}-trunk-parent"
+  network_id = var.trunk_parent_network_id
+  lifecycle {
+    ignore_changes = [extra_dhcp_option]
+  }
+}
+
+resource "openstack_networking_port_v2" "trunk-subport" {
+  for_each = var.use_trunk ? toset(var.nodes) : []
+  name = "${split(".", local.fqdns[each.key])[0]}-trunk-subport"
+  network_id = var.trunk_subport_network_id
+  lifecycle {
+    ignore_changes = [extra_dhcp_option]
+  }
+}
+
+resource "openstack_networking_trunk_v2" "trunk" {
+  for_each = var.use_trunk ? toset(var.nodes) : []
+  name = "${split(".", local.fqdns[each.key])[0]}-trunk"
+  admin_state_up = "true"
+  port_id        = openstack_networking_port_v2.trunk-parent[each.key].id
+
+  sub_port {
+    port_id           = openstack_networking_port_v2.trunk-subport[each.key].id
+    segmentation_id   = var.trunk_subport_vlan_id
+    segmentation_type = "vlan"
+  }
+}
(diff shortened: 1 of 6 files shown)


MoteHue:
Open question: I've included this in compute_fixed_image too, but I don't know if there's actually any point in this? Is there any need to use trunk points with virtual machines?

environments/site/tofu/node_group/variables.tf
```diff
@@ -214,3 +214,28 @@ variable "server_group_id" {
...
   default = null
 }
+
+variable "trunk_parent_network_id" {
+  type    = string
+  description = "ID of the network to use as the parent port of the trunk."
+  default = null
+}
+
+variable "trunk_subport_network_id" {
```
jovial, line 224: Potentially make this a list of networks?

environments/site/tofu/node_group/variables.tf
```diff
@@ -214,3 +214,28 @@ variable "server_group_id" {
   type    = string
   default = null
 }
+
+variable "trunk_parent_network_id" {
+  type    = string
+  description = "ID of the network to use as the parent port of the trunk."
```
jovial: Might be worth calling out that this controls the access (native) vlan

environments/site/tofu/node_group/variables.tf
MoteHue: Yep, that's the top in my improvements list. 
>There's still room for improvements here, these were just outside the scope of my initial support:
> - Supporting multiple subports in the trunk

I've only written enough here to cover the scope of the client work. Would be happy to extend this on R&D time if we decide it's the best use of my time. Could be a follow-up PR?

environments/site/tofu/node_group/variables.tf
MoteHue: Good point. This isn't clear if you don't already know how OpenStack trunk support works. Added.

environments/site/tofu/node_group/variables.tf
jovial: Fair - I missed that comment! Do have consider the module interface potentially changing though: string -> list. Meaning people have to update their config.

MoteHue:
TODO: hosts.yml needs to include the trunk subnet details for the node networks

environments/site/tofu/node_group/network.tf
```diff
@@ -12,3 +12,34 @@ data "openstack_networking_subnet_v2" "subnet" {
 
   name = each.value.subnet
 }
+
+resource "openstack_networking_port_v2" "trunk-parent" {
```
MoteHue: I should use underscored rather than hyphens

sjpb:
> Open question: I've included this in compute_fixed_image too, but I don't know if there's actually any point in this? Is there any need to use trunk points with virtual machines?

Whether the instance is a VM or BM has nothing to do with whether you're using the fixed or mutable image version of the instance, so yes its needed on both.

environments/site/tofu/node_group/variables.tf
sjpb: I think that's a really good point, we shouldn't merge until we're happy with that interface. So then it becomes a site-specific fix (= site deals with changes from upstream merge) rather than everyone having to deail with changes.

PR #950, opened 2026-04-24
Host Slurm logo image and reference it in ui-meta

m-bull:
The current logo URL returns a 404. Host the GPL-licensed image ourselves and reference it in ui-meta for all of the Slurm appliance variants.


sjpb (approved):
LGTM

PR #952, opened 2026-05-02
Add grubby with copy fail mitigations

sjpb:
Adds `grubby` role to modify kernel command line, with defaults to mitigate CVE-2026-31431 ("Copy Fail") via disabling AF_ALG functions built into the kernel. Role is enabled by default for `fatimage` builds.

TODO: decide if we actually need any reboot logic. Not very obvious what right thing is here TBH:
- Don't need to reboot a build in this case, do we ever??
- bootstrap.yml contains some logic to reboot on e.g. package changes, so could hook that if we DO ever need a restart, rather than doing x2.
- Could say you never want to reboot a running cluster via site.
- Could say you should give users the option, which when set (not by default) will do a reboot (probably via existing bootstrap reboot) IF there's been a change. This is probably the best get-out.
- Could say you should warn users if there's been a change. Very hard to make obvious. There's no way to force "needs-restarting" return true, apparently.
TODO: tidy up dev comments and convert test notes to comments here
TODO: actually build image

### DIFF
--- a/ansible/.gitignore
+++ b/ansible/.gitignore
@@ -106,3 +106,5 @@ roles/*
 !roles/mounts/**
 !roles/opkssh/
 !roles/opkssh/**
+!roles/grubby/
+!roles/grubby/**
--- a/ansible/bootstrap.yml
+++ b/ansible/bootstrap.yml
@@ -36,6 +36,14 @@
     - ansible.builtin.import_role:
         name: journald
 
+- hosts: grubby
+  become: true
+  gather_facts: false
+  tags: grubby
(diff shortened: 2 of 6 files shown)


elelaysh:
There is also https://github.com/stackhpc/ansible-collection-linux/tree/main/roles/grubcmdline to tweak the cmdline.

I found it difficult to use because it removed everything from the cmdline except what was specified, so I had to include default entries in the config.

sjpb:
Replaced by #956, now kernels available.

PR #955, opened 2026-05-06
Identify temporary vulnerability exclusions in sections in .grype.yaml

elelaysh:
We intend to allow some vulnerabilities temporarily, while keeping a close eye on them.
By allowing a vulnerability as soon as it has been analyzed, we the CI status can get back to _green_ and we will be sure to notice new vulnerabilities thereafter.
If we don't allow vulnerabilities temporarily the Check will be "red" more often due to them and we might fail to notice new vulnerabilities.

We shall put those vulnerability ignore lists in `VULNERABILITIES WE ARE STILL MONITORING` and temporary path exclusions in `TEMPORARY EXCLUSIONS`.

We also intend to run scheduled jobs that will run grype without those temporary allow lists, to keep an eye on them and remove them from the list when not needed.

I've not created the scheduled run yet, but made a simple program to remove temporary exclusions from the file.
We would use it like `./dev/remove-tmp-grype-exclusions.py .grype.yaml > .grype.clean.yaml && grype -c .grype.clean.yaml ...`

### DIFF
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -2,6 +2,29 @@
 # ignore rules for grype, with justification
 ignore:
 
+# {{{ BEGIN VULNERABILITIES WE ARE STILL MONITORING
+# Please add temporary/undecided vulnerability exclusions there,
+# as opposed to vulnerabilities that we are confident we can exclude.
+
+# go compiler bug introduces potential unverified memory access wr loops with induction variables on under or overflow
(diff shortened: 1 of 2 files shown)


sjpb:
Should be superseded by https://github.com/stackhpc/ansible-slurm-appliance/pull/979 using grype-analyse

PR #958, opened 2026-05-07
Fix pulp sync on RL8

### LINKED ISSUE #779: Pulp sync is broken on RL8
At https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/adhoc/sync-pulp.yml#L10, because vars "win" over inventory,  `pulp_site_target_distribution_version` can't be overridden to e.g. RL8.

sjpb:
Fixes a number of issues with pulp sync:
- `pulp_site_target_distribution_version` could not be overridden when using the ansible/adhoc/sync-pulp.yml playbook (closes #779)
- The existing templating for `pulp_site_target_distribution_version` in ansible/adhoc/sync-pulp.yml was incorrect for versions which do not sort as floats (e.g. 8.10)
- Logic in `pulp_site` role for constructing repo lists incorrectly assumed that all repo keys are present for every supported distribution

Does not need CI except linting, is not checked by that.

### DIFF
--- a/ansible/fatimage.yml
+++ b/ansible/fatimage.yml
@@ -32,12 +32,15 @@
 
 - name: Sync pulp repos with upstream
   hosts: pulp_site
+  gather_facts: true
   tasks:
     - ansible.builtin.include_role:
         name: pulp_site
         tasks_from: sync.yml
         apply:
           delegate_to: localhost
+      vars:
+        pulp_site_target_distribution_version: "{{ hostvars[groups['pulp_site'][0]]['ansible_facts']['distribution_version'] }}"
       when: appliances_mode != 'configure'
 
 - import_playbook: bootstrap.yml
--- a/ansible/adhoc/sync-pulp.yml
+++ b/ansible/adhoc/sync-pulp.yml
@@ -4,6 +4,3 @@
     - ansible.builtin.include_role:
         name: pulp_site
         tasks_from: sync.yml
-      vars:
-        # default distribution to *latest* specified for baseos repo:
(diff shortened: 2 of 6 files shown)


ansible/fatimage.yml
```diff
@@ -32,12 +32,15 @@
...
   hosts: pulp_site
+  gather_facts: true
   tasks:
     - ansible.builtin.include_role:
         name: pulp_site
         tasks_from: sync.yml
         apply:
           delegate_to: localhost
+      vars:
+        pulp_site_target_distribution_version: "{{ hostvars[groups['pulp'][0]]['ansible_facts']['distribution_version'] }}"
```
wtripp180901: Don't think this group exists anymore? I think we can just make this the OS version of the build VM

wtripp180901 (changes requested):
One issue with running syncs during fatimage build

ansible/fatimage.yml
sjpb: Agh that was what it was supposed to be, via `groups['pulp_site']`. Is there a better way of doing it?

wtripp180901: think that's probably the best we can do while we still have to delegate to running the venv on localhost

wtripp180901:
```suggestion
        pulp_site_target_distribution_version: "{{ hostvars[groups['pulp_site'][0]]['ansible_facts']['distribution_version'] }}"
```

wtripp180901 (approved):

PR #960, opened 2026-05-07
Fix cvmfs config templating in eessi role

MoteHue:
The variable `cvmfs_config` was being overloaded, being both used to register the output of the template job in
`ansible/roles/eessi/tasks/configure.yml`, and to actually define config for templating in
`ansible/roles/eessi/templates/cvmfs.config.j2`. This resulted in the ansible output being written to `/etc/cvmfs/local.conf`, instead of the desired config.


sjpb (approved):
LGTM

PR #963, opened 2026-05-12
Fix mounts role changes permissions on mounted contents

elelaysh:
Only create the mountpoint if it doesn't exist, 
to avoid changing permissions of the mounted content with the file module.

### DIFF
--- a/ansible/roles/mounts/tasks/main.yml
+++ b/ansible/roles/mounts/tasks/main.yml
@@ -1,12 +1,23 @@
+---
+
+- name: Check mount directories
+  ansible.builtin.stat:
+    path: "{{ item.path }}"
+  register: _mounts_mountpoints
+  loop: "{{ mounts.values() }}"
+  when: item.enabled | default(true)
+
 - name: Ensure mount directories exist
   ansible.builtin.file:
-    path: "{{ item.value.path }}"
+    path: "{{ stat_result.item.path }}"
     state: directory
     owner: root
     group: root
     mode: '0755'
-  loop: "{{ mounts | dict2items }}"
-  when: item.value.enabled | default(true)
+  loop: "{{ _mounts_mountpoints.results | default([]) }}"
+  loop_control:
+    loop_var: stat_result
+  when: (stat_result.stat | default(false)) and not stat_result.stat.exists
 
 - name: Ensure tmp.mount mask status is correct
   # RL8-specific fix; otherwise rootfs comes up read-only after a reboot!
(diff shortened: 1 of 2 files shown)


ansible/roles/mounts/tasks/main.yml
```diff
@@ -1,12 +1,23 @@
...
     state: directory
     owner: root
     group: root
     mode: '0755'
-  loop: "{{ mounts | dict2items }}"
-  when: item.value.enabled | default(true)
+  loop: "{{ _mounts_mountpoints.results | default([]) }}"
+  loop_control:
+    loop_var: stat_result
+  when: (stat_result.stat | default(false)) and not stat_result.stat.exists
```
sjpb: I don't get this line, what's the first bit trying to do? Check `.stat` is present? But

```shell
$ ansible localhost -m debug -a msg="{{ {} | bool }}"
localhost | SUCCESS => {
    "msg": false
}
$ ansible localhost -m debug -a msg="{{ {'a':1} | bool }}"
localhost | SUCCESS => {
    "msg": false
}
```

What's wrong with just doing `when: stat_results.stat.exists | default(false)` ?

```shell
ansible localhost -m debug -a msg="{{ stat_results.stat.exists | default(false) }}"
localhost | SUCCESS => {
    "msg": false
}
```

ansible/roles/mounts/tasks/main.yml
sjpb, line 1: Because this whole role is copied into the image by compute-init, this PR needs an image build

sjpb (changes requested):

ansible/roles/mounts/tasks/main.yml
elelaysh: You're right: `mydict.notthere.x | default(false)` will evaluate correctly.
Because of the negation, I have to write it as `when: not (stat_results.stat.exists | default(true))`

sjpb: Yes i think this is ok, but it would be helpful to add a comment first:
```suggestion
  # Only run when mountpoint is missing, to avoid resetting owner/mode on mounted directories
  # Also has to cope with .stat missing when enabled: false
  when: not (stat_results.stat.exists | default(true))
```

sjpb (changes requested):

elelaysh:
Build for d1a4039: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/25746767710

sjpb (approved):

PR #949, opened 2026-04-22
Fix Ansible assert boolean condition for RPM detection

mohaa7:
# Fix Ansible assert boolean condition for RPM check

## Summary
This PR fixes a failure in the Lustre build role caused by newer versions of ansible-core (>= 2.15) enforcing strict boolean evaluation in conditional statements.

The previous implementation used:

that: _lustre_find_rpms.files | length

The expression returns an integer. Older Ansible versions treated non-zero integers as truthy, but ansible-core >= 2.15 requires conditionals to explicitly evaluate to a boolean. As a result, builds fail with:

"Conditionals must have a boolean result"

## Change
Use the `matched` attribute returned by the `ansible.builtin.find` module and explicitly compare it to zero to produce a boolean result.

### Before (ansible/roles/lustre/tasks/install.yml)

- name: Check rpms found
  ansible.builtin.assert:
    that: _lustre_find_rpms.files | length
    fail_msg: "No lustre repos found with lustre_rpm_globs = {{ lustre_rpm_globs }}"

### After (ansible/roles/lustre/tasks/install.yml)

- name: Check rpms found
  ansible.builtin.assert:
    that:
      - _lustre_find_rpms.matched > 0
    fail_msg: "No lustre repos found with lustre_rpm_globs = {{ lustre_rpm_globs }}"

## Why
- ansible-core >= 2.15 enforces strict boolean evaluation.
- `_lustre_find_rpms.files | length` returns an integer.
- `_lustre_find_rpms.matched > 0` explicitly returns a boolean.
- The `matched` field is the native count provided by the `find` module and is more appropriate than recomputing length.

## Impact
- Fixes Packer image builds failing during the Lustre RPM detection step.
- Maintains compatibility with older Ansible versions.
- Ensures forward compatibility with modern ansible-core releases.
- No functional behavior change beyond correcting the conditional evaluation.

## Testing
Tested with ansible-core >= 2.15 during a Packer build:
- Lustre RPMs were generated using `make rpms`
- RPMs were successfully discovered via `ansible.builtin.find`
- Assertion passed when RPMs were present
- Build completed successfully

No regressions observed.

PR #964, opened 2026-05-14
Fix logic in build for desktop and lustre installs

sjpb:
Adopted from  #949.

- Fixes error logic in fatimage.yml for vnc desktop installation - so requires build
- Fixes logic in lustre installation: ansible-core >= 2.15 requires conditionals to explicitly evaluate to a boolean so lustre builds fail with "Conditionals must have a boolean result".

### DIFF
--- a/ansible/fatimage.yml
+++ b/ansible/fatimage.yml
@@ -152,7 +152,7 @@
       ansible.builtin.include_role:
         name: openondemand
         tasks_from: vnc_compute.yml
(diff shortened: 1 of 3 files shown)


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/25866324329

wtripp180901 (approved):
LGTM

PR #965, opened 2026-05-15
Mitigate ssh-keysign-pwn exploits

elelaysh:
CVE-2026-46333. See https://seclists.org/oss-sec/2026/q2/529 for details.
Disabling PTRACE_ATTACH for non-admins by settings the tunable
`kernel.yama.ptrace_scope=2` blocks known exploits.


elelaysh:
Build for eeeabae: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/25923600353

ansible/mitigations.yml
```diff
@@ -23,3 +23,32 @@
...
       changed_when: true
+
+- name: Apply ssh-keysign-pwn mitigation
+  hosts: mitigate_ssh_keysign_pwn
+  become: true
+  gather_facts: false
+  tasks:
+    # See https://docs.kernel.org/admin-guide/LSM/Yama.html for explanation
+    - name: Apply ssh-keysign-pwn mitigation to live system
+      ansible.builtin.command: sysctl kernel.yama.ptrace_scope=2
```
sjpb, line 37: Is there a reason this isn't using https://docs.ansible.com/projects/ansible/latest/collections/ansible/posix/sysctl_module.html?

A search shows we use it in https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/roles/podman/tasks/configure.yml#L16 so it must be available at our versions.

sjpb: Also the value is set to 2 (admin-only attach) internal comms show testing with 3
(no attach) only. Are we happy to rely on 3rd party reports that 2 also blocks it?

sjpb (commented):
Functionally looks fine. Using ansible.posix.sysctl might be preferable, and I haven't seen testing using 2.

ansible/mitigations.yml
elelaysh: Thanks, I will change it to ansible.posix.sysctl.
As to why I used a command: it felt more straightforward at the time.

Regarding value 2: I did test it when I wrote the playbook but not reported. To be honest I've not tested as root to confirm that root still can use attach_ptrace but it's not the point.

sjpb: > I did test it when I wrote the playbook but not reported. To be honest I've not tested as root to confirm that root still can use attach_ptrace but it's not the point.

Yes that's good enough for me, thanks.

elelaysh: Actually, let's stick with the command + templated file, so that we have clear comments in the resulting /etc/sysctl* instead of everything in /etc/sysctl.conf

sjpb:
Confirmed sysctl set correctly n boot of RL9 image in this PR:

```shell
[rocky@RL9-login-0 ~]$ cat /var/lib/image/image.json 
{
    "branch": "sec/disable-ptrace",
    "build": "openhpc-rl9-260515-1436-ce6c64e9",
    "cuda": "-",
    "doca": "-",
    "kernel": "5.14.0-611.54.1.el9_7.x86_64",
    "ofed": "-",
    "os": "Rocky 9.7",
    "slurm-ohpc": "24.11.5"
[rocky@RL9-login-0 ~]$ cat /proc/sys/kernel/yama/ptrace_scope
2
```

sjpb (approved):

PR #966, opened 2026-05-19
Add pkinit support to freeipa integration

sjpb:
- Adds support for enroling FreeIPA hosts using pkinit, i.e. certs/keys stored vault-encrypted in the repository. The previous approach using one-time-passwords and persisted keytabs remains the default for backwards compatibility.
- Removes support for deploying a freeipa server for development. This was not exercised  by CI or any client config, and made freeipa configuration overly complicated.

The `freeipa` role/group is not currently supported for compute-init so no image build is required.

Also fixes an issue with the ansible configuration during the lint workflow.


ansible/roles/freeipa/defaults/main.yml
```diff
@@ -1,15 +1,13 @@
...
-# freeipa_server_ip:
-freeipa_setup_dns: "{{ groups['freeipa_server'] | length > 0 }}"
-freeipa_client_ip: "{{ ansible_host }}" # when run on freeipa_client group!
-# freeipa_host_password:
-freeipa_user_defaults:
-  ipa_pass: "{{ freeipa_admin_password | quote }}"
-  ipa_user: admin
-freeipa_users: [] # see community.general.ipa_user
-freeipa_ca_cert_file: ''
-_freeipa_keytab_backup_path: "{{ hostvars[groups['control'].0].appliances_state_dir  }}/freeipa/{{ inventory_hostname }}/krb5.keytab"
```
wtripp180901, line 15: This is still referenced in enrol-otp

ansible/roles/freeipa/defaults/main.yml
```diff
@@ -1,15 +1,13 @@
...
-# freeipa_admin_password:
-# freeipa_server_ip:
-freeipa_setup_dns: "{{ groups['freeipa_server'] | length > 0 }}"
-freeipa_client_ip: "{{ ansible_host }}" # when run on freeipa_client group!
-# freeipa_host_password:
-freeipa_user_defaults:
-  ipa_pass: "{{ freeipa_admin_password | quote }}"
-  ipa_user: admin
-freeipa_users: [] # see community.general.ipa_user
-freeipa_ca_cert_file: ''
```
wtripp180901, line 14: Think we still need this too since enrol-otp is checking for empty string rather than null

wtripp180901 (changes requested):

wtripp180901 (approved):

PR #967, opened 2026-05-19
Use same dependencies for ansible-lint as actual environment

sjpb:
For linting to be correct, the python, ansible-core, ansible-lint, and galaxy role/collections all need to match what is used in the repo. The ansible/ansible-lint action cannot achieve that so this switches back to using dev/setup-env.sh to do the pip and galaxy installs then running ansible-lint directly in "offline" mode so it doesn't also try to install them. That required some changes so that ansible-lint and superlinter don't run on galaxy-installed roles.

Also
- Fixes a typo so `dev/run-linters.sh --fast` works.
- Removes deprecated terrascan config

This PR does not change the version of ansible-lint used; ansible-core has been bumped as little as possible for compatibility and all resulting python packages frozen in requirements.txt.


elelaysh (commented):
Looks good. Why the extra dev dependencies (black, mypy, ...) in requirements.txt?

sjpb:
> Looks good. Why the extra dev dependencies (black, mypy, ...) in requirements.txt?

All those changes are just from installing ansible-lint into my environment.

dev/run-linters
```diff
@@ -33,7 +34,7 @@ podman run --rm \
 	--env FIX_MARKDOWN_PRETTIER=true --env FIX_MARKDOWN=true \
 	--env FIX_NATURAL_LANGUAGE=true --env FIX_TERRAFORM_FMT=true\
 	--env FIX_SPELL_CODESPELL=true --env FIX_GITHUB_ACTIONS_ZIZMOR=true \
-	--env VALIDATE_CHECKOV=false --env VALIDATE_GITLEAKS \
+	--env VALIDATE_GITLEAKS \
```
elelaysh: CHECKOV burns with all my CPUs at 100% for 5 minutes. I'd rather keep it disabled if possible.

requirements.txt
```diff
@@ -48,11 +55,14 @@ oslo.serialization==5.8.0
...
 packaging==23.2
 passlib==1.7.4
+pathspec==1.1.1
 pbr==7.0.3
 pillow==12.1.1
 platformdirs==4.5.0
 prettytable==3.17.0
 psutil==7.1.3
+pulp-cli==0.23.2
+pulp-glue==0.23.2
```
elelaysh: those might have crept in?

requirements.txt
```diff
@@ -73,15 +84,21 @@ rfc3986==2.0.0
...
 rpds-py==0.29.0
+ruamel.yaml==0.19.1
+ruamel.yaml.clib==0.2.15
 schema==0.7.8
 setuptools==80.9.0
 six==1.17.0
 soupsieve==2.8
 stevedore==5.6.0
+subprocess-tee==0.4.2
+tabulate==0.10.0
```
elelaysh: this one might have crept in

requirements.txt
sjpb: Yeah you're right, from debugging

dev/run-linters
elelaysh: thanks

elelaysh (approved):
Thanks for fixing the CI

PR #969, opened 2026-05-21
Fix galaxy validation for non-symmetric difference filter

sjpb:
Fixes two issues with validation of current galaxy installs against requirements.yml:
- Correctness is now independent of ordering
- The `difference` filter turns out not to be symmetric, hence some cases where installs were not up to date could be missed.

---
The assymetry in the difference filter can be demonstrated like this:
```
- hosts: localhost
  become: false
  gather_facts: false
  vars:
    list_a: [a, b, c]
    list_b: [a,    c]
  tasks:
    - name: unexpected
      ansible.builtin.debug:
        msg: |
          list_a first: {{ list_a | difference(list_b) }}
          list_b first: {{ list_b | difference(list_a) }}
```
```
TASK [ansible.builtin.debug] *************************************************************************************************************************************************************
ok: [localhost] => {}

MSG:

list_a first: ['b']
list_b first: []
```


wtripp180901 (approved):
LGTM

PR #973, opened 2026-05-28
Run cuda-samples in adhoc/cudatests.yml

elelaysh:
Adding the ability to run all cuda-samples from the same playbook as nvbandwidth: ansible/adhoc/cudatests.

This runs 155 test runs for 147 executables of cuda-samples.

Running is not wrapped in slurm. It could be improved to be run in multiple tasks, with device constrained per task, to target each gpu on the system but it is left as a future improvement.

Compilation happens on all nodes: it crashes the system in a specific hardware instance so I think it's interesting to always compile.

### DIFF
--- a/ansible/adhoc/cudatests.yml
+++ b/ansible/adhoc/cudatests.yml
@@ -1,6 +1,5 @@
 ---
 - hosts: cuda
-  become: true
   gather_facts: true
   tags: cuda_bandwidth
   tasks:
@@ -8,3 +7,31 @@
       ansible.builtin.import_role:
         name: cuda
         tasks_from: bandwidth.yml
+
+- hosts: cuda
+  gather_facts: true
+  tags: cuda_samples
+  tasks:
+    - name: Compile cuda samples
+      ansible.builtin.import_role:
+        name: cuda
+        tasks_from: samples.yml
+    - name: Prepare output dir
+      ansible.builtin.tempfile:
+        path: "{{ cuda_samples_path }}/cuda-samples-{{ cuda_version_short }}"
+        prefix: "output-"
+        state: directory
+      register: cuda_samples_output_dir
+    - name: Show output dir
+      ansible.builtin.debug:
+        msg: "cuda samples will output files to: {{ cuda_samples_output_dir.path }}"
+    - name: Run cuda samples
+      ansible.builtin.command: >-
+        python3 ./run_tests.py
+          --config test_args.json
+          --dir build/bin
+          --output {{ cuda_samples_output_dir.path }}
+          --parallel {{ cuda_samples_run_tests_concurrency }}
+      args:
+        chdir: "{{ cuda_samples_path }}/cuda-samples-{{ cuda_version_short }}"
+      changed_when: true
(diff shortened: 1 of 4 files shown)


ansible/adhoc/cudatests.yml
```diff
@@ -1,10 +1,37 @@
...
     - name: Run CUDA bandwidth tasks
       ansible.builtin.import_role:
         name: cuda
         tasks_from: bandwidth.yml
+
+- hosts: cuda
+  gather_facts: true
+  tags: cuda_samples
+  tasks:
+    - name: Compile cuda samples
```
sjpb: Should this not all just be added to ansible/roles/cuda/tasks/samples.yml, like the compile? I think it was originally in there, then samples broke when cuda was updated (wasn't available at all) and we swapped in bandwidth insteda. But I can't see why its in the adhoc not the role? Especially as how is things like `cuda_samples_path` available here, its a role default and not in the inventory

ansible/roles/cuda/defaults/main.yml
```diff
@@ -13,11 +13,11 @@ cuda_packages_default:
...
   - "nvidia-fabricmanager-{{ cuda_nvidia_driver_version }}"
 cuda_packages: "{{ cuda_packages_default + ( cuda_packages_fabricmanager if cuda_install_nvidiafabricmanger | bool else [] ) }}"
-cuda_samples_release_url: "https://github.com/NVIDIA/cuda-samples/archive/refs/tags/v{{ cuda_version_short }}.tar.gz"
+# variables for cuda-samples
+cuda_samples_build_concurrency: "{{ ansible_processor_vcpus }}"
 cuda_samples_path: "/var/lib/{{ ansible_user }}/cuda_samples"
-cuda_samples_programs:
-  - deviceQuery
-  - bandwidthTest
+cuda_samples_release_url: "https://github.com/NVIDIA/cuda-samples/archive/refs/tags/v{{ cuda_version_short }}.tar.gz"
```
sjpb: Can we move this back up, just to minimise the diff?

sjpb (changes requested):
Meat of it looks fine, think it needs moving around a bit.

ansible/adhoc/cudatests.yml
elelaysh: Re. why variables work here:
> Variables from vars/ and defaults/ are imported into play scope unless you disable it via the public option in import_role/include_role.
[source](https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_reuse_roles.html#role-directory-structure)

elelaysh: It works even in previous tasks in the play the role is imported in but not in previous plays

elelaysh: Thanks for the explanation. I didn't dare put it in the role, because it seemed strange to not have it already.
I've now moved the tasks to the role.

ansible/roles/cuda/defaults/main.yml
elelaysh: done

PR #974, opened 2026-05-28
wip: Use ephemeral keys for CI deployment

sjpb:


sjpb:
Confirmed can ssh to CI instances (using debug key)

PR #975, opened 2026-05-28
Run gpuburn as part of hpctests

elelaysh:
- gpuburn is compiled once, then run in one task per GPU, one job per node.
- gpuburn can target muliple GPUs at once, but it doesn't work for a mix of full GPUs and MIGs. Running one gpuburn per GPU/MIG works around this limitation.

Please note that gpuburn can provoke hardware errors due to overconsumption/overheating in some hardware configurations.

### DIFF
--- a/.github/workflows/stackhpc.yml
+++ b/.github/workflows/stackhpc.yml
@@ -254,7 +254,7 @@ jobs:
         run: |
           . venv/bin/activate
           . environments/.stackhpc/activate
-          ansible-playbook -vv ansible/adhoc/hpctests.yml
+          ansible-playbook -vv ansible/adhoc/hpctests.yml --skip-tags hpctests-gpu
 
       - name: Run EESSI tests
         run: |
--- a/ansible/roles/hpctests/library/scontrol_node_info.py
+++ b/ansible/roles/hpctests/library/scontrol_node_info.py
@@ -1,0 +1,79 @@
+#!/usr/bin/python
+# pylint: disable=missing-module-docstring
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, StackHPC
+# Apache 2 License
+import json
+import re
+
+from ansible.module_utils.basic import AnsibleModule  # pylint: disable=import-error
+
+ANSIBLE_METADATA = {
+    "metadata_version": "0.1",
+    "status": ["preview"],
+    "supported_by": "community",
+}
+
+DOCUMENTATION = """
+---
+module: scontrol_node_info
+short_description: Get information about Slurm nodes via scontrol
+version_added: "0.0"
+description: >
+    Gets all the available information from Slurm's `scontrol show nodes` command about specified nodes.
+    The returned `info` property is a dict with keys from scontrol --
+    All parameters and values a list of strings in specified node order.
+options
+    nodes:
+        description:
+            - Slurm nodenames for which information is required.
+        required: true
+        type: list
+requirements:
+    - "python >= 3.6"
+author:
+    - Steve Brasier, StackHPC
+"""
+
+EXAMPLES = """
+TODO
+"""
+
+
+def run_module():  # pylint: disable=missing-function-docstring
+    module_args = {
+        "nodes": {
+            "type": "list",
+            "required": True,
+        }
+    }
+
+    module = AnsibleModule(argument_spec=module_args, supports_check_mode=True)
+    result = {"changed": False}
+    if module.check_mode:
+        module.exit_json(**result)
+
+    _, stdout, _ = module.run_command(
+        ["scontrol", "show", "nodes", "--json", ",".join(module.params["nodes"])],
+        check_rc=True,
+    )  # `--nodes` doesn't filter enough, other partitions are still shown
+    info = json.loads(stdout)
+    for node_info in info.get("nodes", []):
+        node_info["gpus"] = {}
+        if "gres" in node_info:
+            gres = node_info["gres"]
+            for g in gres.split(","):
+                if m := re.match(r"""^gpu:(.+):([0-9]+)\(S:[0-9]+\)$""", g):
+                    resource, cnt = m.group(1), m.group(2)
+                    node_info["gpus"][resource] = int(cnt)
+    result["info"] = info.get("nodes", [])
+    module.exit_json(**result)
+
+
+def main():  # pylint: disable=missing-function-docstring
+    run_module()
+
+
+if __name__ == "__main__":
+    main()
--- a/ansible/roles/hpctests/library/scontrol_node_info.py
+++ b/ansible/roles/hpctests/library/scontrol_node_info.py
@@ -33,7 +33,7 @@ options
 requirements:
     - "python >= 3.6"
 author:
-    - Steve Brasier, StackHPC
+    - Eric Le Lay, StackHPC
 """
 
 EXAMPLES = """
@@ -64,7 +64,8 @@ def run_module():  # pylint: disable=missing-function-docstring
         if "gres" in node_info:
             gres = node_info["gres"]
             for g in gres.split(","):
-                if m := re.match(r"""^gpu:(.+):([0-9]+)\(S:[0-9]+\)$""", g):
+                m = re.match(r"""^gpu:(.+):([0-9]+)\(S:[0-9]+\)$""", g)
+                if m:
                     resource, cnt = m.group(1), m.group(2)
                     node_info["gpus"][resource] = int(cnt)
     result["info"] = info.get("nodes", [])
--- a/ansible/roles/hpctests/tasks/run-gpuburn.yml
+++ b/ansible/roles/hpctests/tasks/run-gpuburn.yml
@@ -23,6 +23,29 @@
   ansible.builtin.debug:
     msg: "Will run --gres={{ hpctests_gpuburn_gres }}"
 
+- name: Compute node chunk size
+  ansible.builtin.set_fact:
+    hpctests_gpuburn_node_chunk_size: >-
+      {{ ((hpctests_computes.stdout_lines | length) * (hpctests_gpuburn_node_chunk_percent | int) / 100) | round(method="ceil") | int }}
+  when: hpctests_gpuburn_node_chunk_size is  not defined
+
+- name: Ensure node chunk size is not zero
+  ansible.builtin.assert:
+    that: hpctests_gpuburn_node_chunk_size | int > 0
+    fail_msg: >
+      Can't run gpuburn on node chunks of size 0.
+      Please change hpctests_gpuburn_node_chunk_percent or specify
+      hpctests_gpuburn_node_chunk_size directly
+
+- name: Compute node chunk count
+  ansible.builtin.set_fact:
+    hpctests_gpuburn_node_chunk_count: >-
+      {{ ((hpctests_computes.stdout_lines | length) / (hpctests_gpuburn_node_chunk_size | int)) | round(method="ceil") | int }}
+
+- name: Show node chunks
+  ansible.builtin.debug:
+    msg: "Will run gpuburn on {{ hpctests_gpuburn_node_chunk_count }} chunks of {{ hpctests_gpuburn_node_chunk_size }} nodes at a time"
+
 - name: Create sbatch script
   ansible.builtin.template:
     src: gpuburn.sh.j2
@@ -40,9 +63,10 @@
     chdir: "{{ hpctests_rootdir }}/gpuburn"
   register: hpctests_gpuburn_sbatch
   vars:
-    timeout_h: "{{ ((hpctests_gpuburn_minutes | int) // 60) }}"
-    timeout_m: "{{ ((hpctests_gpuburn_minutes | int) % 60) }}"
-  async: "{{ ((hpctests_gpuburn_minutes | int)+ 1) * 60 }}" # give async an extra minute after slurm
+    timeout_minutes: "{{ (hpctests_gpuburn_minutes | int) * (hpctests_gpuburn_node_chunk_count | int) }}"
+    timeout_h: "{{ ((timeout_minutes | int) // 60) }}"
+    timeout_m: "{{ ((timeout_minutes | int) % 60) }}"
+  async: "{{ ((timeout_minutes | int) + 1) * 60 }}" # give async an extra minute after slurm
   poll: 15  # check every 15 seconds
 
 - name: Extract results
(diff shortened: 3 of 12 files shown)


ansible/roles/hpctests/templates/gpuburn-build.sh.j2
```diff
@@ -0,0 +1,31 @@
...
+{% endif %}
+
+echo "arch: {{ hpctests_hpl_arch }}"
+echo "compute level: {{ hpctests_cuda_compute_level }}"
+{{ hpctests_pre_cmd }}
+# start from a clean state
+rm -r gpu-burn
+git clone https://github.com/wilicc/gpu-burn.git
+cd gpu-burn
+git checkout 671f4be92477ce01cd9b536bc534a006dbee058f # latest commit in master at this time
```
sjpb: I see there are no releases, but can we at least make this configurable via a variable?

ansible/roles/hpctests/library/scontrol_node_info.py
```diff
@@ -0,0 +1,79 @@
...
+options
+    nodes:
+        description:
+            - Slurm nodenames for which information is required.
+        required: true
+        type: list
+requirements:
+    - "python >= 3.6"
+author:
+    - Steve Brasier, StackHPC
```
sjpb:
```suggestion
    - Eric Le Lay, StackHPC
```

ansible/roles/hpctests/library/scontrol_node_info.py
sjpb, line 1: Can we not replace ansible/roles/hpctests/library/slurm_node_info.py with this?

ansible/roles/hpctests/templates/gpuburn.sh.j2
```diff
@@ -0,0 +1,30 @@
...
+echo SLURM_JOB_NODELIST: $SLURM_JOB_NODELIST
+echo SLURM_JOB_ID: $SLURM_JOB_ID
+echo gpuburn install dir: {{ hpctests_gpuburn_install_dir }}
+{{ hpctests_pre_cmd }}
+
+# shellcheck gets confused about jinja templates
+#shellcheck disable=SC1036,SC1056
+srun --output=%x.%N.%t.out --error=%x.%N.%t.out \
+	"{{ hpctests_gpuburn_install_dir }}/gpu_burn" \
+		-c "{{ hpctests_gpuburn_install_dir }}/compare.ptx" \
```
sjpb, line 30: this `-c` option doesn't appear to be shown in https://github.com/wilicc/gpu-burn/tree/671f4be92477ce01cd9b536bc534a006dbee058f?

ansible/roles/hpctests/templates/gpuburn.sh.j2
sjpb, line 1: I think we should make the default configuration less stressful than the max possible, to avoid issues with not-fully-ready powersupplies etc. Its not clear to me whether e.g. using the `    -m N%  Use N% of the available GPU memory` option would tune that in the same way that it does for HPL, but I'd assume so?

ansible/roles/hpctests/tasks/run-gpuburn.yml
```diff
@@ -0,0 +1,102 @@
...
+  block:
+    - ansible.builtin.stat:
+        path: "{{ hpctests_gpuburn_install_dir }}/gpu_burn"
+      register: gpuburn_exe_stat
+    - ansible.builtin.assert:
+        that:
+          - gpuburn_exe_stat.stat.exists and gpuburn_exe_stat.stat.executable
+        fail_msg: "Run build-gpuburn.yml to produce {{ hpctests_gpuburn_install_dir }}/gpu_burn"
+
+- name: Get Slurm node info
```
sjpb, line 12: Wondered if we could do this using the facts stackhpc.openhpc sets, but you can't, that's only "scontrol show config" parameters.

.github/workflows/stackhpc.yml
```diff
@@ -254,7 +254,7 @@ jobs:
         run: |
           . venv/bin/activate
           . environments/.stackhpc/activate
-          ansible-playbook -vv ansible/adhoc/hpctests.yml
+          ansible-playbook -vv ansible/adhoc/hpctests.yml --skip-tags hpctests-gpu
```
sjpb, line 257: Could we somehow rig it so we don't exclude it, but it just finds no gpus to run on so doesn't actually do anything? Then at least we'd get warning in CI if earlier steps fail.

sjpb (changes requested):

sjpb (commented):
I think we need a comment along the lines of that for hpl in the hpctests/README.md: 
>  The default of hpctests_hpl_mem_frac=0.3 will not significantly load nodes

But actually if you look at that link the markdown is all broken below that, could you fix at the same time please?

ansible/roles/hpctests/library/scontrol_node_info.py
elelaysh: I don't know. From our use in hpl-solo.yml it seems so. Would you prefer it?

ansible/roles/hpctests/templates/gpuburn.sh.j2
elelaysh: readme has not been updated.
See https://github.com/wilicc/gpu-burn/blob/671f4be92477ce01cd9b536bc534a006dbee058f/gpu_burn-drv.cpp#L777

ansible/roles/hpctests/templates/gpuburn-build.sh.j2
elelaysh: agreed

ansible/roles/hpctests/templates/gpuburn.sh.j2
elelaysh: now it will run on 10% of the nodes at a time by default. Use `-e hpctests_gpuburn_chunk_percent=100` to run on all nodes at once.

.github/workflows/stackhpc.yml
elelaysh: It will not be able to compile the test without cuda builtin. Not sure what else makes sense to test...

elelaysh:
Thank you for the review. Please see updated PR.

ansible/roles/hpctests/tasks/run-gpuburn.yml
```diff
@@ -0,0 +1,126 @@
...
+- name: Select GPUs
+  ansible.builtin.set_fact:
+    hpctests_gpuburn_gres: "{% for gpu in (hpctests_nodeinfo.info[0].gpus | default({})) | dict2items %}{% if loop.index0 > 0 %},{% endif %}gpu:{{ gpu.key }}={{ gpu.value }}{% endfor %}"
+  when: hpctests_gpuburn_gres is not defined
+
+- name: Show selected GPUs
+  ansible.builtin.debug:
+    msg: "Will run --gres={{ hpctests_gpuburn_gres }}"
+
+- name: Compute node chunk size
```
sjpb, line 26: So what happens if not all nodes in your job (selected via `hpctests_nodes` or `hpctests_partition`) don't have GPUs? The job won't run on nodes without them (due to --gres) but the nodelist for srun is calculated using all nodes - so e..g in a case where you run on a partition which combines two openhpc_nodegroups, one with GPUs and one without, some nodes in a batches' srun nodelist won't have GPUs. I think that's "safe", but it makes the batch size confusing. However the only other approach I can immediately think of would require a "dynamic" loop during the job, using `SLURM_JOB_NODELIST` env var, which would only include nodes actually allocated to the job (i.e. would exclude those filtered out from hpctests_partition/hpctests_nodes because they didn't have a GPU gres.

sjpb (changes requested):

ansible/roles/hpctests/library/scontrol_node_info.py
sjpb: Just seems un-ideal having two very nearly identical libraries - your approach with json is better now that's always available

ansible/roles/hpctests/tasks/run-gpuburn.yml
elelaysh: Indeed, the batch size could be made larger (based on `hpctests_gpuburn_node_chunk_percent`) if half the nodes don't have a gpu by computing it inside the batch scritp. Since we are trying to reduce the chunk size anyway with those settings, I don't think it matters. Also it becomes harder to grasp what the value will be for the user.

I find specifying the `hpctests_gpuburn_node_chunk_size` manually is better anyway. But since we want to be safe with as many configurations (small or big) as possible we need to use `hpctests_gpuburn_node_chunk_percent`.

The issue is I need the chunk count to compute the job timeout, so I would have to do filtering of nodes based on `hpctests_nodeinfo` in ansible. Do you think it's worth the extra error risk?

elelaysh:
rebased on latest main

sjpb (changes requested):
I think maybe the approach of limiting power by using a subnet of nodes is wrong; it looks like nvidia-smi supports ` --power-limit`. So can we just use that, and then run gpuburn on the entire set of (selected) nodes at once? And default to something small - maybe done as a %age of what nvidia-smi reports as max power limit?

elelaysh:
Here is a sample session of 

1. getting power limit info,
2. trying to set something lower than the min (fails)
3. trying without root (fails)
4. trying as root (succeeds)
5. power limit info again

Note that there isn't any way to reset power to "ipmi default": one has to remember the value before overriding it and restore it manually

```
# nvidia-smi -q -d power
==============NVSMI LOG==============

Timestamp                                              : Fri Jun 26 10:53:53 2026
Driver Version                                         : 595.71.05
CUDA Version                                           : 13.2

Attached GPUs                                          : 4
GPU 00000000:06:00.0
    GPU Power Readings
        Average Power Draw                             : 78.83 W
        Instantaneous Power Draw                       : 78.75 W
        Current Power Limit                            : 500.00 W
        Requested Power Limit                          : 700.00 W
        Default Power Limit                            : 700.00 W
        Min Power Limit                                : 200.00 W
        Max Power Limit                                : 700.00 W

(...)

# nvidia-smi -pl 50
Provided power limit 50.00 W is not a valid power limit which should be between 200.00 W and 700.00 W for GPU 00000000:06:00.0
Terminating early due to previous errors.

# nvidia-smi -pl 200
Failed to set power management limit for GPU 00000000:06:00.0: Insufficient Permissions
Terminating early due to previous errors.

# sudo nvidia-smi -pl 200
Power limit for GPU 00000000:06:00.0 was set to 200.00 W from 700.00 W.
Power limit for GPU 00000000:0C:00.0 was set to 200.00 W from 700.00 W.
Power limit for GPU 00000000:26:00.0 was set to 200.00 W from 700.00 W.
Power limit for GPU 00000000:2C:00.0 was set to 200.00 W from 700.00 W.
All done.

#  nvidia-smi -q -d power

==============NVSMI LOG==============

Timestamp                                              : Fri Jun 26 11:41:26 2026
Driver Version                                         : 595.71.05
CUDA Version                                           : 13.2

Attached GPUs                                          : 4
GPU 00000000:06:00.0
    GPU Power Readings
        Average Power Draw                             : 78.66 W
        Instantaneous Power Draw                       : 78.75 W
        Current Power Limit                            : 200.00 W
        Requested Power Limit                          : 200.00 W
        Default Power Limit                            : 700.00 W
        Min Power Limit                                : 200.00 W
        Max Power Limit                                : 700.00 W
    Power Samples
        Duration                                       : 2.36 sec
        Number of Samples                              : 119
        Max                                            : 79.03 W
        Min                                            : 78.41 W
        Avg                                            : 78.67 W
    GPU Memory Power Readings
        Average Power Draw                             : 42.02 W
        Instantaneous Power Draw                       : N/A
    Module Power Readings
        Average Power Draw                             : N/A
        Instantaneous Power Draw                       : N/A
        Current Power Limit                            : N/A
        Requested Power Limit                          : N/A
        Default Power Limit                            : N/A
        Min Power Limit                                : N/A
        Max Power Limit                                : N/A
    EDPp Multiplier                                    : 100.00%
(...)
```

sjpb:
> trying without root (fails)

`man nvidia-smi` documents  --power-limit as requiring root, so this is expected.

sjpb:
> remember the value before overriding it and restore it manually

That's annoying - so I guess needs some error handling in case the gpuburn actually fails, to avoid leaving it at the non-default value.

sjpb:
Annoying there doesn't seem to be a structured output option for nvidia-smi. I wonder if using http://pypi.python.org/pypi/nvidia-ml-py/ via a module is best. This is assuming you agree with the approach!

elelaysh:
> This is assuming you agree with the approach!

I think limiting how many nodes run the test at once is better suited to protect the rack/DC than limitting individual GPUs, given the power cap can't be lowered that much (eg 28% on H200).

On a practical note, changing the power profile necessitates sudo privileges, which will annoy operators trying to run hpctests without the full privileged rocky user access.

On the other hand I have not thought about limiting to a portion of nodes the other PR #978. Maybe it will not be as straightforward as what I did in b5790d1.

sjpb:
>  given the power cap can't be lowered that much (eg 28% on H200).

What am I missing here? Doesn't your example above show it can be limited to 28% of full power, 200W vs 700W? To me, if we could default it to 30% power, that's enough to be sensible/useful/cover us, and its really easy to explain what's going to happen then, rather than selecting a %age of nodes some of which may not actually have GPUs in the first place.

I guess it's still a bit hard to work out a sensible way to control this - e.g. I guess our input variable "gpuburn_power" or whatever goes from 0% (min power, 200W in this case) to 100% at max power (700W here), so its not 0% = 0W, but I can live with that. Edit: I'd much prefer this power-limit approach I think to the batching approach.

sjpb:
I'd forgotten about sudo though. While the *operator*/ansible needing that is ok, the default hpctests user doesn't / generally shouldn't. So maybe that kills it :-/

ansible/roles/hpctests/templates/gpuburn.sh.j2
elelaysh: changed to running on 1 node at a time, increased by setting `hpctests_gpuburn_chunk_size=N`. This is unambiguous

ansible/roles/hpctests/tasks/run-gpuburn.yml
elelaysh: I've removed hpctests_gpuburn_node_chunk_percent. Now it's running 1 node at a time.

elelaysh:
Rebased on latest main and removed `hpctests_gpuburn_node_chunk_percent` and set to run on 1 node at a time.

Tweaking the power settings of the cards in the test indeed requires `sudo`. That is not acceptable.
I'm convinced that running 1 node at a time is enough for safety. Would you agree?

ansible/roles/hpctests/library/scontrol_node_info.py
elelaysh: I'll remove it in a later PR, then

PR #977, opened 2026-05-29
Use all ansible connection properties for dev/ansible-ssh

sjpb:
Adds support for environment variables and ansible configuration to the `dev/ansible-ssh` tool. Makes dev more convenient since #974 uses environment variables for  StackHPC CI SSH configuration.

Tool is not tested in CI so can ignore everything except linting.

### DIFF
--- a/dev/ansible-ssh
+++ b/dev/ansible-ssh
@@ -38,9 +38,7 @@ if __name__ == "__main__":
     for line in output.splitlines():
         match = re.match(r".*SSH: EXEC (.*) '/bin/sh", line)
         if match is not None:
-            if len(match.groups()) != 1:
-                sys.exit("ERROR: Could not extract SSH command from output")
-            ssh_cmd = match.groups()[0]
+            ssh_cmd = match.group(1)
             # strip out verbosity flags
             ssh_cmd = re.sub(r"\s+-v+\b", "", ssh_cmd)
             # strip out our timeout arg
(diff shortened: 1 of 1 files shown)


dev/ansible-ssh
```diff
@@ -1,69 +1,54 @@
...
-    cmd = shlex.split(expanded)
-    if no_known_hosts:
-        cmd = cmd[0:1] + NO_HOSTKEY_CHECK_OPTS + cmd[1:]
+    ssh_cmd = None
+    for line in output.splitlines():
+        match = re.match(r".*SSH: EXEC (.*) '/bin/sh", line)
+        if match is not None:
+            if len(match.groups()) != 1:
+                sys.exit("ERROR: Could not extract SSH command from output")
+            ssh_cmd = match.groups()[0]
```
elelaysh:
```suggestion
            ssh_cmd = match.group(1)
```

elelaysh: There will never be 0 groups when you have capturing brackets.

dev/ansible-ssh
```diff
@@ -1,69 +1,54 @@
...
+            # strip out verbosity flags
+            ssh_cmd = re.sub(r"\s+-v+\b", "", ssh_cmd)
+            # strip out our timeout arg
+            ssh_cmd = re.sub(r"-o ConnectTimeout=1", "", ssh_cmd)
+            continue
```
elelaysh: we can exit now
```suggestion
            break
```

elelaysh (changes requested):
Great quality of life improvement.
A few tweaks

elelaysh (approved):
Good that you left the `-n ` as a noop, for commands in the history/memory

elelaysh (approved):

PR #978, opened 2026-05-29
Run NVIDIA HPL as part of hpctests

elelaysh:
-  the nvidia/hpc-benchmarks docker image is imported and run using apptainer
- NVIDIA HPL is run per host, using mpirun
- the HPL.dat file is selected based on given or discovered GPU model

It is possible to run one big NVIDIA HPL accross hosts, but only on clusters without cgroup resource isolation per task.
Since it is a common configuration, we run NVIDIA HPL via `mpirun` inside a slurm task per node as a workaround.

Please note that NVIDIA HPL can provoke hardware errors due to overconsumption/overheating in some hardware configurations and durations.

### DIFF
--- a/.github/workflows/stackhpc.yml
+++ b/.github/workflows/stackhpc.yml
@@ -254,7 +254,7 @@ jobs:
         run: |
           . venv/bin/activate
           . environments/.stackhpc/activate
-          ansible-playbook -vv ansible/adhoc/hpctests.yml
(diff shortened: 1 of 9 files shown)


elelaysh:
TODO: scontrol_node_info.py is identical to #975, to have independent pull requests.

elelaysh:
rebased on latest main

PR #981, opened 2026-06-03
Allow multiple keys to be injected into user's authorized_keys

jovial:
Use cases:

1) User has multiple keys e.g primary yubikey and a backup yubikey 
2) Multiple operators can have their own public key configured for the
   rocky user. This improves the audit trail rather than sharing a
common key and allows the operators to each maintain their own checkout of the slurm appliance.
3) Similar to 2, but one KEY for CI and one key for local operator.

We already run basic_users in ansible-init, so the only change that needs to made is to make the public_key take a list rather than a single key.

### DIFF
--- a/ansible/roles/basic_users/tasks/main.yml
+++ b/ansible/roles/basic_users/tasks/main.yml
@@ -86,9 +86,31 @@
     - item.create_home | default(true) | bool
     - ansible_hostname == basic_users_homedir_server
 
+- name: Normalise user key configuration
+  ansible.builtin.set_fact:
+    _basic_users_users: |
+      {%- set ns = namespace(users=[]) -%}
+      {%- for user in basic_users_users -%}
+      {# public_key is deprecated; include it for backward compatibility #}
+      {%-   set keys = namespace(items=[]) -%}
+      {%-   for key in user.authorized_keys | default([]) -%}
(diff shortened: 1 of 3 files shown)


ansible/roles/basic_users/tasks/main.yml
```diff
@@ -86,9 +86,21 @@
...
+      {# public_key is deprecated; include it for backward compatibility #}
+      {%-   set keys = (user.authorized_keys | default([])) + (([user.public_key] if user.public_key is defined else [])) -%}
+      {%-   set _ = ns.users.append(user | combine({'authorized_keys': keys})) -%}
+      {%- endfor -%}
+      {{ ns.users }}
+
 # The following tasks run on a single *client* node, so that home directory
 # paths are easily constructed, becoming each user so that root-squash
 # doesn't matter
+
```
jovial, line 113: Added as comment refers to multiple tasks

jovial:
I've realized this will never work for my use case... Was hoping to avoid tofu wanting to reprovision all nodes due to user_data changing when adding keys or removing keys. I thought we could basically inject with ansible-init, but this actually only works for compute nodes, whereas I was hoping we could use this for operators running ansible, but clearly that will never work.

PR #982, opened 2026-06-03
Allow multiple keys to be injected into user's authorized_keys

jovial:
Adds support for multiple keys to basic_users, along with ability to remove keys.

Use cases:

1) User has multiple keys e.g primary yubikey and a backup yubikey 
2) Removing old keys with state: absent

Adds support for disabling and adding additional keys to rocky user; this can be useful for debugging if LDAP is down.

### DIFF
--- a/ansible/.gitignore
+++ b/ansible/.gitignore
@@ -106,3 +106,5 @@ roles/*
 !roles/mounts/**
 !roles/opkssh/
 !roles/opkssh/**
+!roles/authorized_keys/
+!roles/authorized_keys/**
--- a/ansible/bootstrap.yml
+++ b/ansible/bootstrap.yml
@@ -110,6 +110,16 @@
       ansible.builtin.meta: reset_connection
       become: false
 
+- hosts: cluster
+  gather_facts: false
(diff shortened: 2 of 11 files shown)


jovial:
Based on a complaint about terraform wanting to reprovision the baremetal computes to change ssh keys when we were managing these with user_data.  I thought we could manage the SSH public keys in ansible instead. I didn't want to ignore changes to user_data in terrrform since that could mask important updates.

Was hoping that this would allow users to all checkout the code with their own account, but of course the compute_init technique can't be used for the control plane and login nodes.

I thought this would all be possible just by changing basic_users, but turns out it was more complicated since that is only for users on the shared storage.

sjpb:
Yes, can we discuss?? I think this needs some design. There is already `appliances_local_users`, and as you say the TF integration needs some thought. TF *has* to control the provisioning key really, so this is not simple ...

sjpb:
> baremetal computes to change ssh keys when we were managing these with user_data

This is a valid problem though, I have wondered about being able to do provisioning key rotation without delete/recreate.

PR #986, opened 2026-06-05
Migrate Grafana dashboards from Graph to Time Series (except Slurm dashboard)

wtripp180901:
Dashboards have been rebuilt from their upstream counterparts with slurmstats annotation options re-added
As these are being hosted in local files that we can modify, the tooling to replace the parameterised datasources with our fixed datasources is no longer needed for those dashboards
Slurm exporter dashboard is still using Angular dashboards as there appear to be unrelated issues with the exporter breaking the dashboard


wtripp180901:
Superseded by https://github.com/stackhpc/ansible-slurm-appliance/pull/987

PR #989, opened 2026-06-12
Fix nhc timeouts on /proc/cpuinfo with large cpu count

elelaysh:
Abridged version of the patch from nhc's dev branch, also proposed to ohpc in https://github.com/openhpc/ohpc/pull/2614

### DIFF
--- a/ansible/roles/nhc/tasks/runtime.yml
+++ b/ansible/roles/nhc/tasks/runtime.yml
@@ -1,15 +1,8 @@
 ---
-- name: Ensure nhc is patched
-  ansible.builtin.include_tasks:
-    file: install.yml
-  when: appliances_mode != 'configure'
-
 - name: Collect facts
   ansible.builtin.include_tasks: facts.yml
-  when: appliances_mode != 'build'
 
 - name: Ensure NHC configuration directory exists
-  when: appliances_mode != 'build'
   # When running site.yml after login/control upgrade, nhc group might be
   # enabled in repo, but as the compute nodes have not yet been upgraded they
   # do not have the package and hence this directory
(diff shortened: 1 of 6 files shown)


elelaysh:
nhc role is now called from fatimage.

ansible/roles/nhc/tasks/runtime.yml
sjpb, line 1: This is maybe a nit, but I (really) don't like the fact we've got `appliances_mode` switches INSIDE the role, this is the only place we have this. So it even less obvious than normal when you're trying to work out what runs when.

I'd prefer to do:
- role has 2 tasks/, `install.yml` (basically as-is) and `configure.yml` (basically current main). Maybe also a `main.yml` which calls both, just to be standard/allow dev etc.
- in the nhc block in slurm, we do the conditional there, so

```
- name: Setup Node Health Checks
  # Has to be done here as it requires openhpc repos etc for installation
  hosts: nhc
  become: true
  tags: nhc
  tasks:
    - ansible.builtin.include_role:
        name: nhc
        tasks_from: install.yml
    - ansible.builtin.include_role:
        name: nhc
        tasks_from: configure.yml
      when appliances_mode != 'build'
```

or whatever the appropriate logic is ... it is nice in dev to be able to do `ansible-playbook .. -e appliancee_mode=all` where all is just "Not build or configure".

Or some variation, we've done it a number of ways TBH. But I think hiding the conditional in the role is going to confuse us.

sjpb (changes requested):

sjpb:
Note to myself because I got confused: nhc now runs for fatimage only (not extra builds) because fatimage is in compute, and compute is in nhc.

ansible/roles/nhc/tasks/runtime.yml
elelaysh: done as suggested. Thank you

elelaysh:
Built image in https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/27762388854

sjpb (approved):

PR #993, opened 2026-06-19
Squid - limit default acl ports

axelsimon:
The default Squid config considers ports for several ancient protocols to be needed and safe.

We really don't (or shouldn't) need Gopher,  WAIS or FTP enabled by default in 2026, so this change removes all ports except for 80 and 443.

A second commit also renames SSL to TLS, in the spirit of updating things.

Mitigates CVE-2026-47729.


elelaysh (commented):
LGTM.
@sjpb?

sjpb (approved):
LGTM. Role is not use by compute-init so no image build required. This is a breaking change!

PR #1001, opened 2026-07-16
Update OS packages to fix ghostlock in RL8

jovial:
- Bumped RockyLinux 8 snapshots to bring in: `kernel-4.18.0-553.144.1.el8_10.x86_64.rpm` in RL8. This fixes the ghostlock CVE.
- Only bumped BaseOS as AppStream is missing some perl 5.26 packages at this moment in time.

### DIFF
--- a/ansible/ci/check_compute_rebuild.yml
+++ b/ansible/ci/check_compute_rebuild.yml
@@ -3,7 +3,7 @@
   become: false
   tasks:
     - name: Set localhost fact for appliance image version
-      set_fact:
+      ansible.builtin.set_fact:
         appliance_image_label: "RL{{ ansible_distribution_major_version }}"
       delegate_to: localhost
       delegate_facts: true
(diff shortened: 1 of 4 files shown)


jovial:
Running image build here: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/29572954682

jovial:
Think we need to wait for another repo sync. There seems to be one package that insists on an older perl:

```
 ==> openstack.openhpc:  Problem: module perl-DBI:1.641:8100020260714235741:60d027e1.x86_64 from appstream requires module(perl:5.24), but none of the providers can be installed
 ```
 
 Whereas everything else wants 5.26.

jovial:
No joy with the snapshots today. Rocky 9 was hitting an openssl issue:

https://github.com/python/cpython/issues/151504

so I've reverted those bumps and focused on Rocky 8 Only. Appstream still broken as of 17th July 2026.

sjpb:
@jovial when this completes, can you please check/confirm that the new logic has actually worked OK on RL8, i.e. the task `Check if current branch image differs from last release` should set `ci_cluster_image_updated: true`.

sjpb:
> Check if current branch image differs from last release

I did it:
RL8: "ci_cluster_image_updated": true
RL9: "ci_cluster_image_updated": false

So, fine.

sjpb (approved):

PR #1002, opened 2026-07-21
Support running HPC tests in a Slurm reservation

priteau:
This can be used to validate nodes while they are in a maintenance reservation.


wtripp180901 (approved):
LGTM

PR #1005, opened 2026-07-23
Add tool for live patching CVEs

wtripp180901:
- Adds script for live patching CVEs e.g
```
ansible-playbook ansible/adhoc/patch.yaml -e '{
  "patch_repos": [
    {
      "repo_name": "baseos",
      "subpath": "rocky/9.8/BaseOS/x86_64/os",
      "timestamp": "20260707T222716",
      "repo_file": "rocky"
    }
  ],
  "patch_packages": [
    "kernel"
  ]
}'
```

- Adds a tool `ansible/adhoc/check-kernels.yml` for ensuring that nodes scheduled for reboot have loaded a new kernel.

### DIFF
--- a/ansible/adhoc/check-kernels.yml
+++ b/ansible/adhoc/check-kernels.yml
@@ -1,0 +1,18 @@
+---
+- name: Check all hosts are on the same kernel
+  hosts: cluster
+  gather_facts: true
+  tasks:
+    - name: Check all hosts are running the same kernel
+      run_once: true
+      vars:
+        kernel_summary: >-
+          {% set result = {} %}
+          {% for host in ansible_play_hosts %}
+            {% set k = hostvars[host]['ansible_kernel'] | default('Unknown') %}
+            {% set _ = result.update({k: (result[k] | default([])) + [host]}) %}
+          {% endfor %}
+          {{ result }}
+      ansible.builtin.debug:
+        msg: "{{ (kernel_summary | from_yaml) | to_nice_yaml }}"
+      failed_when: (kernel_summary | from_yaml).keys() | length != 1
--- a/ansible/adhoc/check-kernels.yml
+++ b/ansible/adhoc/check-kernels.yml
@@ -4,7 +4,7 @@
   gather_facts: true
   tasks:
     - name: Check all hosts are running the same kernel
-      run_once: true
+      run_once: true # noqa: run-once[task]
       vars:
         kernel_summary: >-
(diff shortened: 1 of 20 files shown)


sjpb (commented):
- Lets drop "cve" everywhere for brevity
- Ignore the source control issue for this initial go
- Lets not actually do the reboot, BUT should work thro draft instructions / fix things so that reboot actually does work (both for slurm-controlled and normal, noting latter needs to not run from controller and we don't set a rebootprogram by default)
- Looked at proposed "schema" for repo definition, can't think of a better way!
- Don't worry about exposing ark password for now
- Lets split the legacy/modern Pulp approaches up for clarity

ansible/adhoc/check-kernels.yml
```diff
@@ -0,0 +1,18 @@
+---
+- name: Check all hosts are on the same kernel
```
sjpb, line 2: Is this always a prerequisite? I mean, I would expect all `cluster` to be on the same kernel, but I'm not sure why this is required?

ansible/adhoc/patch.yml
```diff
@@ -0,0 +1,132 @@
...
+#   "patch_repos": [
+#     {
+#       "repo_name": "baseos",
+#       "subpath": "rocky/9.8/BaseOS/x86_64/os",
+#       "timestamp": "20260707T222716",
+#       "repo_file": "rocky"
+#     }
+#   ],
+#   "patch_packages": [
+#     "kernel"
```
sjpb: I guess actually this would break e.g. ofed, nvidia-drivers, lustre. So maybe we shouldn't use kernel as an example?

sjpb (dismissed):
Checked variables/contents for pulp vars for current and legacy
Checked legacy version comment is correct
Checked message

ansible/adhoc/check-kernels.yml
sjpb: Actually - what calls this at all? Is it just informational? If so, ignore above comment!

wtripp180901: If we're patching kernels for a CVE and relying on Slurm reboot jobs to roll out reboots after the current job has finished (which could be any amount of time), I added this as tool to check which nodes have been updated (and check that the reboot jobs have actually worked). It's not called by anything, it's purely an adhoc monitoring tool

sjpb (approved):

PR #1007, opened 2026-07-28
Fix idempotency of mitigate-refluxfs.yml

MoteHue:


sjpb (commented):
Looks fine but needs linter errors fixing, just add `changed_when: true` for commands where appropriate, and set file perms please.

wtripp180901 (approved):
LGTM

PR #1008, opened 2026-07-28
Remove codespell

MoteHue:


priteau:
Are you removing it only because it is failing?

elelaysh:
Codespell catches some typos in docs so we'd rather keep it.

PR #1011, opened 2026-07-29
Switch to include_role for roles in prometheus collection

jovial:
These roles contain `argument_specs.yml` that rely on facts.

This does play well with tags:

```
TASK [prometheus.prometheus.node_exporter : Validating arguments against arg spec 'main' - Prometheus Node Exporter] *************************************************************************************************************************
fatal: [cclr-dev-control]: FAILED! => {}

MSG:

The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'system'. 'dict object' has no attribute 'system'. 'dict object' has no attribute 'system'. 'dict object' has no attribute 'system'
fatal: [cclr-dev-login-0]: FAILED! => {}
```

The issue is that tags may cause us not to gather facts.

Prevent the error by using a dynamic include over a static import which prevents the validation from running.

### DIFF
--- a/ansible/monitoring.yml
+++ b/ansible/monitoring.yml
@@ -34,7 +34,7 @@
   hosts: node_exporter
   tags: node_exporter
   tasks:
-    - ansible.builtin.import_role:
+    - ansible.builtin.include_role:
         name: prometheus.prometheus.node_exporter
 
 - name: Deploy Open OnDemand exporter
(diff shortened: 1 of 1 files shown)


jovial:
Trying image build here: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/30464858485

sjpb (approved):
LGTM, image build worked and appears to have installed prometheus

PR #1012, opened 2026-07-29
Fix logic when with-mkhomedir is not currently configured

jovial:
We only checked that `sssd` appeared in the output of `authselect current --raw`.  This meant that if `with-mkhomedir` was missing, but sssd was present, we failed to add with-mkhomedir.


sjpb:
Will require an image build as role is copied into image: ansible/roles/compute_init/tasks/install.yml

sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/31712192691

elelaysh (approved):
LGTM

PR #1014, opened 2026-08-04
Include git commit in /var/lib/image/image.json

elelaysh:
The git ref is present in the image name by default, which goes into the build field.
But we might as well include it explicitely, in case the image suffix is overridden and for clarity.

### DIFF
--- a/ansible/cleanup.yml
+++ b/ansible/cleanup.yml
@@ -99,6 +99,7 @@
   vars:
     image_info:
       branch: "{{ lookup('pipe', 'git rev-parse --abbrev-ref HEAD') }}"
+      commit: "{{ lookup('pipe', 'git rev-parse HEAD') }}"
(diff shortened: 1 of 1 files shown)


ansible/cleanup.yml
```diff
@@ -99,6 +99,7 @@
   vars:
     image_info:
       branch: "{{ lookup('pipe', 'git rev-parse --abbrev-ref HEAD') }}"
+      commit: "{{ lookup('pipe', 'git rev-parse HEAD') }}"
```
priteau, line 102: Note that the first eight characters of the commit SHA are in the `build` value.

elelaysh: yes. It is clearer for operators with an explicit field, though

sjpb (approved):
Fine by me although redundant. Lets not bother doing a build, the next build can pick this up.

PR #1016, opened 2026-08-04
Fix ansible-init scontrol update check

MoteHue:
With the update to Slurm v25.11.4 in the ansible-slurm-appliance v2.20, the error message from scontrol when updating to an invalid state has changed [1]. Fixes the `when` clause to check for this new error message.

1. https://github.com/SchedMD/slurm/commit/439717cb4889289136917ec78ca8e219d38e6f78


wtripp180901 (approved):
LGTM

elelaysh:
It might have needed an image build, but there will one before next release

PR #1017, opened 2026-08-10
Add modulejail adhoc playbook and use in mitigations

elelaysh:
1. new `ansible/adhoc/modulejail.yml` to compute `kernel_modules_modulejail_denylist`
2. `kernel_modules_modulejail_denylist` is used in `ansible/mitigations.yml`

loop in `ansible/mitigations.yml` doesn't scale to 2000 modules so rewrite loop in python module.

See docs/experimental/modulejail.md

Note that the `kernel_modules` group now defaults to the whole cluster. ie kernel denylisting will happen when running `site.yml` on the live cluster. Previously it was only run when building the image.


docs/experimental/modulejail.md
```diff
@@ -0,0 +1,40 @@
+# Limiting the loadable module list to mitigate future CVEs
+
+There has been multiple instances of CVEs found in less-used kernel modules these past months.
+
+Mitigating the vulnerability is as simple as adding it to [kernel_modules_vulnerable_denylist](../../environments/common/inventory/group_vars/all/kernel_modules.yml)
+and re-running the `ansible/mitigations.yml` playbook.
```
sjpb, line 6: Well, ideally you'd rebuild the image.

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,40 @@
...
+
+But it doesn't protect the system 2 cases:
```
sjpb:
```suggestion
But it doesn't protect the system in two cases:
```

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,40 @@
...
+- the attacker tricked an admin into loading the vulnerable module before the vulnerability is known;
+- the attacker ran a command that automatically loaded the module (eg. sctp_diag is loaded by the `ss -S` command).
+
+We propose to be proactive and ban any unused module at the time from being loaded.
+
+This has been integrated using the [modulejail](https://github.com/jnuyens/modulejail) script.
+
+## Using modulejail
+
+1. add hosts in the `modulejail_sample_hosts` group in your staging inventory
```
sjpb, line 19: Can we explain how, e.g. show the config snippet needed to add it?

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,40 @@
...
+2. `ansible-playbook -D ansible/adhoc/modulejail.yml`:
```
sjpb, line 20:
```suggestion
2. `ansible-playbook --diff ansible/adhoc/modulejail.yml`:
```

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,40 @@
...
+   2.1 installs the modulejail script
+   2.2 collects all loaded modules to constitute a whitelist of needed modules
+   2.3 runs modulejail in dry-mode with this whitelist and collects the result
+   2.4 generates the `kernel_modules_modulejail_denylist` variable in `environments/staging/inventory/group_vars/all/modulejail.yml`.
+3. keep `environments/staging/inventory/group_vars/all/modulejail.yml` under Git control.
+4. `ansible-playbook -D ansible/mitigations.yml` to deploy the new deny list on all nodes
```
sjpb, line 26: mitigations.yml runs during site.yml anyway, but currently the hosts for it are `kernel_modules` which by default is `builder`. So I don't think this will actually do anything?

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,40 @@
...
+5. Rebuild the image to include the new deny list, or ensure ansible/mitigations.yml is run on newly provisionned nodes.
```
sjpb, line 27: Hmm this is a bit confusing c.f. 4. Can we explain the workflow a bit more?

docs/experimental/modulejail.md
sjpb, line 19:
```suggestion
1. Add hosts in the `modulejail_sample_hosts` group in your staging inventory
```

and similar lots of other places

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,40 @@
...
+## Enabling a module
+
+1. if a module happens to be needed, edit `/etc/modprobe.d/appliance.conf` to remove the 2 lines containing it;
+   you can now `modprobe` the module.
+2. add it to `additional_modules` in `ansible/adhoc/modulejail.yml` and rerun the it
+3. run the `ansible/mitigations.yml` playbook to remove it from `/etc/modprobe.d/appliance.conf` on all hosts.
+
+## Discussion
+
+- Should mitigations.yml be included in compute-init and kernel_modules.yml exported?
```
sjpb, line 38: Well its a good question. It would let you change the list and have reimaged nodes it come up on boot with the new list. So maybe that is better than having to run an image build.

The other alternative is making it so site.yml does actually run the mitigations. But for the case where slurm updates a node, that updated node would be unprotected.

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,40 @@
...
+- 3-steps build, run modulejail, rebuild image is heavy.
+- should the `modulejail.yml` variable file be in site or current inventory?
```
sjpb, line 40: Definitely site in my view; you test it in staging, then apply the tested config to production.

sjpb (commented):
I haven't reviewed throughly yet, just tried to understand the workflow and make some helpful comments. I think there is a bit of a cyclical problem though; we need a live cluster (with mounts etc) to create the allowlist, but really we want the denylist "in the image". So you really need to a) run the modulejail adhoc on staging, but b) subsequently have it apply to any host you bring up. Using site.yml for that would actually be ok, except for cases where slurm reimages compute nodes, where site never runs against them.

There's also a question for me if we want to split the config, say control/login/compute or gpu/non-gpu nodes might be fairly different? Maybe that is not worth worrying about as a first attempt anyway.

elelaysh:
Thank you Steve.

> I think there is a bit of a cyclical problem though; we need a live cluster (with mounts etc) to create the allowlist, but really we want the denylist "in the image". So you really need to a) run the modulejail adhoc on staging, but b) subsequently have it apply to any host you bring up. Using site.yml for that would actually be ok, except for cases where slurm reimages compute nodes, where site never runs against them.

Please note that we need live hosts for the allow (`lsmod`) and deny (`find /usr/lib -name *.ko*`) lists.

Yes, that's indeed clunky to build an up-to-date allow list in the image.

1. It is possible to rebuild the image once you have run the adhoc playbook on staging, then test the new image again on staging, then deploy on prod. That is heavy.
2. on sites without compute-init, hosts need to run `site.yml` to join the cluster anyway. Before that they are in limbo, with in particular, no sssd config allowing local users so privilege escalation is not an issue
3. on sites with compute-init, fresh mitigations should be pulled from the control host and run anyway I think. At least the `kernel_modules_deny_list` should be exported I think.

> There's also a question for me if we want to split the config, say control/login/compute or gpu/non-gpu nodes might be fairly different? Maybe that is not worth worrying about as a first attempt anyway.

Indeed there will be common modules + modules per hardware + modules for VMs. But what would be the goal in building denylists per class? With eg a VM for login nodes, users will have access to all module combinations so a vulnerable module would be accessible to them anyway...

docs/experimental/modulejail.md
elelaysh: Agreed. What I wanted to convey is that it's easy to mitigate when it's an less-used module: it can be added to `/etc/modprobe.d/appliance.conf` on all hosts cluster without any disturbance.

docs/experimental/modulejail.md
elelaysh: This would depend on the site, but eg.

```
# environments/staging/inventory/groups
[modulejail_sample_hosts]
staging-control
staging-baremetal-cpu-1
staging-baremetal-other-vendor-cpu-1
staging-vgpu-1
staging-baremetal-gpu-1
```

Does this help?

docs/experimental/modulejail.md
elelaysh: Err. yes: I always add cluster to `kernel_modules` children.

docs/experimental/modulejail.md
elelaysh: yes: it's confusing. I'll rewrite when we have decided on compute-init integration

docs/experimental/modulejail.md
elelaysh: ok. Will change in `adhoc/modulejail.yml`

docs/experimental/modulejail.md
sjpb: Yeah that's exactly the sort of esxample. I think we try to make it "copy-and-paste" where possible. And that conveys the important thing - one from each different "type" whatever that means.

docs/experimental/modulejail.md
sjpb: Hmm. Do you think we should actually change that to be the default then? What happens if you change the allowed modules from under a job? Maybe that's pathological and failing the job is actually OK in that case, but I think we want to be clear on intent.

sjpb:
> Thank you Steve.
> ...

So maybe there's actually two logical cases:

1. For sites without slurm-controlled-rebuild, running site.yml -> mitigations.yml (with cluster in kernel_modules group) is enough to protect really. We'd get most protection if we run mitigations early (maybe even before bootstrap, if it does not need network access?) or at least before we've provisioned users in iam.yml or extra.yml
2. For sites with slurm-controlled rebuild, that does NOT protect upgraded nodes. So you either run an adhoc once nodes are upgraded, or we build it into compute-init. For me the latter is clearly better.

So overall a production workflow probably goes
1. Build site image(s)
2. Deploy to staging
3. Run adhoc to get module list and populate environment vars
4. Run mitigations.yml or something to apply these to staging
5. Deploy production.
6. Run site.yml to configure cluster; this runs mitigations and protects cluster

If cluster does not use slurm-controlled rebuild, an upgrade just goes through 1-6 again.

If it is a slurm-controlled rebuild then an upgrade does 1-4 then:
5a. Deploy production login/control
6a. Run site.yml to reconfigure cluster; this runs mitigations and applies current protections to all nodes (so actually, exactly same as 6)
7a. Submit slurm jobs to upgrade compute nodes
8a. Node boots, compute-init retrieves module list and applies.

So in conclusion: we need to be able to trigger the protection code from:
1. site.yml
2. a specific playbook (or tags), so we can apply to the cluster we gathered modules on without having to re-run site
3. compute-init

I think what I don't really like with compute-init is we now get three states:

1. Old image with old module list
2. Old image with new module list
3. New image with new module list

which breaks the "atomicity" of an image-based update. BUT I think the alternatives are worse:
a. Rebuild image to add new module list - too slow
b. Exclude non-upgraded compute nodes when applying mitigations - doesn't make sense

So impact on this PR:
- already have mitigations.yml to run directly, no change needed
- we need to make site.yml actually run the mitigations via changing the kernel_modules group
- need to add compute-init support

ansible/adhoc/modulejail.yml
```diff
@@ -0,0 +1,137 @@
+---
+- name: Prepare module denylist using modulejail
+  hosts: modulejail_sample_hosts
+  gather_facts: false
+  vars:
```
sjpb, line 5: [Play vars beat inventory vars](https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_variables.html#understanding-variable-precedence), so none of these are overridable. Is that OK?? If not you'll either have to make this a role (and these role defaults) OR move them to environments/common

ansible/adhoc/modulejail.yml
```diff
@@ -0,0 +1,137 @@
...
+          --profile none
+      environment:
+        MODULEJAIL_NO_UPDATE_CHECK: "1"
+      register: _modulejail
+      changed_when: false
+      become: false
+
+    # shows blocked module count for information
+    - ansible.builtin.debug:
+        var: "_modulejail.stdout_lines[0]"
```
sjpb, line 102: Can we make this more informative wthout the code, e.g.


```suggestion
        msg: "Blocking {{ _modulejail.stdout_lines[0] }} modules"
```

sjpb (changes requested):

elelaysh:
> I think what I don't really like with compute-init is we now get three states:
> 
>     1. Old image with old module list
> 
>     2. Old image with new module list
> 
>     3. New image with new module list

Should we key mitigations by image build_info?

This would let us keep applying the old list for old image when running site.yml after the control/login nodes are upgraded and still pick the new list when reimaged nodes state.

It introduces complexity when generating the variable file from the adhoc/modulejail playbook and/or in git history. 

For instance, the adhoc playbook could generate 2 var files:
1. `modulejail-{{ build_info }}.yml` containing `modulejail_denylist_version[{{ build_info }}] = [modules for this version]`
2. overwrite `modulejail.yml` containing `kernel_modules_denylist_modulejail_latest: "{{ modulejail_denylist_version[build_info }}"`

Then the mitigations playbook chooses which denylist to apply, defaulting to latest.

I'm not sure it's sound, to be honest.

Maybe compute_init export tasks can help prepare the right module list for the right host?

ansible/adhoc/modulejail.yml
elelaysh: I will move them to common

ansible/adhoc/modulejail.yml
elelaysh: The message is already structured.
Is it ok if I replace var with msg to produce this?

```
TASK [debug] ********************************************************************************************************************************************************************************************************
ok: [staging-cpu-01] => {}

MSG:

modulejail: DRY-RUN: would blacklist 2229 of 2350 modules (profile=none) -> /etc/modprobe.d/modulejail-blacklist.conf
ok: [staging-login-01] => {}

MSG:

modulejail: DRY-RUN: would blacklist 2229 of 2350 modules (profile=none) -> /etc/modprobe.d/modulejail-blacklist.conf
ok: [staging-control] => {}

MSG:

modulejail: DRY-RUN: would blacklist 2229 of 2346 modules (profile=none) -> /etc/modprobe.d/modulejail-blacklist.conf
```

sjpb:
> Should we key mitigations by image build_info?

I think the problem is that really you want your latest mitigations on all hosts. Else (with compute-init) you end up with vulnerable nodes while they're waiting for their current jobs to finish, which could be weeks.

So I think we just have a single mitigations list, and accept the multiple states. This is all experimental anyway TBH, so I think we try the "simple option" to start with.

docs/experimental/modulejail.md
elelaysh: changed for kernel_modules to be the default

docs/experimental/modulejail.md
elelaysh: thank you. I think I fixed them all

ansible/mitigations.yml
sjpb, line 1: Note to self: from site this runs before bootstrap

ansible/adhoc/patch.yml
sjpb, line 1: Given you're fixing one typo, could you fix another and remove the backtick at the end of line 105 please?

ansible/roles/kernel_modules/README.md
sjpb, line 1: Generally we document role vars here; it at least tells you want to grep for. I think we should do that, pluspoint out the defaults are in environments/inventory/group_vars/all/kernel_modules.yml because it interacts with vars for `ansible/adhoc/modulejail.yml`.

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,113 @@
+# Limiting the loadable module list to mitigate future CVEs
+
+There has been multiple instances of CVEs found in less-used kernel modules these past months.
```
sjpb:
```suggestion
There have been multiple instances of CVEs found in less-used kernel modules these past months.
```

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,113 @@
...
+
+Mitigating the vulnerability is as simple as adding it to [kernel_modules_vulnerable_denylist](../../environments/common/inventory/group_vars/all/kernel_modules.yml)
+and re-running the `ansible/mitigations.yml` playbook to patch your live cluster and/or rebuilding the image.
+
+But it doesn't protect the system in two cases:
```
sjpb:
```suggestion
However this does not protect the system in two cases:
```

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,113 @@
...
+
+- the attacker tricked an admin into loading the vulnerable module before the vulnerability is known;
```
sjpb:
```suggestion
- The attacker tricked an admin into loading the vulnerable module before the vulnerability is known.
```

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,113 @@
...
+- the attacker ran a command that automatically loaded the module (eg. sctp_diag is loaded by the `ss -S` command).
```
sjpb:
```suggestion
- The attacker ran a command that automatically loaded the module (eg. sctp_diag is loaded by the `ss -S` command).
```

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,113 @@
...
+
+We propose to be proactive and ban any unused module at the time from being loaded.
+
+This has been integrated using the [modulejail](https://github.com/jnuyens/modulejail) script.
+
+## Using modulejail
+
+1. Add hosts in the `modulejail_sample_hosts` group in your staging inventory.
+
+Pick one from each hardware/VM kind to cover all combinations of loaded modules.
```
sjpb, line 21: The indent under this 1. is wrong, see rendered view

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,113 @@
...
+
+On smaller deployments it can be run on all nodes:
+
+```yaml
+# environments/staging/inventory/groups
+[modulejail_sample_hosts:children]
+cluster
+```
+
+2. Run `ansible-playbook --diff ansible/adhoc/modulejail.yml`. It:
```
sjpb: Nested list wrong in rendered view

ansible/adhoc/modulejail.yml
```diff
@@ -0,0 +1,122 @@
+---
+- name: Prepare module denylist using modulejail
+  hosts: modulejail_sample_hosts
+  gather_facts: false
+  tasks:
+    - name: "Download modulejail"
```
sjpb: Nit but can we remove " from name: on this and following task - don't have them elsewhere

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,113 @@
...
+(in `ansible/final.yml`).
+
+We submit Slurm jobs to upgrade compute nodes. When they boot, they fetch the current `kernel_modules_denylist` and
+apply it as part of compute_init.
+
+## Enabling a module
+
+1. If a module happens to be needed, edit `/etc/modprobe.d/appliance.conf` to remove the 2 lines containing it;
+   you can now `modprobe` the module.
+2. Add it to `additional_modules` in `ansible/adhoc/modulejail.yml` and rerun the it
```
sjpb: I don't think this is right - should it be `modulejail_allowlist_modules` in `environments/common/inventory/group_vars/all/modulejail.yml`?

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,113 @@
...
+The `kernel_modules_denylist` variable is saved in the hostvars for each compute node on the cluster export of
+the control node when we run the compute_init `export.yml` tasks. This is run as part of `ansible/site.yml`
+(in `ansible/final.yml`).
+
+We submit Slurm jobs to upgrade compute nodes. When they boot, they fetch the current `kernel_modules_denylist` and
+apply it as part of compute_init.
+
+## Enabling a module
+
+1. If a module happens to be needed, edit `/etc/modprobe.d/appliance.conf` to remove the 2 lines containing it;
```
sjpb, line 75: Do we actually want to suggest 1? Or just use 2 and 3 to do it entirely via the appliance?

environments/site/inventory/groups
sjpb, line 1: Can we note this is changed, and the implication, e.g. migitigations now applied to live clusters, in the PR description please?

sjpb (changes requested):
Looks really nice. No real functional comments now and the contents of doc are very thorough. Just some minor comments really to tidy docs.

docs/experimental/modulejail.md
elelaysh: compute-init integration has been documented

docs/experimental/modulejail.md
elelaysh: Thanks for the catch, and correct! Now that it is moved to environments, I'm splitting it in `modulejail_allowlist_modules_default` and `modulejail_allowlist_modules_extra`.

ansible/roles/kernel_modules/README.md
elelaysh: ah, missed this. Thanks

ansible/adhoc/patch.yml
elelaysh: sure, done

docs/experimental/modulejail.md
elelaysh: yes, to show it is not magic. I've added that it is a stopgap measure and reworded the next item

environments/site/inventory/groups
elelaysh: good point, done

elelaysh:
Thank you for the review, @sjpb. Please consider another review or should I squash commits before?

docs/experimental/modulejail.md
elelaysh: kernel_modules are now part of compute-init

docs/experimental/modulejail.md
elelaysh: since I couldn't do nested lists I've removed the top-level one

elelaysh:
f53c10a Images Built in https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/33856448169

docs/experimental/modulejail.md
```diff
@@ -0,0 +1,116 @@
...
+(use `--tags kernel_modules` to not do anything else).
+
+The `kernel_modules` inventory group covers the whole `cluster`, so all active nodes are patched with the new denylist at this moment.
+
+### Without compute_init
+
+Without `compute_init` usage on a cluster, nodes are inactive until a first run of the `ansible/site.yml` playbook to configure them.
+In particular they are not accessible to non-admin users before that, so any LPE vulnerability would be inapplicable.
+
+As soon as they are provisionned, we run `ansible/site.yml` to configure them and apply the module denylist.
```
sjpb:
```suggestion
As soon as they are provisioned, we run `ansible/site.yml` to configure them and apply the module denylist.
```

sjpb (changes requested):

sjpb (approved):

PR #1018, opened 2026-08-11
Avoid rebooting nodes from site.yml

sjpb:
Fixes site.yml rebooting nodes if dnf thinks a reboot is required. This is not safe as it could break running jobs.

The problem is that site.yml runs a `bootstrap.yml` [reboot task](https://github.com/stackhpc/ansible-slurm-appliance/blob/9bfc38743017618a83790b34774d5ddcb4e9c570/ansible/bootstrap.yml#L292C14-L292C22) on `update` and `selinux` groups and the latter was set to `cluster` in the common environment. Setting groups in common env is not current practice anyway as it cannot be overriden.

Fix is to change the selinux group to default in site groups to `fatimage`, because actually changing selinux state is really only appropriate during build anyway.

### DIFF
--- a/environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
+++ b/environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
@@ -1,6 +1,6 @@
 {
   "cluster_image": {
-    "RL8": "openhpc-RL8-260807-0756-398534e4",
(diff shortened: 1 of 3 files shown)


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/31581938315

elelaysh (approved):
LGTM

PR #1019, opened 2026-08-12
dev/delete-cluster.py: purge volume snapshots

elelaysh:
This is required since a snapshot is taken when testing Slurm upgrades:
Volumes with snapshots need the '--purge' option to be deleted

### DIFF
--- a/dev/delete-cluster.py
+++ b/dev/delete-cluster.py
@@ -15,7 +15,11 @@ import json
 import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


elelaysh:
Fixed run: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/31597663671/job/94117054711#step:6:18

sjpb (approved):

PR #1020, opened 2026-08-12
Run syft in the packer VM and reference sbom in openstack image property

elelaysh:
1. Run syft in the packer VM at the end of the build to avoid lengthy image downloads for scan
    this adds 4-5 minutes to the build.

2. Store SBOMs in S3 for access from outside github ci.  
    Add metadata, including image name, id, hash, for traceability. 
    branches/PRs: openhpc-branch-sboms (flush by bucket policy)  
    master: openhpc-images-prerelease  (flush on new image push)  
    tags: openhpc-images  (never flushed)

3. Add a link to the SBOM in the openstack image metadata, so the scan action can use it to run grype directly.   
  This saves 30-50 minutes that were spent downloading, mounting and scanning the image.  

4. Preserve metadata when copying image and S3 objects


elelaysh:
Uploading with additional metadata and preventing overwrites, using s3cmd or minio-client

```
# s3cmd  put --add-header 'If-None-Match: *' --add-header 'x-amz-meta-github-run: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/31611182780' /tmp/openhpc-RL9-260812-0915-83ace25d-sbom.syft-json.zip s3://openhpc-sboms-test/openhpc-RL9-260812-0915-83ace25d-sbom.syft-json.zip.2 --disable-multipart

upload: '/tmp/openhpc-RL9-260812-0915-83ace25d-sbom.syft-json.zip' -> 's3://openhpc-sboms-test/openhpc-RL9-260812-0915-83ace25d-sbom.syft-json.zip.2'  [1 of 1]
 35548521 of 35548521   100% in    8s     3.95 MB/s  done
ERROR: S3 error: 412 (PreconditionFailed)
```

```
# mc put -q -H 'if-none-match: *' -H 'x-amz-meta-github-run: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/31611182780' /tmp/openhpc-RL9-260812-0915-83ace25d-sbom.syft-json.zip leafcloud-stackhpc-ci/openhpc-sboms-test/openhpc-RL9-260812-0915-83ace25d-sbom.syft-json.zip.2  --disable-multipart

`/tmp/openhpc-RL9-260812-0915-83ace25d-sbom.syft-json.zip` -> `leafcloud-stackhpc-ci/openhpc-sboms-test/openhpc-RL9-260812-0915-83ace25d-sbom.syft-json.zip.2`
mc: <ERROR> unable to upload. At least one of the pre-conditions you specified did not hold.
```

elelaysh:
Here is a clean build including uploading the sbom: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/31799484712
See the link to the sbom (and the image checksum -- quotes to be removed) in the summary.

Here is a scan downloading the sbom and putting it in cache, then running as usual: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/31809903704/job/94797758257
It is broken because all exclusions use absolute paths and relative paths are in the sbom.

.grype.yaml
sjpb, line 1: What's happened here? Looks noisy but I'd hope non-functional?

environments/common/inventory/groups
sjpb, line 1: This change needs to be replicated in environments/site/inventory/groups too

.github/workflows/fatimage.yml
```diff
@@ -34,17 +34,19 @@ jobs:
...
-            inventory_groups: fatimage
+            inventory_groups: fatimage,sbom
     env:
       ANSIBLE_FORCE_COLOR: 'true'
       OS_CLOUD: openstack
       CI_CLOUD: ${{ github.event.inputs.ci_cloud == 'default' && vars.CI_CLOUD || github.event.inputs.ci_cloud }}
       ARK_PASSWORD: ${{ secrets.ARK_PASSWORD }}
       PACKER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       PACKER_ON_ERROR: ${{ github.event.inputs.cleanup_on_failure == 'true' && 'cleanup' || 'abort' }}
+      S3_BUCKET: openhpc-sboms-test
```
sjpb: Needs changing before merge.

.github/workflows/fatimage.yml
```diff
@@ -159,12 +163,45 @@ jobs:
...
+        run: |
+          sudo apt-get update
+          sudo apt-get --yes install s3cmd
+
+      - name: Upload SBOM
+        run: |
+          . venv/bin/activate
+          sbom_filename="${STEPS_MANIFEST_OUTPUTS_IMAGE_NAME}.syft-json.zip"
+          sbom_url="${S3_BUCKET_PUBLIC_URL}/${sbom_filename}"
+          # `s3cmd --add-header 'If-None-Match: *'` doesn't work in multipart mode
```
sjpb:
```suggestion
          # Uses If-None-Match header to prevent over-writing - this does not work in multipart mode
```

.github/workflows/fatimage.yml
```diff
@@ -159,12 +163,45 @@ jobs:
...
+          STEPS_MANIFEST_OUTPUTS_IMAGE_NAME: ${{ steps.manifest.outputs.image-name }}
+          STEPS_MANIFEST_OUTPUTS_IMAGE_CHECKSUM: ${{ steps.manifest.outputs.image-checksum }}
+
       - name: Set image properties
         run: |
           . venv/bin/activate
           . dev/image-set-properties.sh "${STEPS_MANIFEST_OUTPUTS_IMAGE_ID}"
         env:
           STEPS_MANIFEST_OUTPUTS_IMAGE_ID: ${{ steps.manifest.outputs.image-id }}
+          STEPS_MANIFEST_OUTPUTS_IMAGE_NAME: ${{ steps.manifest.outputs.image-name }}
```
sjpb: This isn't used it it, why's it needed?

.github/workflows/release-image.yml
```diff
@@ -6,6 +6,7 @@ name: Release images
       - published # should work for both pre-releases and releases
 env:
   IMAGE_PATH: environments/.stackhpc/tofu/cluster_image.auto.tfvars.json
+  S3_BUCKET_PUBLIC_URL: https://leafcloud.store/swift/v1/AUTH_f39848421b2747148400ad8eeae8d536/openhpc-images
```
sjpb, line 9: This variable is used for both prerelease and release buckets across the changed workflows here. Can we make it less confusing?

.github/workflows/scan.yml
```diff
@@ -86,42 +85,25 @@ jobs:
           path: ${{ steps.manifest.outputs.sbom_name }}
           key: ${{ steps.manifest.outputs.sbom_name }}
 
-      - name: Download image
+      - name: Download image SBOM
         # NB: use /mnt as that is a separate disk with more space
```
sjpb: Comment isn't right any more

ansible/gen-sbom.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: Set sbom filename
+      ansible.builtin.set_fact:
+        sbom_filename: "{{ ansible_hostname }}.syft-json"
+    - name: Install and run syft
+      become: true
+      changed_when: true
+      ansible.builtin.shell: |
+        cd /tmp
+        set -eo pipefail
+        curl -L -o syft_1.51.0_linux_amd64.tar.gz  https://github.com/anchore/syft/releases/download/v1.51.0/syft_1.51.0_linux_amd64.tar.gz
```
sjpb: Could we parameterise the syft version and sha? Might encourage us to update it occasionally!

ansible/generate-sbom.yml
sjpb, line 1: I'd prefer either just `sbom.yml` or `generate-sbom.yml` than a bit of a random abbreviation

ansible/gen-sbom.yml
```diff
@@ -0,0 +1,40 @@
...
+        tar xf syft_1.51.0_linux_amd64.tar.gz syft
+        time ./syft scan \
+          --exclude ./dev --exclude ./proc --exclude ./sys --exclude ./run \
+          --config .grype.yaml \
+          --source-name "$sbom" --source-version 1 \
+          --output syft-json="$sbom" \
+          dir:/
+        ls -hl "$sbom"
+        zip "${sbom}.zip" "$sbom"
+        # dd if="$sbom" of="${sbom}.zip" bs=1M count=32
```
sjpb: remove?

.github/workflows/s3-image-sync.yml
```diff
@@ -120,8 +142,25 @@ jobs:
 
       - name: Upload Image to S3
```
sjpb, line 145: The properties which are set here as headers on the s3 image, are they used anywhere? E.g. the sync doesn't replicate them. I'm struggling a bit to understand the approach.

So what I can see is that:
- On image creation in openstack, we scan and upload the sbom to s3. Fine, it clearly can't live in glance. So we use glance properties to store the location. Fine.
- On sync, we download the image from openstack, get the sbom url from glance properties, and download that from s3. Fine.
- We also get a load of other properties from the sbom s3 object header.
- Then we upload the image to s3, setting the header properties from the sbom object on the image object.

Which mostly makes sense but:
- I'm not clear why the s3 image needs all those properties, which (I think?) duplicate those on the sbom anyway? Are they used or going to be used somewhere? E.g. I don't think the image sync copies them back into glance properties?
- The actual logic is pretty hard to follow from all the shell pipelines. Would it be clearer to have something like e.g. a script which does "copy these headers from this s3 object to that s3 object"?
- I still don't really understand why the sbom s3 url has to be different. Why can't we just use the existing pre-release/release buckets and store the sboms along with the images?

.grype.yaml
sjpb: Presumably something to do with the changed scan location/logic. But if we need relative paths now, we need a comment in the .grype to say that and why.

.github/workflows/s3-image-sync.yml
elelaysh: Thank you Steve for the review. I've straightened things up a bit (see *(NEW)* markers).

2 properties are added to the s3 image:
- `stackhpc_sbom`, which is used as you describe by the scan action to download the SBOM for the image
- `stackhpc_github_run`, which is not used in scripts, but we can use to find the job that produced the image *(NEW)*

The properties on the SBOM file in S3 are for traceability checking:
- `x-amz-meta-github-run`: not used by scripts, but for humans to check job logs that produced the SBOM (subject to github workflow log retention)
- `x-amz-meta-stackhpc-image-id`, `x-amz-meta-stackhpc-image-name`, `x-amz-meta-stackhpc-image-checksum`: identify the image that was parsed to produce the SBOM.  
They are used in 2 places:
- we compare them with the image properties during scan, to ensure we report data for the correct image.  *(NEW)*
- we compare again when publishing the image, to ensure the SBOM matches the image.  
It's arguably not airtight: content is changed between the syft run and the volume conversion to image  (we do some more cleaning via ansible), and a new image with the same name could be potentially be injected, but let's ignore it.

The properties on the QCOW image in S3 are for consumers of the image in S3, including outside the org.
They are there to support robustly downloading released images and confirming SBOM traceability:

- `x-amz-meta-github-run` is the run that produced the image
- `x-amz-meta-stackhpc-image-id` is to reuse the same image id (for convenience and idempotence in some places)
- `x-amz-meta-stackhpc-image-name` is the image name (arguably redundant with S3 object filename)
- `x-amz-meta-stackhpc-image-sha512` is the hash of the raw image, which can be used to compare with an existing image in another cloud and to doubly ensure integrity after download and image convert back to raw (qcow2 format doesn't have internal checksums)
- `x-amz-meta-stackhpc-qcow2-sha512` is the S3 object file checksum to ensure integrity of the downloaded file (the S3 host could provide this)
- `x-amz-meta-stackhpc-sbom-url` link to the SBOM from the image object, to abstract away from "the SBOM is next to the image"

Conversely, there are a few similar properties on the SBOM in S3:
- `x-amz-meta-github-run` is the run that produced the SBOM;
- `x-amz-meta-stackhpc-image-id`, `x-amz-meta-stackhpc-image-name`, `x-amz-meta-stackhpc-image-sha512` to confirm which image the SBOM relates to.

elelaysh: > Why can't we just use the existing pre-release/release buckets and store the sboms along with the images?

this will hold us back if we want to use a dependency tracking tool.

ansible/generate-sbom.yml
elelaysh: agreed

ansible/gen-sbom.yml
elelaysh: yes

ansible/gen-sbom.yml
elelaysh: yes, good point

.github/workflows/scan.yml
elelaysh: right

.grype.yaml
elelaysh: unfortunately `syft scan /` will produce relative paths in the SBOM.
grype can't apply exclusions starting with / to the root of the sbom, for some reason.
This leaves us with having to make all exclusions relative.
I had tried a bit to workaround this but couldn't find a solution.

environments/common/inventory/groups
elelaysh: done

.github/workflows/fatimage.yml
elelaysh: Changed to openhpc-branch-sboms

.grype.yaml
elelaysh: I've added a comment but leave the discussion open in case you would like to discuss more

.github/workflows/s3-image-sync.yml
sjpb: I guess my concern is that we're doing some of the same stuff as the azimuth manifest.json (e.g. see https://github.com/azimuth-cloud/azimuth-images/releases/tag/0.34.0), but in a different and (to me) less obvious/harder to work with way than just json in a file.

sjpb: But anyway, if we can get it working (see conflicts, looks like it needs an image rebuild :-() I'm happy to ignore my concerns and just merge it. It looks like the rest of my concerns have been dealt with.

sjpb (commented):
Looks OK to me, pending resolution of conflicts, which looks like it'll need another image build.

.github/workflows/s3-image-sync.yml
elelaysh: I think it's working properly now

.github/workflows/fatimage.yml
elelaysh: good point. Fixed

wtripp180901 (approved):

PR #1022, opened 2026-08-14
Fix pulp sync python version for RL8

sjpb:
Fixes an error where pulp sync on a RL8 host fails saying python3.9 is not found.


sjpb:
CI cancelled as doesn't test this - linting passing is sufficent.

elelaysh (approved):
LGTM

PR #1025, opened 2026-08-18
Fix proxy defaults for ansible-core 2.16.14

technowhizz:
The [CVE-2024-11079 ](https://github.com/ansible/ansible/pull/84353/changes#diff-b36fa2cf8e062dc662300f9e707a65fcdd9d6c3a51927a589643b67d3352e4f0R101) fix in ansible-core 2.16.14 changed hostvars iteration to include the implicit localhost once it has variables. Implicit localhost does not define ansible_host, causing no_proxy generation to fail when mapping ansible_host across every host.

This commit filters hostvars to entries that define ansible_host before constructing the default no_proxy list.

### DIFF
--- a/ansible/roles/proxy/defaults/main.yml
+++ b/ansible/roles/proxy/defaults/main.yml
@@ -1,7 +1,7 @@
 ---
 # proxy_http_proxy:
 proxy_https_proxy: "{{ proxy_http_proxy }}"
-proxy_no_proxy_defaults: "{{ ['localhost', '127.0.0.1', '169.254.169.254'] + groups['all'] + hostvars.values() | map(attribute='ansible_host') }}"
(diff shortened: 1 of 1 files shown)


elelaysh (approved):
LGTM

sjpb:
> when mapping ansible_host across every host.

Assuming "every host" means "all", I think this is generally the wrong thing to be doing. Usually, we should be using "cluster" instead of "all" when we mean all remote hosts. Because "all" may also include things like NFS servers which are in inventory so we can reference them, but that the appliance does not control. "cluster" specifically means "all the hosts we do control".

So, I think that would have been a better fix and really is still needed. But I can't see this as being harmful, so no need to revert.

sjpb:
Note my comment above is wrong, this does `hostvars.values() | map(attribute='ansible_host')` so the groups all has nothing to do with it.

PR #1026, opened 2026-08-18
Change README workflow badge to use correct workflow

sjpb:
Didn't get updated when workflows changed for linting.


elelaysh (approved):
good catch!

PR #1027, opened 2026-09-01
Remove test CA cert from release images

sjpb:
The `.stackhpc` environment contains a cert to test the `cacerts` role. By design, this role can be run during build, but this means the test cert ends up in release images. This PR adds a hook to delete it again. Note there is no security impact here; the key for the test cert was deleted immediately on creation.

Also adds build inventory groups information to `/var/lib/image/image.json`.


sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/34205727787

wtripp180901 (approved):

PR #1028, opened 2026-09-01
Fix opensearch role idempotence regarding owner/group

elelaysh:
- `username_facts` parses simple forms of `/etc/subuid` `/etc/subgid`.
- a simple computation gives the likely uid:gid of the opensearch user (rootless podman with `--userns=host`)

It can be overriden via `opensearch_host_user_id` `opensearch_host_group_id`, but then the situation might be more complex, eg using `--userns=auto:size=1000` or `--uidmap=...` that makes it unpredictible.

Also removed the :U flag from the `podman run` command in the service, because we generate files with correct permissions in the first place.


sjpb (approved):
LGTM

PR #988, opened 2026-06-08
Support baremetal ConnectX-7 infiniband in nhc role

elelaysh:
They are named eg mlx5_0:1.

 ibN also exists but it is not what is used by nhc healthchecks

PR #1029, opened 2026-09-01
Fix nhc export task undefined variable in templating by copying live config

elelaysh:
Avoid templating failure when the `facts.yml` tasks have not been called - introduced by https://github.com/stackhpc/ansible-slurm-appliance/pull/988.

Eg. when calling `ansible-playbook -D ansible/final.yml --tags compute_init` during development.

Note: the preferred pattern is now  to fetch custom files from live hosts in the export phase instead of duplicating the templating task to a different destination with `delegate_to: control`, to avoid this kind of issues.
Now don't forget to run  `ansible-playbook -D ansible/site.yml --tags nhc,compute_init --limit my-test-host` when you are developing the nhc role in compute_init mode. This will generate the config file on the live node, then export it.

### DIFF
--- a/ansible/roles/nhc/tasks/export.yml
+++ b/ansible/roles/nhc/tasks/export.yml
@@ -1,8 +1,15 @@
 ---
 # Used for compute-init
-- name: Template out host specific NHC config
-  ansible.builtin.template:
-    src: "{{ nhc_config_template }}"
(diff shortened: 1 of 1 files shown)


sjpb (approved):
Looks good - does not require image build as export.yml is only called from site

PR #1031, opened 2026-09-03
Don't always run rebuild role during image build

sjpb:


sjpb:
Implemented as part of #1032

PR #1032, opened 2026-09-03
Make slurm-controlled reboots drain nodes first and run in batches

sjpb:
Changes the approach for slurm-controlled rebuilds from using jobs with `--reboot` flag to `scontrol reboot`. There are several subtle changes:
- Nodes are now drained first to prevent cases where jobs could launch onto un-updated nodes.
- The default "nextstate" for `scontrol reboot` is RESUME (needed due to the explicit drain step), but the compute-init script no longer unconditionally resumes the node its running on. So both the compute-init script (via getting to slurmd start) and Slurm itself (via joining checks and NHC) decide if the node has successfully resumed.
- Rebuild calls are batched to try to avoid overloading the OpenStack Nova API which has occured in production deployments.
- The rebuild automation is now idempotent; if re-run it will be a no-op for nodes which have already been upgraded to the desired image, rather than a reboot.

The change of approach means the reboot partition is no longer required.

See ansible/roles/rebuild/README.md and docs/experimental/slurm-controlled-rebuild.md for more details.


sjpb:
Running image build ahead of fixing markdown linting: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/33778039719

sjpb:
Also this needs #1031 merging in first.

sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/34227430092

sjpb:
Image build: https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/35121404981

ansible/roles/rebuild/README.md
```diff
@@ -1,52 +1,45 @@
...
-  in the relevant partition.
+- `rebuild_clouds_key`: Optional. Name of key under `clouds:` entry in above
+  to use, default `openstack`.
+
+## Role Variables for tasks/rebuild.yml
+
+- `rebuild_nodes`: Optional list. Inventory hostnames/nodenames to consider
+  submitting for rebuild. The default is all nodes in the `compute_init` group.
+  **IMPORTANT:** to avoid jobs landing on a mix of updated and non-updated nodes,
+  this group must contain _all_ nodes in partitions for which it includes nodes.
```
m-bull: nit: Could use a specific note about nodes in multiple partitions

ansible/roles/rebuild/README.md
```diff
@@ -1,52 +1,45 @@
...
+- `rebuild_dryrun`: Optional bool. If `true` it will echo commands instead of
```
m-bull: nit: How is this presented to the user - does this get printed to screen with debug task? "echo" is ambiguous IMHO

ansible/roles/rebuild/tasks/rebuild.yml
```diff
@@ -1,11 +1,67 @@
...
   loop_control:
-    loop_var: _rebuild_job_current_partition
+    label: "{{ _rebuild_hostlists.stdout }}"
+
+- name: Set compute nodes requiring rebuild to DRAIN state
+  # NB: This must be done separately from the reboot call loop so that only rebuilt
+  # nodes are eligible to run new jobs
+  ansible.builtin.command:
+    cmd: >-
+      {% if rebuild_dryrun | bool %}echo{% endif %}
```
m-bull: I see what you mean in the docs now - its truly an `echo`. How useful is this to the user - in normal circumstances this never makes it to being output anywhere I don't think?

ansible/roles/rebuild/tasks/rebuild.yml
```diff
@@ -1,11 +1,67 @@
...
+  register: _rebuild_hostlists
+  changed_when: false
+  loop: "{{ _rebuild_targets | batch(rebuild_batch_size | int) }}"
   loop_control:
-    loop_var: _rebuild_job_current_partition
+    label: "{{ _rebuild_hostlists.stdout }}"
+
+- name: Set compute nodes requiring rebuild to DRAIN state
+  # NB: This must be done separately from the reboot call loop so that only rebuilt
+  # nodes are eligible to run new jobs
```
elelaysh:
```suggestion
  # NB: Ensures that old nodes won't receive new jobs right now, because scontrol reboot ASAP will be 
  # called in batches so nodes from later batches wouldn't yet be prevented from taking jobs.
```

ansible/roles/rebuild/tasks/rebuild.yml
```diff
@@ -1,11 +1,67 @@
...
+  ansible.builtin.command:
+    cmd: >-
+      {% if rebuild_dryrun | bool %}echo{% endif %}
+      scontrol update
+      state=DRAIN
+      nodename={{ item.stdout }}
+      reason='{{ rebuild_reason }}'
```
elelaysh: should we escape rebuild_reason?

ansible/roles/rebuild/README.md
```diff
@@ -1,52 +1,45 @@
...
+
+See [docs/experimental/slurm-controlled-rebuild.md](../../../docs/experimental/slurm-controlled-rebuild.md)
+for interaction with other roles and how to enable this.
 
-An OpenStack clouds.yaml file containing credentials for a cloud under the
-"openstack" key.
+## Requirements
 
-## Role Variables
+An OpenStack clouds.yaml file containing credentials for a cloud.
```
elelaysh:
```suggestion
An OpenStack `clouds.yaml` file containing credentials for a cloud.
```

ansible/roles/rebuild/README.md
```diff
@@ -1,52 +1,45 @@
...
-  must contain those nodes. Not for routine use, but may be useful to e.g.
-  reattempt a rebuild if this failed on specific nodes. Default is all nodes
-  in the relevant partition.
+- `rebuild_clouds_key`: Optional. Name of key under `clouds:` entry in above
+  to use, default `openstack`.
+
+## Role Variables for tasks/rebuild.yml
+
+- `rebuild_nodes`: Optional list. Inventory hostnames/nodenames to consider
+  submitting for rebuild. The default is all nodes in the `compute_init` group.
```
elelaysh:
```suggestion
  submitting for rebuild. The default is all nodes in the `compute_init` group.  
```

ansible/roles/rebuild/README.md
```diff
@@ -1,52 +1,45 @@
...
+  this group must contain _all_ nodes in partitions for which it includes nodes.
+- `rebuild_dryrun`: Optional bool. If `true` it will echo commands instead of
+  issuing them. Default `false`.
+- `rebuild_reason`: Optional string. A message to show in e.g. `sinfo` describing
+  the reason for the rebuild. Default `update`.
+- `rebuild_nextstate`: Optional string, `RESUME` (default), `DOWN` or the empty string.
+  The state rebuilt nodes should go to when a rebuild completes successfully. Note the
+  empty string will provide scontrol's default `nextstate=UNDRAIN` option, but
+  this will result in nodes remaining in DRAIN state as this role explicitly drains
+  nodes to prevent multinode jobs being backfilled onto onto a mix of "old" and
```
elelaysh:
```suggestion
  nodes to prevent multinode jobs being backfilled onto a mix of "old" and
```

ansible/roles/rebuild/tasks/rebuild.yml
elelaysh: I'm not sure: I thought it was because of batching and suggested a rewording, but README say it's to avoid backfill, which is plausible.

docs/experimental/slurm-controlled-rebuild.md
```diff
@@ -1,215 +1,177 @@
...
+
+5. Follow the [rebuild role readme](../../ansible/roles/rebuild/README.md) to
+   configure that role:
    - Add the `control` node into the `rebuild` group.
    - Ensure an application credential to use for rebuilding nodes is available
-     on the deploy host (default location `~/.config/openstack/clouds.yaml`).
-   - If required, override `rebuild_clouds_path` or other variables in the site
-     environment.
+     on the deploy host (default location `~/.config/openstack/clouds.yaml`) and
+     if necessary set `rebuild_clouds_path`
```
elelaysh, line 81:
```suggestion
     if necessary set `rebuild_clouds_path`.
```

docs/experimental/slurm-controlled-rebuild.md
```diff
@@ -1,215 +1,177 @@
...
 
-8. Run `tofu apply` as usual to apply the new OpenTofu configuration.
+6. Some defaults may need to be overridden based on testing. In particular the
+   `rebuild` role variables for batch size and delay may need modifying to avoid
```
elelaysh, line 84:
```suggestion
   `rebuild` role variables for batch size and delay (between `scontrol reboot` commands: the script doesn't wait for rebuild completion) may need modifying to avoid
```
Maybe this suggestion is too much. It is clear in the readme but people might skip it

docs/experimental/slurm-controlled-rebuild.md
```diff
@@ -1,215 +1,177 @@
...
+   ```
+
+7. Run `tofu apply` as usual to apply the new OpenTofu configuration.
 
-   > [!NOTE]
-   > If the cluster image references were updated at step 5, this will be
-   > a disruptive operation and should be planned as part of a normal upgrade
-   > cycle.
-   >
    > [!CAUTION]
```
elelaysh: This renders `[!CAUTION]` in text when I view the file. Will it produce nicer rendering when committed?

docs/experimental/slurm-controlled-rebuild.md
```diff
@@ -1,215 +1,177 @@
...
    > Due to OpenTofu/Terraform state limitations, this will plan to delete and
    > recreate all compute nodes in node groups where `ignore_image_changes: true`.
```
elelaysh:
```suggestion
   > recreate all compute nodes in node groups where `ignore_image_changes: true`
```

elelaysh (commented):
Great job :+1:!
I still have to review the end of the docs/experimental and the environments but here are a few nits.

docs/experimental/slurm-controlled-rebuild.md
```diff
@@ -1,215 +1,177 @@
...
+This section explains both the steps required and what happens at each step. Note
+this supplements the standard [upgrade docs](../../docs/upgrades.md).
+
+1. Build images and update image reference in the OpenTofu configuration in the
+   normal way.
+2. Run `tofu apply`: This rebuilds login and control nodes only to the new
+   image(s), and also updates the `hosts.yml` inventory file with the new compute
+   node image IDs. The login nodes are unavailable to users at this stage and
+   Slurm jobs cannot be submitted. Running jobs will continue, but will not be
+   able to complete.
```
elelaysh, line 131:
```suggestion
   Slurm jobs cannot be submitted. Running jobs will continue, but their completion will not be
   recorded correctly.
```

docs/experimental/slurm-controlled-rebuild.md
```diff
@@ -1,215 +1,177 @@
...
+   point the cluster is functional again; login nodes are available, running jobs
+   can complete and new jobs can be submitted. In this point the cluster is running
+   in a split state:
+   - Login and control nodes are on the new image, with new configuration
+   - Compute nodes are on the old image, with new configuration
+
+   This playbook also
+   - Writes cluster configuration to an NFS share `/exports/cluster` on the
+     control node (via the `compute_init` role).
+   - Deploys and configures the RebootProgram (via the `rebuild` role).
```
elelaysh:
```suggestion
   - Deploys the openstack credentials for the RebootProgram (via the `rebuild` role).
```

docs/experimental/slurm-controlled-rebuild.md
```diff
@@ -1,215 +1,177 @@
...
+
+   ```shell
+   ansible-playbook ansible/adhoc/rebuild-via-slurm.yml
+   ```
+
+   This will:
+   - Find all "rebuildable" nodes which are not currently on the latest image.
+   - In batches (to avoid avoid overloading OpenStack APIs):
+     - Set nodes to the DRAIN state, so that they can complete existing jobs
+       but not run new ones.
```
elelaysh:
```suggestion
   - Set nodes to the DRAIN state, so that they can complete existing jobs
     but not run new ones.
   - In batches (to avoid avoid overloading OpenStack APIs):

```
yes, drain commands will be batched, but without delay, so I find it more helpful to not mention it. Feel free to ignore the suggestion if you disagree.

ansible/roles/rebuild/tasks/rebuild.yml
sjpb: Well, I think it could happen from new jobs or backfill for non-exclusive jobs. But you're right this explicit drain is to do with batching - the readme references a separate backfill logic issue, which the old "job with --reboot" approach without any drain is vulnerable to, even without batching.

I'll try again ...

docs/experimental/slurm-controlled-rebuild.md
sjpb: I think I'd rather avoid this detail here tbh, there's too much going on.

docs/experimental/slurm-controlled-rebuild.md
sjpb: No, it turns out GFM can't do alerts inside numbered lists. The old NOTE was broken too, just no-one noticed. Fixing ...

docs/experimental/slurm-controlled-rebuild.md
sjpb: I don't think this change is correct. Jobs go to "completing" state, until the slurmctld comes back up, then they do. So there is nothing "incorrect" here. And point 3. says when they do complete.

docs/experimental/slurm-controlled-rebuild.md
sjpb: Have made a different change, you're right by default site.yml won't actually deploy it, that's done in the image build.

ansible/roles/rebuild/tasks/rebuild.yml
elelaysh: now it's clear. Thanks

docs/experimental/slurm-controlled-rebuild.md
elelaysh: ok, my mistake

docs/experimental/slurm-controlled-rebuild.md
elelaysh: ok

sjpb:
Extrabuilds failed on

```
==> openstack.openhpc: TASK [cuda : Install cuda packages] ********************************************
==> openstack.openhpc: task path: /home/runner/work/ansible-slurm-appliance/ansible-slurm-appliance/ansible/roles/cuda/tasks/install.yml:34
==> openstack.openhpc: Friday 18 September 2026  10:56:48 +0000 (0:00:00.034)       0:43:14.975 ******
==> openstack.openhpc: fatal: ***: FAILED! => {
==> openstack.openhpc:     "changed": false,
==> openstack.openhpc:     "results": []
==> openstack.openhpc: }
==> openstack.openhpc: 
==> openstack.openhpc: MSG:
==> openstack.openhpc: 
==> openstack.openhpc: Failed to download packages: nvidia-fs-dkms-2.30.2-1.x86_64: Cannot download, all mirrors were already tried without success
```

Manual testing shows the following are available, for RL9:

```
Available Packages
nvidia-fs.x86_64 2.30.2-1 cuda-rhel9-x86_64
nvidia-fs-dkms.x86_64 2.30.2-1 cuda-rhel9-x86_64
```

Repofile URL is https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64.

That repo does list https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/nvidia-fs-dkms-2.30.2-1.x86_64.rpm, but that returns 404. Previous version https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/nvidia-fs-dkms-2.30.2-1.x86_64.rpm is OK.

edit:
[Previous extra build](https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/34403940515/job/102793726240), worked ok with 
```
==> openstack.openhpc:         "Installed: nvidia-fs-dkms-2.30.1-1.x86_64",
```

Note that was 2.30.1, not 2.30.2.

sjpb:
It looks like cuda_packages_default from nvidia-gds triggers this package; we don't currently pin that. The previous successful version was nvidia-gds-13.4.1-1. The current workflow is trying to install 13.4.2-1.

It seems like the repos have got moved around a bit, as per https://docs.nvidia.com/gpudirect-storage/nvidia_fs/index.html#additional-package-availability-notes. So I suspect this has just got missed.

THe release notes suggest we'd be OK to pin back: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html

docs/experimental/slurm-controlled-rebuild.md
```diff
@@ -1,215 +1,174 @@
...
-5. Update image references in the OpenTofu configuration. Normally these should
-   be in:
+1. Decide which compute nodes should have rebuild via Slurm enabled. Generally,
+   this can be all compute nodes. These are referred to as the "rebuildable"
+   nodes below.
+
+2. Configure OpenTofu not to manage image changes on rebuildable nodes.
+
+   For each relevant node group in the OpenTofu `compute` variable, set the
+   parameter `ignore_image_changes: true`. E.g.
```
elelaysh: this is HCL syntax so better use `=` (the example is correct)

```suggestion
   parameter `ignore_image_changes = true`. E.g.
```

elelaysh (dismissed):
Tested as described in docs/slurm-controlled-rebuild.md: it works.

elelaysh (approved):
LGTM

PR #1039, opened 2026-09-14
Scan images in main branch every night and send Slack message on failure

elelaysh:
This adds scheduled grype scans of images in the main branch.

Failures are notified via a Slack incoming-webhook application.


elelaysh:
See https://github.com/stackhpc/ansible-slurm-appliance/actions/runs/35827909224/job/107073632460

PR #1041, opened 2026-09-15
hpctests minor fixes

elelaysh:
- fix `--nodelist` for pingmatrix and pingpong jobs
- add the `hpctests_time` variable to specify runtime limit on jobs


ansible/roles/hpctests/templates/pingmatrix.sh.j2
```diff
@@ -10,14 +10,17 @@
 #SBATCH --reservation={{ hpctests_reservation }}
 {% endif %}
 {%if hpctests_nodes is defined %}
-#SBATCH --nodelist={{ hpctests_computes.stdout_lines[0] }}
+#SBATCH --nodelist={{ hpctests_nodes }}
```
sjpb, line 13: This looks totally valid, given the docs for hpctests_nodes. I don't really understand why though, but tasks/setup.yml generates hpctests_computes from hpctests_nodes. So, if we're moving to using it directly, we should remove that task. But that also needs changes to tasks/hpl-solo.yml.

sjpb (changes requested):

ansible/roles/hpctests/templates/pingmatrix.sh.j2
elelaysh: I reverted the mistake (bad copy-paste) I made in this script in 08ef0634ef048bda64e3c9800c595cebb8ce0701.
But you're right: let me check

PR #994, opened 2026-06-29
Update RockyLinux 8.10 inc. OpenHPC v2.10, disable act_pedit

elelaysh:
- Update system packages on RockyLinux 8.10 including to kernel-core-4.18.0-553.136.1.el8_10 to fix CVE-2026-46331 (pedit-COW). Note system packages for RockyLinux 9.8 are not updated as kernel-core-5.14.0-687.17.1.el9_8 breaks DOCA-OFED.
- Disable kernel module `act_pedit` (on both OSes) to mitigate this CVE for RockyLinux 9.8.
- Update to [OpenHPC v2.10](https://github.com/openhpc/ohpc/releases/tag/v2.10.GA) on RockyLinux 8.10, including Slurm major upgrade from v23.11.10 -> v25.05.8.
- Update `stackhpc.openhpc` role to [v1.6.0](https://github.com/stackhpc/ansible-role-openhpc/releases/tag/v1.6.0), mostly moving appliance defaults into the role. However it fixes slurmdbd upgrade logic and does change defaults for the following `slurm.conf` parameters:
  - RebootProgram: unset -> `/sbin/reboot`: Allows `scontrol reboot` to be used to reboot nodes. If slurm controlled rebuild is configured this is overriden.
  - StateSaveLocation: `/var/spool/slurm` -> `/var/spool/slurmctld`: Slurmd state is lost on rebuild anyway so this is backwards-compatible.
  - [Amended 16/09/2026]: `SlurmdSpoolDir`: `/var/spool/slurm` -> slurm default of `/var/spool/slurmd`. **WARNING: THIS WILL BREAK RUNNING JOBS**

PR #1042, opened 2026-09-17
Bump openhpc role to validate SlurmdSpoolDir changes

sjpb:
#994 (released in v2.20) changed the default for SlurmdSpoolDir via incorporating https://github.com/stackhpc/ansible-role-openhpc/pull/206 as part of stackhpc.openhpc [v1.6.0](https://github.com/stackhpc/ansible-role-openhpc/releases/tag/v1.6.0) This causes running jobs to be lost when upgrading with slurm-controlled rebuilds enabled (or in fact any sitution where the slurmd spool directory contains state and its path is changed in configuration).

This PR fixes that bug by updating the openhpc role; it now raises an error if there is a slurmd spool directory which does not match SlurmdSpoolDir. Note normal upgrades, without slurm-controlled rebuilds enabled, are not affected as there is no slurmd spool directory in that case.

It also now runs the validation for the stackhpc.openhpc role when running the slurm.yml playbook with no tags or `--tags openhpc`, as this is stated in docs/operations.md as the appropriate way to change slurm configuration.

See the underlying role PR for more: https://github.com/stackhpc/ansible-role-openhpc/pull/220


sjpb:
@elelaysh notes:
> stackhpc.openhpc's validate.yml doesn't run from slurm.yml when appliances_mode=configure

So need to work out if its safe to enable this - note it might not be for image build where there isn't a cluster - or fix some other way.

elelaysh (approved):
LGTM

